PR #19, opened 2017-04-27
readme: typos and bad links

LeoColomb:
Very small changes, fixing typos in the Supported providers table! 💌


ross:
Made one minor edit. The names in the list were supposed to be the provider names rather than the company so it'd be DnsimpleProvider.

PR #27, opened 2017-04-30
Skip Alias recordset for Route53 Provider

michaelmcallister:
Fixes https://github.com/github/octodns/issues/26

Alias records are essentially CNAME records, but can be designated to
the Apex of the domain. They are typically for AWS resources only and
are not portable


michaelmcallister:
closing in favour of: https://github.com/github/octodns/pull/216

Issue #25, opened 2017-04-30
domainkey in octodns-dump


bcruik:
I get the following error when DKIM records are unescaped when dumping from cloudflare.

For example "v=DKIM1; k=rsa; ..." fails with:
Exception: Invalid record google._domainkey.domain.com., unescaped ;
However "v=DKIM1\\; k=rsa\\; ..." works fine.
Thats regardless if quotes are around the record.

Is this a bug, just a cloudflare thing or are my domainkey records technically wrong?

zBart:
I had the same issue, as a workaround I changed [line 102 in cloudflare.py](https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L102) from:
```
    _data_for_TXT = _data_for_multiple
```

To:
```
    def _data_for_TXT(self, _type, records):
        return {
            'ttl': records[0]['ttl'],
            'type': _type,
            'values': [r['content'].replace(';', '\;') for r in records],
        }
```
Which escapes the semicolons when the data is fetched. 

I don't know if this is proper behavior but it allowed me to fetch the data without having to change the record. Note that I haven't tested what happens when OctoDNS pushes the data to Cloudflare yet, so beware.

ross:
> I don't know if this is proper behavior but it allowed me to fetch the data without having to change the record.

That looks roughly correct, at least on the right path. There's [something similar in Route53](https://github.com/github/octodns/blob/faa4630362cca9f552271f84047f78d3db0b5791/octodns/provider/route53.py#L67-L68) for dealing with `"`. 

It'll probably require a corresponding change to `_contents_for_TXT`, something like:

```python
    def _contents_for_TXT(self, record):
        for value in record.values:
            yield {'content': value.replace('\;', ';')}
```

@zBart if you want to put together a PR update the tests for that and I'll take a :eyes:, if not I'll put it on my TODO list

ross:
Should be fixed in https://github.com/github/octodns/pull/29

zBart:
Thanks @ross !

Would it be possible to release a new version with this fix? Or is there a way I can tell `pip` to switch to the `master` version (don't have any experience with it)?

ross:
> Or is there a way I can tell pip to switch to the master version (don't have any experience with it)?

You can use a git URL with pip and specify a sha (recommended) or branch name (usually not a great idea since it can change out from under you)

```
-e git+ssh://git@github.com/github/octodns.git@<sha>#egg=octodns
```

[...]

PR #29, opened 2017-05-03
Handle Cloudflare not escaping semicolons

zBart:
This is a fix for #25 as requested by @ross 

In short the Cloudflare API does not escape semicolons in TXT records, which causes validation exceptions.

One important note about the test fixtures:
- For the `tests/fixtures/cloudflare-dns_records-page-2.json` data I added I did an actual request to their API to make sure the data matches what the API does.
- For `tests/fixtures/dnsimple-page-2.json` and `tests/fixtures/powerdns-full-data.json` however, I had to guess the way those values are stored. It would be nice if someone with access to either service can validate if this is correct.

### DIFF
--- a/tests/fixtures/dnsimple-page-2.json
+++ b/tests/fixtures/dnsimple-page-2.json
@@ -127,12 +127,28 @@
       "system_record": false,
       "created_at": "2017-03-09T15:55:11Z",
       "updated_at": "2017-03-09T15:55:11Z"
+    },
+    {
+      "id": 11189901,
+      "zone_id": "unit.tests",
+      "parent_id": null,
+      "name": "txt",
+      "content": "v=DKIM1\\;k=rsa\\;s=email\\;h=sha256\\;p=A/kinda+of/long/string+with+numb3rs",
+      "ttl": 600,
+      "priority": null,
+      "type": "TXT",
+      "regions": [
+        "global"
+      ],
+      "system_record": false,
+      "created_at": "2017-03-09T15:55:09Z",
+      "updated_at": "2017-03-09T15:55:09Z"
     }
   ],
   "pagination": {
     "current_page": 2,
     "per_page": 20,
-    "total_entries": 27,
+    "total_entries": 28,
     "total_pages": 2
   }
 }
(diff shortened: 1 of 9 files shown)


tests/fixtures/dnsimple-page-2.json
```diff
@@ -127,12 +127,28 @@
       "system_record": false,
       "created_at": "2017-03-09T15:55:11Z",
       "updated_at": "2017-03-09T15:55:11Z"
+    },
+    {
+      "id": 11189901,
+      "zone_id": "unit.tests",
+      "parent_id": null,
+      "name": "txt",
+      "content": "v=DKIM1\\;k=rsa\\;s=email\\;h=sha256\\;p=A/kinda+of/long/string+with+numb3rs",
```
ross: This seems to match what DNSimple returns from their API

```json
      "content": "foo\\; bar baz\\; done",
```

tests/fixtures/powerdns-full-data.json
```diff
@@ -86,6 +86,10 @@
                 {
                     "content": "\"have you any wool.\"",
                     "disabled": false
+                },
+                {
+                    "content": "\"v=DKIM1\\;k=rsa\\;s=email\\;h=sha256\\;p=A/kinda+of/long/string+with+numb3rs\"",
```
ross: And this seems to match PowerDNS's behavior

```json
          "content": "\"foo\\; bar baz\\; done\"",
```

tests/test_octodns_provider_cloudflare.py
```diff
@@ -160,9 +160,16 @@ def test_apply(self):
                 'name': 'under.unit.tests',
                 'ttl': 3600
             }),
-        ])
+            # make sure semicolons are not escaped when sending data
+            call('POST', '/zones/42/dns_records', data={
+                'content': 'v=DKIM1;k=rsa;s=email;h=sha256;p=A/kinda+of/long/string+with+numb3rs',
```
ross: This line fails the checks in `./script/lint` (which also just made me realize that I forgot to setup CI on this repo :frowning_face:, looking at that now.

```
sweetums:octodns ross$ ./script/lint
tests/test_octodns_provider_cloudflare.py:165:80: E501 line too long (98 > 79 characters)
```

PR #33, opened 2017-05-10
First pass through NsOneProvider

ross:
This is a quick pass through adding basic record support for NS1. 

* [x] Tests

### DIFF
--- a/octodns/provider/nsone.py
+++ b/octodns/provider/nsone.py
@@ -1,0 +1,189 @@
+#
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from logging import getLogger
+from nsone import NSONE
+from nsone.rest.errors import ResourceException
+
+from ..record import Record
+from .base import BaseProvider
+
+
+class NsOneProvider(BaseProvider):
+    '''
+    NsOne provider
+
+    nsone:
+        class: octodns.provider.nsone.NsOneProvider
+        api_key: env/NS_ONE_API_KEY
+    '''
+    SUPPORTS_GEO = False
+
+    def __init__(self, id, api_key, *args, **kwargs):
+        self.log = getLogger('NsOneProvider[{}]'.format(id))
+        self.log.debug('__init__: id=%s, api_key=***', id)
+        super(NsOneProvider, self).__init__(id, *args, **kwargs)
+        self._client = NSONE(apiKey=api_key)
+
+    def _data_for_A(self, _type, record):
+        return {
+            'ttl': record['ttl'],
+            'type': _type,
+            'values': record['short_answers'],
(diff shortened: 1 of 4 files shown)


octodns/provider/nsone.py
```diff
@@ -0,0 +1,189 @@
...
+
+from logging import getLogger
+from nsone import NSONE
+from nsone.rest.errors import ResourceException
+
+from ..record import Record
+from .base import BaseProvider
+
+
+class NsOneProvider(BaseProvider):
```
weyrick: despite our python client still using our older `nsone` moniker, we were hoping you could shift this to be `NS1Provider` and in general use `ns1` instead of `nsone`, which is what we've rebranded to since the original python client came out.

octodns/provider/nsone.py
```diff
@@ -0,0 +1,189 @@
...
+    '''
+    NsOne provider
+
+    nsone:
+        class: octodns.provider.nsone.NsOneProvider
+        api_key: env/NS_ONE_API_KEY
+    '''
+    SUPPORTS_GEO = False
```
weyrick: it seems like it may be fairly easy to get geo support in, let us know if we can help with that!

octodns/provider/nsone.py
ross: Done.

octodns/provider/nsone.py
ross: Some have been easier than others. Haven't looked deeply enough at NS1's model & api to know what it'd involve. Don't plan to add it in this round since I won't have enough cycles to dig in. Once this is merged I'd be happy to look over a PR that adds support for it 😁.

ross:
@weyrick tests have been fleshed out and a few fixes put in place. If you want to 👀  it again and give a 👍  / 👎  I think it's ready for it.

weyrick:
👍  thanks!

williamsjj:
What level of geo support is included in this version of the NS1 provider?

ross:
> What level of geo support is included in this version of the NS1 provider?

There's currently no geo support in `Ns1Provider`. This PR was just a first pass to get basic record support in place. Working with the geo dns apis of providers is usually quite a bit more involved and testing all of the edge cases takes a bit of work as well. That's not something I'll be able to tackle in the near-term as since its more time that I can spare for the purpose in nights & weekends.

williamsjj:
@ross That's too bad but I totally understand.

@weyrick Any chance NS1 can contribute geo support? Understanding it will be a limited application of filter chains to achieve that. NS1 keeps banging on us to go multi-DNS everytime there's an outage, so this would help that a lot.

vanbroup:
See also ns1/nsone-python#28

PR #40, opened 2017-05-24
use a percentage of change/deleted records rather than an absolute count

joewilliams:
This PR updates the way that a plan is deemed safe or unsafe. It switches to percent based on existing records over a threshold, currently set to 30% for any zone that has over ten existing records.So if a zone has ten records and more than three updates are happening then it will exit as unsafe. A minimum number of existing records was set to avoid cases where the zone is, for instance, two of records and one was updated which would cause an unsafe plan. 

Additionally this PR adds a bit of debug logging and tests.

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -35,18 +35,16 @@ class Plan(object):
             change_counts[change.__class__.__name__] += 1
         self.change_counts = change_counts
 
-        if self.existing:
-            self.log.debug('__init__: Creates=%s, Updates=%s, Deletes=%s'
-                           'Existing=%s',
-                           self.change_counts['Create'],
-                           self.change_counts['Update'],
-                           self.change_counts['Delete'],
-                           len(self.existing.records))
-        else:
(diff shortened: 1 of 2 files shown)


octodns/provider/base.py
```diff
@@ -31,12 +35,42 @@ def __init__(self, existing, desired, changes):
...
+                           'Existing=%s',
+                           self.change_counts['Create'],
+                           self.change_counts['Update'],
+                           self.change_counts['Delete'],
+                           len(self.existing.records))
+        else:
+            self.log.debug('__init__: Creates=%s, Updates=%s, Deletes=%s',
+                           self.change_counts['Create'],
+                           self.change_counts['Update'],
+                           self.change_counts['Delete'])
```
ross: To avoid having to duplicate the full log line/code you could do something like.

```python
try:
    existing_n = len(self.existing.records)
except AttributeError:
    existing_n = 0

self.log.debug('__init__: Creates=%d, Updates=%d, Deletes=%d, Existing=%d',
                        self.change_counts['Create'], self.change_counts['Update'],
                        self.change_counts['Delete'], existing_n)
```

PR #45, opened 2017-05-27
Add SoaRecord object and tests

ross:
* [x] Add SoaRecord object and tests for it's behavior
   * will ignore diffs in `mname` and `serial`, other fields will count
* [ ] Make a pass through adding SOA management support to providers
   * Route53 takes whatever I give it, this means it can be managed but the whole record will be overriten by what's in config including the `mname` and `serial`.
   * Dyn api seems to ignore the call to change the SOA record no matter what I pass in, no error or anything, so I don't think it can be managed through the API. A few of the fields can be managed in the UI. Will ask them about this, but as of now it's not going to support SOA.

Fixes https://github.com/github/octodns/issues/39
/cc @vanbroup

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -70,6 +70,13 @@ class _Route53Record(object):
     _values_for_SPF = _values_for_quoted
     _values_for_TXT = _values_for_quoted
 
+    def _values_for_SOA(self, record):
+        value = record.value
+        return ['{} {} {} {} {} {} {}'.format(value.mname, value.rname,
+                                              value.serial, value.refresh,
+                                              value.retry, value.expire,
+                                              value.minimum)]
+
     def _values_for_SRV(self, record):
         return ['{} {} {} {}'.format(v.priority, v.weight, v.port,
                                      v.target)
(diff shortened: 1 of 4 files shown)


ross:
I have Route53 working, but Dyn seems to ignore SOA updates via the API. If Dyn is a one-off I might continue this, but if the wonkiness continues elsewhere I'm probably going to abort the attempt and not wade into the mess.

vanbroup:
The Dyn API looks to support updates to the SOA record:
https://help.dyn.com/update-soa-record-api/

Same for the used Dyn API Python SDK:
http://dyn.readthedocs.io/en/latest/tm/records/records.html#soarecord

Could the problem be that you try to delete/create an SOA record instead of updating it? I haven't looked at the code so not sure about the used logic.

> NOTE: Dynect users do not have the permissions required to create or delete SOA records on the Dynect System.

vanbroup:
For NS1 you can modify the ttl, refresh, retry, expiry, or nx_ttl within the SOA record by modifying the zone:
https://jsapi.apiary.io/apis/ns1api/reference/zones-and-records/zone/modify-a-zone.html

The SDK documentation for updating the zone information:
http://nsone.readthedocs.io/en/latest/api/zones.html#nsone.zones.Zone.update

PR #47, opened 2017-05-29
Alias support

ross:
This is WIP for the moment. Started working on implementing Route53 provider and quickly ran into the need for the cleanup in https://github.com/github/octodns/pull/46 so I had to side track to do that first.

I still need to look at `ALIAS` support requirements in Dyn, DNSimple, etc. to make sure the underlying schema here makes sense for all of them, but as of now an `AliasRecord` has one or more values each of which references another record in the same zone by `name` and `type`. This matches Route53's requirements and more generally what things need to look like conceptually. 

/cc https://github.com/github/octodns/issues/26 & @michaelmcallister 
/cc https://github.com/github/octodns/issues/30 & @jalogisch

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ The above command pulled the existing data out of Route53 and placed the results
 
 * ALIAS support varies a lot fromm provider to provider care should be taken to verify that your needs are met in detail.
    * Dyn's UI doesn't allow editing or view of TTL, but the API accepts and stores the value provided, this value does not appear to be used when served
-   * Dnsimple's API throws errors when TTL is modified, but it can be edited in the UI and seems to be used when served, there's also a secondary TXT record created alongside the ALIAS that octoDNS ignores
+   * Dnsimple's uses the configured TTL when serving things through the ALIAS, there's also a secondary TXT record created alongside the ALIAS that octoDNS ignores
 
 ## Custom Sources and Providers
 
--- a/README.md
+++ b/README.md
@@ -158,6 +158,12 @@ The above command pulled the existing data out of Route53 and placed the results
 | [TinyDNSSource](/octodns/source/tinydns.py) | A, CNAME, MX, NS, PTR | No | read-only |
 | [YamlProvider](/octodns/provider/yaml.py) | All | Yes | config |
 
+#### Notes
+
+* ALIAS support varies a lot fromm provider to provider care should be taken to verify that your needs are met in detail.
+   * Dyn's UI doesn't allow editing or view of TTL, but the API accepts and stores the value provided, this value does not appear to be used when served
+   * Dnsimple's API throws errors when TTL is modified, but it can be edited in the UI and seems to be used when served, there's also a secondary TXT record created alongside the ALIAS that octoDNS ignores
+
 ## Custom Sources and Providers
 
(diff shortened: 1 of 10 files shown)


ross:
Heads up that I'm bailing on `ALIAS` support for `Route53Provider`. Route53's version of the record is just too different/special and actually won't solve a lot of our ancillary domain's needs. May be revisited in the future, but from the looks of it it's gonna require hacks to pull off what's needed. Going to start with `DynProvider` and go from there.

ross:
This is ready for review. I'm happy with the direction. Hope to come back to Route53 in the future, but for now it's out of scope.

README.md
```diff
@@ -158,6 +158,12 @@ The above command pulled the existing data out of Route53 and placed the results
 | [TinyDNSSource](/octodns/source/tinydns.py) | A, CNAME, MX, NS, PTR | No | read-only |
 | [YamlProvider](/octodns/provider/yaml.py) | All | Yes | config |
 
+#### Notes
+
+* ALIAS support varies a lot fromm provider to provider care should be taken to verify that your needs are met in detail.
+   * Dyn's UI doesn't allow editing or view of TTL, but the API accepts and stores the value provided, this value does not appear to be used when served
+   * Dnsimple's API throws errors when TTL is modified, but it can be edited in the UI and seems to be used when served, there's also a secondary TXT record created alongside the ALIAS that octoDNS ignores
```
nautalice: Can we catch these errors or wrap them with an "expected" comment?

nautalice:
Is there a way to detect if someone is trying to add an ALIAS to Route53 and tell them that's not currently supported?

ross:
> Is there a way to detect if someone is trying to add an ALIAS to Route53 and tell them that's not currently supported?

There's currently a table of type support [in the readme](https://github.com/github/octodns#supported-providers) and (currently debug level) logging when records are [skipped based](https://github.com/github/octodns/blob/98e067002085a715b53f4dddf2f30e8fc7fb38da/octodns/zone.py#L86-L90) [on support](https://github.com/github/octodns/blob/98e067002085a715b53f4dddf2f30e8fc7fb38da/octodns/zone.py#L112-L116)

README.md
ross: Not sure. It might be possible, but the hiccup would be that it may not be clear that it's an `ALIAS` record at the point the error is thrown. Would have to dig into it more.

ross: The other issue is that if it was handled, it'd continue to show up as a diff every time `sync` is run and continue to try and update/fix it and fail over and over. Might be best to look at ignoring TTL altogether on DNSimple `ALIAS` records.

nautalice:
The README could suffice. If logging, maybe do at a WARNING level instead of INFO?

README.md
ross: Double checked this and actually I fixed the error that was happening here and forgot to update the README. So that part doesn't apply any longer. I did leave the bits about DNSimple seeming to obey the ALIAS ttl and creating a ride along record. Thanks for catching this.

ross:
> The README could suffice. If logging, maybe do at a WARNING level instead of INFO?

Created https://github.com/github/octodns/issues/63 to specifically track the logging stuff.

@nautalice I think this is ready for a 👍  / 👎

PR #53, opened 2017-06-04
Major reworking of root NS handling, create & update, no delete

ross:
- Cloudflare, Dyn & DnsimpleProvider don't support editing root NS records, Dyn may be
  able to sort of, but you can't remove their own...
- Improve DnsimpleClient's info on http errors
- Remove special root NS handling code from PowerDnsProvider, in this the
  case it used to support things should now be managed in the zone. Dynamic
  support still exists in PowerDnsBaseProvider to enable "other" use cases
- _is_eligible removed from zone, all records are eligible as far as it's
  concerned now.
- Bunch of tweaks to tests to handle the fact that root NS stuff are now
  included in changes.

Closes https://github.com/github/octodns/issues/38

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,9 +149,9 @@ The above command pulled the existing data out of Route53 and placed the results
 
 | Provider | Record Support | GeoDNS Support | Notes |
 |--|--|--|--|
-| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CNAME, MX, NS, SPF, TXT | No | |
-| [DnsimpleProvider](/octodns/provider/dnsimple.py) | All | No | |
-| [DynProvider](/octodns/provider/dyn.py) | All | Yes | |
+| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CNAME, MX, NS, SPF, TXT | No | Unknown state, don't have access to test |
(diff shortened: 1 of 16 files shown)


ross:
OK. Looked this over more closely and have done some dry-run testing internally. The behavior seems to be as designed/desired so this is probably ready for thinking/review. 

/cc @joewilliams @vanbroup

joschi36:
any news on this?

ross:
@joschi36 IIRC this stalled out when we ran into problems with many/most providers allowing us to manage the root NS records. Specifically that each provider's behavior was different and hard unify. It'd likely require a new provider ability "SUPPORTS_ROOT_NS" etc and even though would get tricky to handle all the edge cases. It may be possible to go through and get everything working in a best effort fashion, but this PR probably isn't sufficient. It's been left here to record the thinking/work.

It'd be nice to have, but unfortunately isn't high on the priorities list.

joschi36:
For us this has high priority as we plan to migrate around 600 zones with multiple octodns repos in github. For us it isn't an option and I'm currently working on a new PR resolving this issue for the providers that support it.

dchauran:
@joschi36 do you have a link to that? My team also needs this.

@basirjamil, @Arunothia FYI

ross:
There's some WIP towards this over in https://github.com/github/octodns/pull/434.

PR #56, opened 2017-06-05
Add support for ignored records.

ross:
This seems to work well enough, but I want to mull it over a bit before I'm ready to :shipit:. Partially to think over the specifics, but probably more to figure out if the `octodns` key thing makes sense and would be able to support some of the other "special" stuff that will eventually need to be added (specifically per-provider info.)

```yaml
www:
  octodns:
    ignored: true
  type: A
  value: 1.2.3.4
```

/cc https://github.com/github/octodns/issues/49
/cc @dolvany

### DIFF
--- a/octodns/record.py
+++ b/octodns/record.py
@@ -112,6 +112,9 @@ class Record(object):
             raise Exception('Invalid record {}, missing ttl'.format(self.fqdn))
         self.source = source
 
+        octodns = data.get('octodns', {})
+        self.ignored = octodns.get('ignored', False)
+
     def _data(self):
         return {'ttl': self.ttl}
 
--- a/octodns/zone.py
+++ b/octodns/zone.py
@@ -76,8 +76,12 @@ class Zone(object):
 
         # Find diffs & removes
         for record in filter(_is_eligible, self.records):
+            if record.ignored:
+                continue
             try:
                 desired_record = desired_records[record]
+                if desired_record.ignored:
+                    continue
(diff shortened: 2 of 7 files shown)


ross:
Decided I like this pattern, `octodns` key for stuff that's directing octoDNS to do something. There's a few other places it'll likely get used (configurable health check urls probably being next.)

dolvany:
@ross, thanks for the feature add! I am curious how this feature interacts with multiple records. Given the below example, how would I go about ignoring the A record, but not the TXT record? How would I ignore both?
```
www:
- ttl: 30
  type: A
  values:
  - 1.1.1.1
  - 2.2.2.2
- ttl: 30
  type: TXT
  value: v=spf1 -all
```

ross:
> @ross, thanks for the feature add! I am curious how this feature interacts with multiple records. Given the below example, how would I go about ignoring the A record, but not the TXT record? How would I ignore both?

In that `www` is an array. Each item is a record, `octodns` would be a key in one or more of those. 

```yaml
www:
- ttl: 30
  type: A
  values:
  - 1.1.1.1
  - 2.2.2.2
- octodns:
    ignored: true
  ttl: 30
  type: TXT
  value: v=spf1 -all
```

dolvany:
@ross, hmm, I thought I tried that permutation, but I guess not. I took the ignored directive for a test drive just now and it works as advertised. Thanks!

PR #58, opened 2017-06-08
Fixed UnboundLocalError in manager.py

offmindby:
I've got following traceback while running this command-line
`octodns-sync --debug --config-file=./config/production.yaml --doit --target pdns test.it`

2017-06-08T14:21:02  [140735208611840] INFO  Manager __init__: config_file=./config/production.yaml
2017-06-08T14:21:02  [140735208611840] DEBUG Manager __init__:   configuring providers
2017-06-08T14:21:03  [140735208611840] DEBUG PowerDnsProvider[pdns] __init__: id=pdns, host=10.135.192.22, port=8081, nameserver_values=None, nameserver_ttl=600
2017-06-08T14:21:03  [140735208611840] DEBUG PowerDnsProvider[pdns] __init__: id=pdns, apply_disabled=False
2017-06-08T14:21:03  [140735208611840] DEBUG Route53Provider[route53] __init__: id=route53, access_key_id=AKIAJ6***LLNA, secret_access_key=***
2017-06-08T14:21:03  [140735208611840] DEBUG Route53Provider[route53] __init__: id=route53, apply_disabled=False
2017-06-08T14:21:03  [140735208611840] INFO  Manager sync: eligible_zones=['***.it'], eligible_targets=['pdns'], dry_run=False, force=False
2017-06-08T14:21:03  [140735208611840] INFO  Manager 
No changes were planned

2017-06-08T14:21:03  [140735208611840] DEBUG Manager sync:   checking safety
  File "/Users/test/tmp/octodns/bin/octodns-sync", line 11, in <module>
    sys.exit(main())
  File "/Users/test/tmp/octodns/lib/python2.7/site-packages/octodns/cmds/sync.py", line 39, in main
    dry_run=not args.doit, force=args.force)
  File "/Users/test/tmp/octodns/lib/python2.7/site-packages/octodns/manager.py", line 275, in sync
    if dry_run or config.get('always-dry-run', False):
UnboundLocalError: local variable 'config' referenced before assignment

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -273,7 +273,7 @@ class Manager(object):
             for target, plan in plans:
                 plan.raise_if_unsafe()
 
-        if dry_run or config.get('always-dry-run', False):
+        if dry_run or self.config.get('always-dry-run', False):
             return 0
 
         total_changes = 0


octodns/manager.py
```diff
@@ -273,7 +273,7 @@ def sync(self, eligible_zones=[], eligible_targets=[], dry_run=True,
             for target, plan in plans:
                 plan.raise_if_unsafe()
 
-        if dry_run or config.get('always-dry-run', False):
+        if dry_run or self.config.get('always-dry-run', False):
```
ross: This is definitely a bug. Adding `self.` will prevent the incorrect access, but it'll actually get things looking into the wrong place for `always_dry_run`. That's supposed to be a per-zone config that prevents changes from being made to that zone. Looks like some refactoring has happened since the last time I tested this specific path. Probably the first thing to do is add a test case for it so that it doesn't happen again and the correct fix can be put in place. 

I'll have to look at it closer, but I think a bit more involved changes may be necessary. With how that function is structured now it may require adding a dry_run flag to the plan so that it can be skipped when encountered rather than aborting the whole sync. 

It looks like this path hasn't been tested since a long time ago when `sync` was per-zone :frowning_face:. Thanks for catching it. I think you may have b/c of an ~edge case where your zone didn't have the trailing `.` so the `config` var in the foreach up above was never set. 

`octodns-sync --debug --config-file=./config/production.yaml --doit --target pdns test.it`

vs 

`octodns-sync --debug --config-file=./config/production.yaml --doit --target pdns test.it.`

ross:
https://github.com/github/octodns/pull/60 replaces this with a fix and unit tests. Thanks.

PR #59, opened 2017-06-09
Escape unescaped semicolons coming out of Route53

ross:
Route53 automatically escapes unescaped `;` when it serves TXT records, but octoDNS enforces them since that's technically correct.

This handles pulling records from Route53 when they don't have `\;`. They won't show as diffs, but any time octoDNS goes to modify these records for any reason they'll be fixed.

/cc @brianeclow @blakestoddard please give this a look/try and 👍 / 👎 
Fixes https://github.com/github/octodns/issues/34


blakestoddard:
I just gave it a run and it worked perfectly for me! Thanks for knocking this out so quickly!

PR #67, opened 2017-06-15
Configurable geo healthcheck host & path

ross:
#This adds new `octodns` custom config values to records to support providing custom healthcheck paths and hosts. E.g.

```
'':
  geo:
    AS-JP:
    - 2.3.4.5
  octodns:
    healthcheck:
      host: foo.bar.io
      path: /_ok
  ttl: 60
  type: A
  value: 1.2.3.4 
```

~~NOTE: this will not clean up the previous Route53 healthchecks do to changes made in the way their `CallerReferenece`s work.~~


jespada:
@ross can I same how help with this PR so it gets merged :) ?

octodns/provider/base.py
```diff
@@ -64,7 +64,8 @@ def plan(self, desired):
             self.log.info('plan:   filtered out %s changes', before - after)
 
         # allow the provider to add extra changes it needs
-        extra = self._extra_changes(existing, changes)
+        extra = self._extra_changes(existing=existing, desired=desired,
+                                    changes=changes)
```
ross: This change allows providers to catch them by name rather than position and thus ignore the params they don't want with `**kwargs`.

octodns/provider/route53.py
```diff
@@ -230,7 +230,7 @@ class Route53Provider(BaseProvider):
 
     # This should be bumped when there are underlying changes made to the
     # health check config.
-    HEALTH_CHECK_VERSION = '0000'
+    HEALTH_CHECK_VERSION = '0001'
```
ross: Have to bump Route53's HC version, this will cause a replacement of all health-checks via _extra_changes including all Geo records in the plan.

octodns/provider/route53.py
```diff
@@ -560,30 +572,33 @@ def get_health_check_id(self, record, ident, geo, create):
...
             'RequestInterval': 10,
-            'ResourcePath': '/_dns',
-            'Type': 'HTTPS',
+            'ResourcePath': healthcheck_path,
+            'Type': healthcheck_protocol,
         }
-        ref = '{}:{}:{}'.format(self.HEALTH_CHECK_VERSION, record._type,
-                                uuid4().hex[:16])
+        ref = '{}:{}:{}:{}'.format(self.HEALTH_CHECK_VERSION, record._type,
+                                   record.name, uuid4().hex[:16])
```
ross: Include the record name in the healthcheck ref, not just the type.

octodns/provider/dyn.py
```diff
@@ -389,6 +489,34 @@ def populate(self, zone, target=False, lenient=False):
...
+                continue
+            label = '{}:{}'.format(record.fqdn, record._type)
+            try:
+                monitor = self.traffic_director_monitors[label]
+            except KeyError:
+                self.log.info('_extra_changes: health-check missing for %s',
+                              label)
+                extra.append(Update(record, record))
+                continue
+            if _monitor_matches(monitor, record.healthcheck_host,
```
theojulienne: `_monitor_matches` seems to actually be checking if the monitor has any differences to the variables provided. Possibly invert the logic?

PR #76, opened 2017-06-23
Route53Provider: Handle limit of 32k character limit exceeded

aikram24:
Updated Doc, change the default max_changes to 100.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
+build
+log
+config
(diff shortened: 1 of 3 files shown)


ross:
Hi @aikram24 sorry I didn't see this PR come through back when we were discussing things in https://github.com/github/octodns/issues/73. As mentioned there I'd prefer to handle the 32k char case similar to how we already break things up for the number of records. I'll add looking into this on to my TODO list.

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-route53/. Feel free to rework/reopen this PR there.

PR #81, opened 2017-06-28
Ns1 rate limit handling

ross:
What should the default retry delay be?

/cc https://github.com/github/octodns/issues/51#issuecomment-311495388
/cc @weyrick @vanbroup @pashap

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -28,13 +28,11 @@ class Ns1Provider(BaseProvider):
 
     ZONE_NOT_FOUND_MESSAGE = 'server error: zone not found'
 
-    def __init__(self, id, api_key, rate_limit_delay=1, *args, **kwargs):
+    def __init__(self, id, api_key, *args, **kwargs):
         self.log = getLogger('Ns1Provider[{}]'.format(id))
-        self.log.debug('__init__: id=%s, api_key=***, rate_limit_delay=%d', id,
-                       rate_limit_delay)
+        self.log.debug('__init__: id=%s, api_key=***', id)
         super(Ns1Provider, self).__init__(id, *args, **kwargs)
         self._client = NSONE(apiKey=api_key)
-        self.rate_limit_delay = rate_limit_delay
 
     def _data_for_A(self, _type, record):
         return {
@@ -177,10 +175,10 @@ class Ns1Provider(BaseProvider):
         meth = getattr(nsone_zone, 'add_{}'.format(_type))
         try:
             meth(name, **params)
-        except RateLimitException:
+        except RateLimitException as e:
             self.log.warn('_apply_Create: rate limit encountered, pausing '
                           'and trying again')
-            sleep(self.rate_limit_delay)
+            sleep(e.period)
             meth(name, **params)
 
     def _apply_Update(self, nsone_zone, change):
@@ -192,10 +190,10 @@ class Ns1Provider(BaseProvider):
         params = getattr(self, '_params_for_{}'.format(_type))(new)
         try:
             record.update(**params)
-        except RateLimitException:
+        except RateLimitException as e:
             self.log.warn('_apply_Update: rate limit encountered, pausing '
                           'and trying again')
-            sleep(self.rate_limit_delay)
+            sleep(e.period)
             record.update(**params)
 
     def _apply_Delete(self, nsone_zone, change):
@@ -205,10 +203,10 @@ class Ns1Provider(BaseProvider):
         record = nsone_zone.loadRecord(name, _type)
         try:
             record.delete()
-        except RateLimitException:
+        except RateLimitException as e:
             self.log.warn('_apply_Delete: rate limit encountered, pausing '
                           'and trying again')
-            sleep(self.rate_limit_delay)
+            sleep(e.period)
             record.delete()
 
     def _apply(self, plan):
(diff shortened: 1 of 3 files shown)


weyrick:
@ross Thanks! If you want to be exact, you can do some math using the `limit` and `period` properties returned in the rate limit exception. But, I think a default of 1s is pretty reasonable as well.

ross:
> @ross Thanks! If you want to be exact, you can do some math using the limit and period properties returned in the rate limit exception. But, I think a default of 1s is pretty reasonable as well.

I was unable to cause a rate-limit error, probably partly due to round trip time of requests, pretty low throughput in my tests. When I try and I hit the number of records limit first, that was after 60s, so it's doing less than 1s. 

```
nsone.rest.errors.ResourceException: server error: You have reached your maximum record count of 50 records. Please contact support@ns1.com to raise the limit.
```

I could potentially parallelize it a bit and try and hit things to get real data, but it'd probably be eaiser if you can just provide an example of the data that'll come back or even better the math I should use to calculate the sleep. If you'd prefer we can go with the `1s` sleep and see how that goes.

vanbroup:
I would say that the max number of queries is `period / limit`, but for small zones you might not want to delay every request.

- If the number of requests is smaller or equal to limit we don't need to worry about the backoff.
- If the number of requests is larger than limit we need to enable the backoff.

Say that we have 1000 updates but we are only allowed to do 100 (limit) updates per minute (period) this means that the backoff should be enabled and that we can only do 0.6 requests per second.

In this case if the quest itself takes 300ms we still need to wait 1.3 seconds before we can make the next request. 

If we statically define a sleep period I would suggest doing this in the provider config as ms. The amount of time the client should wait could be impacted by multiple factors including the connection latency. The standard should be safe for everyone but could slow down the overall provisioning process.

ross:
Yeah, fwiw, my questions are less about the math and more about the expected values. I'm hesitant to commit code/logic that I can't test out. 

The client library also doesn't seem to provide any access to those headers until the exception is raised so trying to pace things doesn't really make sense given that I'd have to wait until the problem had happened (exception thrown) to know what pace would be allowed, by which point it's too late. 

Looking at this more deeply now I'm probably should have avoided the client library as it's getting in the way more than it's helping

The headers are set on the response, they just don't seem to be accessible through the lib.

```
[<Response [200]>,
 {'X-RateLimit-Remaining': '99', ... 'X-RateLimit-By': 'customer', ... 'X-RateLimit-Limit': '100', 'X-RateLimit-Period': '200', ...}]
```

Anyway, based on the [KB article](https://ns1.com/articles/api-rate-limiting) the rate-limit period is in seconds, so 200s there.

Seems like a 1s `sleep` whenever we hit a problem won't do any good and just spacing out things with sleeps between every operation would be annoying for small changes. Given that I can't tell when to do what the best option I seem to have is to just sleep for the rate limit period `e.period` before trying again at which point the tokens should have refilled. For huge changes that should be pretty much optimal, for small changes they should never hit the limit. 

Without replacing the client library (which is probably worth considering if/when advanced record support is added) that's probably the best that can be done.

weyrick:
@ross You're right, the client lib should provide access to those headers before an exception occurs - I will get that change on our roadmap.

Regarding how to handle this re: sleeps for the time being - I don't think it's a good strategy to sleep for `period`. As the KB article mentions, we use a token bucket algorithm which allows for bursting. `period` is the time period it would need to regenerate _all_ tokens ... but it only needs 1 token per request. So in your example above, under that strategy, it would sleep for 200 seconds and then be able to make another 100 requests (as fast as possible, in one burst if it wanted).

I think a better strategy would be to let it burst as fast as it can, then send requests at the rate limit - in this case, that's 100req/200s = .5 req/s. Meaning, at this limit, requests will succeed once every 2 seconds.

I thought the 1s default sleep was reasonable, because at most of our rate limits, this would mean at most 3 retries before the request would succeed (with 1s pauses in between), and otherwise, no (extremely) lengthy pauses.

Does that make sense?

ross:
> Regarding how to handle this re: sleeps for the time being - I don't think it's a good strategy to sleep for period. As the KB article mentions, we use a token bucket algorithm which allows for bursting. period is the time period it would need to regenerate all tokens ... but it only needs 1 token per request. So in your example above, under that strategy, it would sleep for 200 seconds and then be able to make another 100 requests (as fast as possible, in one burst if it wanted).

Yeah, it's not an ideal strategy, but it's one that will 100% work. As mentioned above it's pretty close to optimal for really huge changes (100's of records etc.) and shouldn't be encountered during normal/small changes. 

> I think a better strategy would be to let it burst as fast as it can, then send requests at the rate limit - in this case, that's 100req/200s = .5 req/s. Meaning, at this limit, requests will succeed once every 2 seconds.

That'd work, but adding in the pacing was a bigger change than I wanted to dig into atm. Especially since the rate info I'd get back would only correctly apply for the type of operation that threw the initial exception. E.g. if it was a zone create the rate/cost info I'd get would be completely different from a record create, update, or delete. Generally though I was aiming for getting something that'd work and moving on to other requests/core functionality 😄 

> I thought the 1s default sleep was reasonable, because at most of our rate limits, this would mean at most 2 or 3 retries before the request would succeed (with 1s pauses in between), and otherwise, no (extremely) lengthy pauses.

As things were implemented there's only a 1-shot retry so the 1s sleep wouldn't have been able to do that. It needed to sleep until it was sure things would work.

weyrick:
Gotcha ok, agreed it's a simple strategy that should work - so long as operators don't kill the script thinking it's hung during long sleep periods. And, we can always iterate on this later for improvements. Thanks!

ross:
> so long as operators don't kill the script thinking it's hung during long sleep periods.

Yup. I updated things to include a `warn` level log message saying exactly how long things were going to sleep. That should hopefully get people's attention in contexts where they're running it interactively.

PR #83, opened 2017-06-29
Supply 'Marker' to Route53 client when paging in order to correctly r…

anthonyvia:
…etrieve the next page of results.

If you don't supply 'Marker' to the client, you will loop forever retrieving the first page of results.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -250,7 +250,7 @@ class Route53Provider(BaseProvider):
             more = True
             start = {}
             while more:
-                resp = self._conn.list_hosted_zones()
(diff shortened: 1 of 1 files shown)


octodns/provider/route53.py
```diff
@@ -250,7 +250,7 @@ def r53_zones(self):
             more = True
             start = {}
             while more:
-                resp = self._conn.list_hosted_zones()
+                resp = self._conn.list_hosted_zones(**start)
```
ross: Doh! Good catch.

anthonyvia:
Thanks for the quick turnaround!

PR #84, opened 2017-06-30
Azure DNS Support PR

h-hwang:
Passes all tests and 100% coverage. script/lint claims some imports are unused on line 10 of provider/azuredns.py but are used in <..., eval(class_name)> in line 55.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@ output/
 tmp/
 build/
 config/
-./rb.txt
-./doit.txt
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
+Makefile
+build/
+config/
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
-Makefile
 build/
 config/
+./rb.txt
+./doit.txt
--- a/doit.txt
+++ b/doit.txt
@@ -1,0 +1,4 @@
+#!/bin/bash
+#script to rebuild octodns quickly
+
+octodns-sync --config-file=./config/production.yaml --doit
--- a/doit.txt
+++ b/doit.txt
@@ -1,4 +1,0 @@
-#!/bin/bash
-#script to rebuild octodns quickly
-
-octodns-sync --config-file=./config/production.yaml --doit
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -352,8 +352,8 @@ class AzureProvider(BaseProvider):
 
     def _data_for_PTR(self, azrecord):
         try:
-            ptdrname = azrecord.ptr_records[0].ptdrname
-            return {'value': _check_endswith_dot(ptdrname)}
+            ptrdname = azrecord.ptr_records[0].ptrdname
+            return {'value': _check_endswith_dot(ptrdname)}
         except:
             return {'value': '.'}
 
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -8,54 +8,66 @@ import sys
 
 from azure.common.credentials import ServicePrincipalCredentials
 from azure.mgmt.dns import DnsManagementClient
-from azure.mgmt.dns.models import *
+from azure.mgmt.dns.models import ARecord, AaaaRecord, CnameRecord, MxRecord, \
+    SrvRecord, NsRecord, PtrRecord, TxtRecord, Zone
 
-from collections import defaultdict
-# from incf.countryutils.transformations import cca_to_ctca2 TODO: add geo sup.
+from functools import reduce
+
 import logging
 import re
 from ..record import Record, Update
 from .base import BaseProvider
 
-
-#TODO: changes made to master include adding /build, Makefile to .gitignore and 
-# making Makefile.
-# Only made for A records. will have to adjust for more generic params types
 class _AzureRecord(object):
-    def __init__(self, resource_group, record, values=None, ttl=1800):
-        # print('Here4',file=sys.stderr)
+    ''' 
+        Wrapper for OctoDNS record.
+    azuredns.py:
+        class: octodns.provider.azuredns._AzureRecord
+        An _AzureRecord is easily accessible to the Azure DNS Management library
+        functions and is used to wrap all relevant data to create a record in 
+        Azure.
+    '''
+    
+    def __init__(self, resource_group, record, values=None):
+        ''' 
+            :param resource_group: The name of resource group in Azure
+            :type  resource_group: str 
+            :param record: An OctoDNS record 
+            :type  record: ..record.Record
+            :param values: Parameters for a record. eg IP address, port, domain
+                           name, etc. Values usually read from record.data
+            :type  values: {'values': [...]} or {'value': [...]}
+            
+            :type return: _AzureRecord
+        '''
         self.resource_group = resource_group
-        self.zone_name = record.zone.name[0:len(record.zone.name)-1] # strips last period
+        self.zone_name = record.zone.name[0:len(record.zone.name)-1]
         self.relative_record_set_name = record.name or '@'
         self.record_type = record._type
         
-        type_name = '{}records'.format(self.record_type).lower()
+        data = values or record.data
+        format_u_s = '' if record._type == 'A' else '_'
+        key_name ='{}{}records'.format(self.record_type, format_u_s).lower()
         class_name = '{}'.format(self.record_type).capitalize() + \
                      'Record'.format(self.record_type)
-        if values == None:
-            return
-        # TODO: clean up this bit.
-        data = values or record.data #This should fail if it gets to record.data? It only returns ttl. TODO
-        
-        #depending on mult values or not
-        #TODO: import explicitly. eval() uses for example ARecord from azure.mgmt.dns.models
-        self.params = {}
-        try:
-            self.params = {'ttl':record.ttl or ttl, \
-               type_name:[eval(class_name)(ip) for ip in data['values']] or []}
-        except KeyError: # means that doesn't have multiple values but single value
-            self.params = {'ttl':record.ttl or ttl, \
-               type_name:[eval(class_name)(data['value'])] or []}
+        self.params = None
+        if not self.record_type == 'CNAME':
+            self.params = self._params(data, key_name, eval(class_name))
+        else:
+            self.params = {'cname_record': CnameRecord(data['value'])}
+        self.params['ttl'] = record.ttl
 
+    def _params(self, data, key_name, azure_class):
+        return {key_name: [azure_class(v) for v in data['values']]} \
+            if 'values' in data else {key_name: [azure_class(data['value'])]}
 
-    
 
 class AzureProvider(BaseProvider):
     '''
     Azure DNS Provider
-    
-    azure.py:
-        class: octodns.provider.azure.AzureProvider
+
+    azuredns.py:
+        class: octodns.provider.azuredns.AzureProvider
         # Current support of authentication of access to Azure services only 
         # includes using a Service Principal: 
         # https://docs.microsoft.com/en-us/azure/azure-resource-manager/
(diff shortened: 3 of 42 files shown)


doit.txt
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
ross: Please remove this file.

.gitignore
```diff
@@ -9,3 +9,7 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
+build/
+config/
```
ross: `build/` and `config/` entries probably make sense, but please remove the latter 2.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+            :param record: An OctoDNS record
+            :type  record: ..record.Record
+            :param values: Parameters for a record. eg IP address, port, domain
+                           name, etc. Values usually read from record.data
+            :type  values: {'values': [...]} or {'value': [...]}
+
+            :type return: _AzureRecord
+        '''
+        self.resource_group = resource_group
+        self.zone_name = record.zone.name[0:len(record.zone.name) - 1]
```
ross: The `0` can be omitted here.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+        class_name = '{}'.format(self.record_type).capitalize() + \
+                     'Record'.format(self.record_type)
+
+        self.params = getattr(self, '_params_for_{}'.format(record._type))
+        self.params = self.params(record.data, key_name, eval(class_name))
+        self.params['ttl'] = record.ttl
+
+    def _params(self, data, key_name, azure_class):
+        return {key_name: [azure_class(v) for v in data['values']]} \
+            if 'values' in data else {key_name: [azure_class(data['value'])]}
```
ross: The multi-line if/else here is a little hard to follow. Might be cleaner to do a traditional if/else rather than a one-liner. 

Fwiw, most of the providers do this by having different "_params" functions for types with `values` than `value`. Not a problem, just an option to consider.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+        self.resource_group = resource_group
+        self.zone_name = record.zone.name[0:len(record.zone.name) - 1]
+        self.relative_record_set_name = record.name or '@'
+        self.record_type = record._type
+
+        if delete:
+            return
+
+        format_u_s = '' if record._type == 'A' else '_'
+        key_name = '{}{}records'.format(self.record_type, format_u_s).lower()
```
ross: If you don't mind can you drop a comment in here about what `key_name` is for. Not knowing Azure I'm not sure if it's something required by its api or if it's an implementation detail of your provider.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+                                      data['value']['exchange']))
+        return {key_name: params}
+
+    def _params_for_CNAME(self, data, key_name, azure_class):
+        return {'cname_record': CnameRecord(data['value'])}
+
+    def _equals(self, b):
+        '''Checks whether two records are equal by comparing all fields.
+            :param b: Another _AzureRecord object
+            :type  b: _AzureRecord
```
ross: Would be nice to make a full pass of doing formal `__doc__` like this across all of octoDNS, ... someday.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+            if char != 'ttl':
+                try:
+                    for rec in self.params[char]:
+                        string += 'Record: {}; '.format(rec.__dict__)
+                except:
+                    string += 'Record: {}; '.format(self.params[char].__dict__)
+        return string
+
+
+def _validate_per(string):
```
ross: The name of this function when I ran across it down below didn't really tell me what it does. Maybe something like `_validate_trailing_dot`?

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+        # The Azure Active Directory Application ID (aka client ID) req:
+        client_id:
+        # Authentication Key Value req:
+        key:
+        # Directory ID (referred to tenant ID) req:
+        directory_id:
+        # Subscription ID req:
+        sub_id:
+        # Resource Group name req:
+        resource_group:
```
ross: Please go ahead and add example/recommended values here. Any "secrets"/"auth" stuff should probably use `env` and the rest can be whatever makes sense.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+
+        TODO: change config file to use env vars instead of hard-coded keys
+
+        personal notes: testing: test authentication vars located in
+                /home/t-hehwan/vars.txt
```
ross: Probably best to remove this sort of comment.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+    '''
+    SUPPORTS_GEO = False
+    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'MX', 'NS', 'PTR', 'SRV', 'TXT'))
```
ross: Cool. Hopefully it wasn't too much trouble to merge in all the changes that landed last week.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+            :type return: str or None
+        '''
+        self.log.debug('_check_zone: name=%s', name)
+        try:
+            if name in self._azure_zones:
+                return name
+            self._dns_client.zones.get(self._resource_group, name)
+            self._azure_zones.add(name)
+            return name
+        except:
```
ross: This should catch a specific exception type that has to do with the zone not existing rather than all possible exceptions. Hopefully the Azure client lib has clean/specific does not exist exception, but if not there's probably :crossed_fingers: a generic client exception that's throw where you can then look at the exception message.  

Catching any exception like this will hide all sort of problems and make tracking down issues harder for users. E.g. if the problem is an auth error it'll take both the get & create calls before raising. This would even catch problems in the code itself, `KeyError`, `ValueError`, ... anything, and try and still try and create the zone.

For a more specific problematic case an intermittent error (network, azure hiccup etc.) that fails the fetch in a call where `create=True` would then try and create a zone that probably already exists. That'd be confusing and potentially scary for a user.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+                raise
+
+    def populate(self, zone, target=False, lenient=False):
+        '''
+            Required function of manager.py.
+
+            Special notes for Azure.
+            Azure zone names omit final '.'
+            Azure root records names are represented by '@'. OctoDNS uses ''
+            Azure records created through online interface may have null values
```
ross: Interesting. That's case I haven't tested, but octoDNS might let you set that up with `values: []`. No clue what all the providers would do with it.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+            :type  zone: octodns.zone.Zone
+            :param target: Checks if Azure is source or target of config.
+                           Currently only supports as a target. Does not use.
+            :type  target: bool
+            :param lenient: Unused. Check octodns.manager for usage.
+            :type  lenient: bool
+
+            :type return: void
+        '''
+        zone_name = zone.name[0:len(zone.name) - 1]
```
ross: Another `0` that can be omitted.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+        self.log.debug('populate: name=%s', zone_name)
```
ross: Probably best to log the octoDNS's zone.name rather than Azure here for consistency sake.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,369 @@
...
+            :param azrecord: a return of a call to list azure records
+            :type  azrecord: azure.mgmt.dns.models.RecordSet
+
+            :type  return: dict
+
+            CNAME and PTR both use the catch block to catch possible empty
+            records. Refer to population comment.
+        '''
+        try:
+            return {'value': _validate_per(azrecord.cname_record.cname)}
```
ross: Some of the changes made last week made the process of pulling data down from providers more "lenient." This may be unneeded now. In fact cause problems when users do a `dump --lenient` which is now intended to pull whatever the provider has, valid as far as octoDNS is concerned or not.

h-hwang: This is a special case for CNAME and PTR which are essentially stored as record objects instead of lists like the other records are. It handles the odd case where a record on Azure has an empty value and so won't pass the' _validate_per' (now renamed) function since it becomes None. I believe this still works with --lenient but will add more test cases to sure.

h-hwang:
Latest commit covers everything you mentioned and adds some more test cases. Thanks for catching the issue on the try except

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,423 @@
...
+            for vals in data['values']:
+                params.append(azure_class(vals['preference'],
+                                          vals['exchange']))
+        else:  # Else there is a singular data point keyed by 'value'.
+            params.append(azure_class(data['value']['preference'],
+                                      data['value']['exchange']))
+        return {key_name: params}
+
+    def _params_for_CNAME(self, data, key_name, azure_class):
+        return {'cname_record': CnameRecord(data['value'])}
```
ross: Looks like this returns a CnameRecord, which doesn't line up with what the other `_params_for_*` functions do. That intentional?

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,423 @@
...
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from azure.common.credentials import ServicePrincipalCredentials
+from azure.mgmt.dns import DnsManagementClient
+from azure.mgmt.dns.models import ARecord, AaaaRecord, CnameRecord, MxRecord, \
+    SrvRecord, NsRecord, PtrRecord, TxtRecord, Zone
```
ross: Several of these imports are unused and failing the ci lint process. You can run what CI runs by doing `./script/cibuild`

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,423 @@
...
+        ar = _AzureRecord(self._resource_group, change.new)
+        create = self._dns_client.record_sets.create_or_update
+
+        create(resource_group_name=ar.resource_group,
+               zone_name=ar.zone_name,
+               relative_record_set_name=ar.relative_record_set_name,
+               record_type=ar.record_type,
+               parameters=ar.params)
+
+        print('*  Success Create/Update: {}'.format(ar), file=sys.stderr)
```
ross: Print should be removed and replaced with call to `self.log.debug` if it's information that's important to surface. Other providers currently don't emit messages for creates, but some underlying client libraries do when `--debug` is enabled.

octodns/provider/azuredns.py
```diff
@@ -0,0 +1,423 @@
...
+    _apply_Update = _apply_Create
+
+    def _apply_Delete(self, change):
+        ar = _AzureRecord(self._resource_group, change.existing, delete=True)
+        delete = self._dns_client.record_sets.delete
+
+        delete(self._resource_group, ar.zone_name, ar.relative_record_set_name,
+               ar.record_type)
+
+        print('*  Success Delete: {}'.format(ar), file=sys.stderr)
```
ross: Same as above.

octodns/provider/azuredns.py
h-hwang: Could you advice on the best way to resolve this? These imports are actually being used implicitly in line 72:  `self.params = self.params(record.data, key_name, eval(class_name))`
where `class_name` is one of the record classes. Does pyflakes support an ignore line comment and if so, would that be acceptable to you to use stylistically?

octodns/provider/azuredns.py
h-hwang: Yes, sorry the other comment I wrote on the import statement is related. They all return a dictionary with a (key,value) pointing to an azure record. aka CnameRecord, ARecord, etc. For style I can instead change `CnameRecord` to `azure_class` for consistency

octodns/provider/azuredns.py
ross: Oh. I didn't catch the `eval` there. I'd suggest avoiding the eval in the first place and doing a `record_type` to `class` map similar to `Record.new`, [e.g.](https://github.com/github/octodns/blob/master/octodns/record.py#L80-L107). That will be explicit and traceable and will also let you get rid of the `class_name` format since you can map directly from `_type` to the supporting class.

At the [class level](https://github.com/github/octodns/pull/84/files#diff-895e411cb9ddf99d0770162b097d381dR43).

```python
TYPE_MAP = {
    'A': ARecord,
...
}
```

And then rather than the format & eval you can do

```python
self.params = self.params(record.data, key_name, self.TYPE_MAP[self.record_type])
```

h-hwang: Oh great I've changed it to what you recommended

requirements.txt
```diff
@@ -16,3 +16,6 @@ python-dateutil==2.6.0
 requests==2.13.0
 s3transfer==0.1.10
 six==1.10.0
+azure-mgmt-dns==1.0.1
+azure-common==1.1.6
+msrestazure==0.4.10
```
ross: Please alphabetize the additions in with everything else.

h-hwang:
Hey Ross the latest commit did the alphabetization and wrote to README. I just merged into it the latest master branch. Is there anything else you're seeing that you'd like me to do?

ross:
> Hey Ross the latest commit did the alphabetization and wrote to README. I just merged into it the latest master branch. Is there anything else you're seeing that you'd like me to do?

Don't think so. Will look at merging when I get the chance to focus on it for a bit.

PR #88, opened 2017-07-05
Remove ; escapes before sending to ns1 and when pulling from

ross:
So apparently NS1 doesn't want `;` to be escaped in its API/UI. This PR attempts to comply. So far it's the only provider I'm aware of that that doesn't adhere to the original RFCs for the values, some make it optional, others require it. 

Still need to add tests.

/cc blakestoddard
/cc https://github.com/github/octodns/issues/79


ross:
OK. I think this should work now :crossed_fingers:

blakestoddard:
Verified using `-e git+ssh://git@github.com/github/octodns.git@818c1e9cc6ba6d155d43a4f3152615d48dabe091#egg=octodns` and everything looks good here!

PR #90, opened 2017-07-10
fix unsafe plan text interpolation

joewilliams:


ross:
Fwiw if you want to verify the format is working you can do something like the following in one of the places we're testing `UnsafePlan`.

```python
        with self.assertRaises(UnsafePlan) as ctx:
            Plan(zone, zone, changes).raise_if_unsafe()
        self.assertEquals('<msg>', ctx.exception.message)
        # or
        self.assertTrue('<bit of msg>' in ctx.exception.message)
```

PR #97, opened 2017-08-12
Add support for increasing Route53 retries

ross:
Allow setting of Route53 retries with `Route53Provider.client_max_attempts`. Most likely case where this would be needed would be throttling limits. This will require `botocore>=1.6.0` so requirements.txt versions have been bumped. Users who don't need `client_max_attempts` can continue to use older versions, but updating is recommended. The default currently seems to be 4. 

```yaml
  route53:
    class: octodns.provider.route53.Route53Provider
    access_key_id: env/DEV_AWS_ACCESS_KEY_ID
    secret_access_key: env/DEV_AWS_SECRET_ACCESS_KEY
    conn_max_attempts: 10
```

Fixes https://github.com/github/octodns/issues/96 /cc @dolvany 
/cc https://github.com/boto/botocore/pull/1260 which added `max_attempts` support


brianeclow:
**NOTE:*** The example should be:
```
  route53:
    class: octodns.provider.route53.Route53Provider
    access_key_id: env/DEV_AWS_ACCESS_KEY_ID
    secret_access_key: env/DEV_AWS_SECRET_ACCESS_KEY
    client_max_attempts: 10
```

Issue #98, opened 2017-08-15
Allow use of https for the PowerDNS provider


poblahblahblah:
Hello,

We would like to use https to secure the connection between octodns and powerdns, but according to https://github.com/github/octodns/blob/master/octodns/provider/powerdns.py#L34 this appears to be "hard coded" to http. Is this a change you would consider? If so, I can get a PR going for making this a configurable option, with http being the default.

ross:
> Is this a change you would consider? If so, I can get a PR going for making this a configurable option, with http being the default.

Definitely. Should hopefully be as simple as adding an optional parameter `scheme` or something like that. Happy to take a look at a PR. Thanks.

PR #99, opened 2017-08-15
Allow proto to be specified for powerdns

poblahblahblah:
This allows a protocol to be set for the PowerDNS API. It defaults to
http to retain backwards compatibility.

### DIFF
--- a/octodns/provider/powerdns.py
+++ b/octodns/provider/powerdns.py
@@ -18,11 +18,13 @@ class PowerDnsBaseProvider(BaseProvider):
                     'SPF', 'SSHFP', 'SRV', 'TXT'))
     TIMEOUT = 5
 
-    def __init__(self, id, host, api_key, port=8081, *args, **kwargs):
+    def __init__(self, id, host, api_key, port=8081, scheme="http", *args,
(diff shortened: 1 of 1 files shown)


poblahblahblah:
Addresses #98

octodns/provider/powerdns.py
```diff
@@ -18,11 +18,13 @@ class PowerDnsBaseProvider(BaseProvider):
                     'SPF', 'SSHFP', 'SRV', 'TXT'))
     TIMEOUT = 5
 
-    def __init__(self, id, host, api_key, port=8081, *args, **kwargs):
+    def __init__(self, id, host, api_key, port=8081, scheme="https", *args,
```
ross: Looks like the default here is https rather than http.

poblahblahblah: whoops

poblahblahblah: Fixed and pushed - sorry about that. I guess I got too excited about https?

ross:
Few things I want to land before cutting a new release. You should be able to grab this one with a pip git url until then. Thanks again.

Issue #62, opened 2017-06-09
CNAME support in GeoIP 


vanbroup:
I have the following config, www (A) and www3 (AAAA) are created but www2 (CNAME) is ignored as [described in the documentation](https://github.com/github/octodns/blob/master/docs/records.md#geodns-support).

```
www:
  geo:
    AS:
      - 1.1.1.1
    AS-JP:
      - 3.3.3.3
    EU:
      - 2.2.2.2
    EU-NL:
      - 5.5.5.5
    NA-US-CA:
      - 4.4.4.4
  ttl: 300
  type: A
  value: 1.2.3.4
www2:
  geo:
    AP:
      - cdn.asia.
    EU:
      - cdn.eu.
    JP:
      - cdn.jp.
    NA-US-CA:
      - cdn.com.
    NL:
      - cdn.nl.
  ttl: 300
  type: CNAME
  value: cdn.net.
www3:
  geo:
    AS:
      - 2001:DB8::2
    AS-JP:
      - 2001:DB8::3
    EU:
      - 2001:DB8::4
    EU-NL:
      - 2001:DB8::5
    NA-US-CA:
      - 2001:DB8::6
  ttl: 300
  type: AAAA
  value: 2001:DB8::1
```
Route53:
![image](https://user-images.githubusercontent.com/722070/26975279-63b951fa-4d1f-11e7-9a74-bfdac49f0c4a.png)

Same for Dyn.

ross:
Geo support was only added to `A` & `AAAA` records as those were the types that had specific use cases, I was able to dig into and figure out exactly how they should work, and test. Without having the geo support the `geo` section is just ignored in general. 

We have `CNAME`s pointed to geo `A`/`AAAA` that we own and we have `CNAME`s pointed to `A`/`AAAA` that external providers own and do anycast/geo with. In your example above are you pointing to different cdn providers in different regions or just different endpoints for the same company?

Adding `CNAME` should look relatively similar to `A`/`AAAA`. The biggest difference is going to be that it's single-value rather than multi-values so it's not trivial. The other part that wasn't immediately clear is what health-checking should look like on a `CNAME` geo record. I can't see any reason why the same checks wouldn't make sense, but without having a specific use case laid out it's hard to evaluate.

vanbroup:
You are right, my example is showing a simple region based CDN selection. This should be quite simple to implement.

[...]

PR #100, opened 2017-08-16
Features# Geo support added for CNAME Records. ISSUE - #62

ashishprajapati:
@ross , I have implemented the Geo support for the cname records. Below, you can find out the example configuration snippet that will work after this update. 

```
www2:
  geo:
    AS:
      - cdn.asia.
    EU:
      - cdn.eu.
    NA-US-CA:
      - cdn.com.
  ttl: 300
  type: CNAME
  value: cdn.net.
```

Please, have a look and let me know if anything.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -92,7 +92,7 @@ class _Route53Record(object):
     def _values_for_value(self, record):
         return [record.value]
 
-    _values_for_CNAME = _values_for_value
+    _values_for_CNAME = _values_for_values
     _values_for_PTR = _values_for_value
 
     def _values_for_MX(self, record):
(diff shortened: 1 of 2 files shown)


ashishprajapati:
I am working on it for fixing it as some of the test cases are failing.

ross:
Hi @ashishprajapati I have an existing start on geo CNAME support in https://github.com/github/octodns/tree/geo-cname, but that work got stalled and I haven't had time to get back to it. 

One thing to keep in mind is that both Dyn and Route53 have Geo so in order to add CNAME support it'd either need to be done in both of them (ideally) or things would have to be reworked so that Geo CNAME support is optional (less than ideal).

ashishprajapati:
Hello @ross I have seen your geo work here at https://github.com/github/octodns/tree/geo-cname  . It make sense that it should be implement for Dyn and Route53 both of them in same build. I am on it and wants to proceed further with it. Is there anything you want to suggest me?

ross:
To be honest I'm not even sure where I left that branch at. I went ahead and created a PR of it, https://github.com/github/octodns/pull/101, so we'd get CI runs etc. If I recall I think what was there was mostly working, but I wasn't super happy with how a few bits of it had turned out. I'll try and find some time to dig back into it this week and figure out the details.

ashishprajapati:
@ross had already started worked on that and it’s open since years so I have closed it now :)

PR #101, opened 2017-08-21
[WIP] Refactoring to support geo cname

ross:
- GeoMixin -> GeoValuesMixin
- GeoValueMixin added
- GeoValue -> GeoBase, GeoValue, and GeoValues
- related tests & cleanup

### DIFF
--- a/octodns/provider/dyn.py
+++ b/octodns/provider/dyn.py
@@ -325,7 +325,7 @@ class DynProvider(BaseProvider):
                 else:
                     data['values'] = ['0.0.0.0']
                 convert_value = \
-                        getattr(self, '_convert_value_{}'.format(_type))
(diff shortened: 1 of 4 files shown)


ashishprajapati:
@ross :) Are we now supporting geo cnames for Dyn and Route53 ? Or this is something, still remaining to be finished. I just reached at my PR that was raised long time ago https://github.com/github/octodns/pull/100 and you also worked around same thing so I stopped it in between. 
Thanks

ross:
> @ross :) Are we now supporting geo cnames for Dyn and Route53 ?

There's "alpha" support for DynProvider currently in master (dark shipped for now) as `dynamic` records. They work quite differently and should be more powerful in general, including support for `CNAME`s. I'm planning to start working on Route53Provider support before too long. Once that lands I'll call it "beta" and go from there :grin:

PR #102, opened 2017-08-21
Fixed TXT record bug when creating the record with Azure

h-hwang:
An Azure SDK library call was done incorrectly for TXT records. This correctly uses `TxtRecord(x)` where `x` is a list of string(s) of the txt values and not the string itself. Resulted in visual bug in Azure and possible problems with dig. API listed here: https://docs.microsoft.com/en-us/python/api/azure.mgmt.dns.models.txtrecord?view=azure-python

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -90,7 +90,6 @@ class _AzureRecord(object):
     _params_for_AAAA = _params
     _params_for_NS = _params
     _params_for_PTR = _params
-    _params_for_TXT = _params
 
     def _params_for_CNAME(self, data, key_name, azure_class):
         return {key_name: azure_class(data['value'])}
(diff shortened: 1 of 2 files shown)


octodns/provider/azuredns.py
```diff
@@ -121,6 +120,12 @@ def _params_for_SRV(self, data, key_name, azure_class):
                                       data['value']['target']))
         return {key_name: params}
 
+    def _params_for_TXT(self, data, key_name, azure_class):
+        if 'values' in data:
+            return {key_name: [azure_class([v]) for v in data['values']]}
+        else:  # API for TxtRecord has list of str, even for singleton
+            return {key_name: [azure_class([data['value']])]}
```
ross: Might be slightly cleaner to do something like the following to separate the parsing/marshaling bit from where the return value is being built.

```python
try:
    values = data['values']
except KeyError:
    values = [data['value']]
return {key_name: [azure_class([v]) for v in values']}
```

h-hwang: I like your suggestion here. There is a similar structure on line 83 as well. Let me push a quick change then!

PR #103, opened 2017-08-26
CAA support

ross:
This PR adds `CAA` record type support to octoDNS and a subset of providers. Current plans state that as of September CAs will require CAA record support - https://cabforum.org/pipermail/public/2017-March/009988.html.

- Spec - https://tools.ietf.org/html/rfc6844
- Wiki - https://en.wikipedia.org/wiki/DNS_Certification_Authority_Authorization

* [ ] Azure - Don't have access /cc @h-hwang CAA support planned?
* [x] CloudFlare beta support for `CAA`, I don't have access yet, also seems to restrict `tag` options in ways that don't seem to line up with the spec/other providers (reaching out)
* [x] DNSimple supports `CAA` but has restrictions on `tag` options that don't seem to line up with the spec/other providers (reaching out)
* [x] Dyn 
* [x] NSOne doesn't support `CAA` records (reaching out)
* [x] PowerDNS
* [x] Route53

Provider unit tests are a WIP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,12 +150,12 @@ The above command pulled the existing data out of Route53 and placed the results
 | Provider | Record Support | GeoDNS Support | Notes |
 |--|--|--|--|
 | [AzureProvider](/octodns/provider/azuredns.py) | A, AAAA, CNAME, MX, NS, PTR, SRV, TXT | No | |
-| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CNAME, MX, NS, SPF, TXT | No | |
-| [DnsimpleProvider](/octodns/provider/dnsimple.py) | All | No | |
+| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CAA, CNAME, MX, NS, SPF, TXT | No | CAA tags restricted |
+| [DnsimpleProvider](/octodns/provider/dnsimple.py) | All | No | CAA tags restricted |
 | [DynProvider](/octodns/provider/dyn.py) | All | Yes | |
 | [Ns1Provider](/octodns/provider/ns1.py) | All | No | |
 | [PowerDnsProvider](/octodns/provider/powerdns.py) | All | No | |
-| [Route53](/octodns/provider/route53.py) | A, AAAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, TXT | Yes | |
+| [Route53](/octodns/provider/route53.py) | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, TXT | Yes | |
 | [TinyDNSSource](/octodns/source/tinydns.py) | A, CNAME, MX, NS, PTR | No | read-only |
 | [YamlProvider](/octodns/provider/yaml.py) | All | Yes | config |
 
--- a/octodns/record.py
+++ b/octodns/record.py
@@ -81,29 +81,16 @@ class Record(object):
                 'A': ARecord,
                 'AAAA': AaaaRecord,
                 'ALIAS': AliasRecord,
-                # cert
+                'CAA': CaaRecord,
                 'CNAME': CnameRecord,
-                # dhcid
-                # dname
-                # dnskey
-                # ds
-                # ipseckey
-                # key
-                # kx
-                # loc
                 'MX': MxRecord,
                 'NAPTR': NaptrRecord,
                 'NS': NsRecord,
-                # nsap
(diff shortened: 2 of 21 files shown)


ross:
> NSOne doesn't support CAA records (reaching out)

Rolled out support 1w ago - https://ns1.com/blog/ns1s-managed-and-dedicated-dns-supports-caa-records. https://github.com/github/octodns/pull/103/commits/22591ae84b80812b0f01bc3c7c7fc12b18310869 added octoDNS support.

README.md
```diff
@@ -150,12 +150,12 @@ The above command pulled the existing data out of Route53 and placed the results
 | Provider | Record Support | GeoDNS Support | Notes |
 |--|--|--|--|
 | [AzureProvider](/octodns/provider/azuredns.py) | A, AAAA, CNAME, MX, NS, PTR, SRV, TXT | No | |
-| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CNAME, MX, NS, SPF, TXT | No | |
-| [DnsimpleProvider](/octodns/provider/dnsimple.py) | All | No | |
+| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CAA, CNAME, MX, NS, SPF, TXT | No | CAA tags restricted |
+| [DnsimpleProvider](/octodns/provider/dnsimple.py) | All | No | CAA tags restricted |
 | [DynProvider](/octodns/provider/dyn.py) | All | Yes | |
 | [Ns1Provider](/octodns/provider/ns1.py) | All | No | |
```
ptoomey3: Is this table meant to reflect support for a record type by octodns or by the provider? In other words, If NS1 doesn't actually support CAA yet, should this be updated to say something like `All (except CAA)`?

octodns/record.py
```diff
@@ -398,6 +385,61 @@ def _process_value(self, value):
...
 
+class CaaValue(object):
+    # https://tools.ietf.org/html/rfc6844#page-5
+
+    @classmethod
+    def _validate_value(cls, value):
+        reasons = []
+        try:
+            flags = int(value.get('flags', 0))
+            if flags not in (0, 128):
```
ptoomey3: How come the limit to 128? [The spec](https://tools.ietf.org/html/rfc6844#section-5.1.1) and [this higher level overview](https://support.dnsimple.com/articles/caa-record/) both list `flags` as ranging from 0-255.

README.md
ross: They do support it, just in their v2 portal (which I didn't know existed.) A subseuqnt commit added support to `Ns1Provider`.

octodns/record.py
ross: All of the providers I specifically tried seemed to restrict the flag to 0 or 128. Tbh, I'm tempted to remove the restrictions/validations from octoDNS and just let the providers do what they do as the implementation seems to be really varied. Some allow anything, which seems to agree with the spec, others only allow 0/128 and the 3 "currently" specified `tag` values. 

The downside to not validating in octoDNS is that dry-runs will look like they're going to work, but fail when you go to `--doit`. 

Since the provider validation is already all over the place that's basically unavoidable, thus my inclination to just let anything through.

ptoomey3: Makes sense..maybe once the spec is more widely deployed there will be more consistency/sanity checks that can be added. So..yeah..maybe it makes sense to just limit it to 0-255 (i.e. basically no validation other than it will fit in a byte as is required for the record response format specified [in the rfc](https://tools.ietf.org/html/rfc6844#section-5.1).).

ross:
> Just curious..what are the restrictions?

I don't know exactly what Dnsimple allows since it's not a drop-down. You can enter anything into the `tag` field, but when you try to "Add Record" get the following:

screen shot 2017-08-28 at 8 09 23 am

Cloudflare has a dropdown with 3 options and doesn't allow you to set the flags/"Issuer Critical":

screen shot 2017-08-28 at 8 10 29 am

Dyn, NS1, PowerDNS seem to allow most anything. Route53 allows arbitrary tags, but not flags.

ross:
> Azure - Don't have access /cc @h-hwang CAA support planned?

We can get this one is a subsequent PR if it's supported. For now moving on with this one so that our security team can put them in place when they're ready.

h-hwang:
Hey @ross Azure does not currently support CAA records. It can be prepared in a PR pretty modularly when it does.

ross:
> Hey @ross Azure does not currently support CAA records. It can be prepared in a PR pretty modularly when it does.

Thanks. Sounds good.

PR #109, opened 2017-09-19
Route53: Allow region to be specified

poblahblahblah:
The branch name is a little misleading since I started this with the goal of adding the (yet to be in existence) cn-north-1 region for AWS China, but quickly realized botocore abstracted most of the work that would need to go into that.

This will allow the user to specify the region they want to use for Route53. The default region appears to be us-east-1 and while updating the tests I noticed that the canned response for the tests was indeed us-east-1.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -233,8 +233,8 @@ class Route53Provider(BaseProvider):
     # health check config.
     HEALTH_CHECK_VERSION = '0000'
 
-    def __init__(self, id, access_key_id, secret_access_key, max_changes=1000,
-                 client_max_attempts=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


ross:
Not sure if passing region here has an effect. Based on the endpoints doc, search for "Amazon Route 53" on http://docs.aws.amazon.com/general/latest/gr/rande.html, there's only a single endpoint `route53.amazonaws.com` in `us-east-1`. It's "Global" in the UI.

poblahblahblah:
You are correct that right now it does nothing. If you try to pass `cn-north-1` it will try to reach out to AWS China, but then it just fails since the service does not exist there yet, although I would imagine it will show up eventually.

Since it does nothing right now I guess we can close it and revisit if/when Route 53 is spun up in AWS China.

Issue #26, opened 2017-04-30
Handle R53 'Alias' Records


michaelmcallister:
Route53 Alias records are kinda like CNAME records, but you can have them at the Apex, which as you can imagine isn't going to translate between providers very well.

They also have a different response for [list-resource-record-sets](https://docs.aws.amazon.com/cli/latest/reference/route53/list-resource-record-sets.html) (example below)

```
{
    "ResourceRecordSets": [
        {
            "AliasTarget": {
                "HostedZoneId": "Z1EXAMPLE",
                "EvaluateTargetHealth": false,
                "DNSName": "s3-website-ap-southeast-2.amazonaws.com."
            },
            "Type": "A",
            "Name": "example.com."
        },
        {
            "ResourceRecords": [
                {
                    "Value": "127.0.0.1"
                }
            ],
            "Type": "A",
            "Name": "test.example.com.",
            "TTL": 300
        },
```
See the [AWS Documentation](https://aws.amazon.com/route53/faqs/#which_dns_records_are_supported) for more info on Alias records

octodns doesn't handle this gracefully, and will throw a ``KeyError``

```
(env) ☁  dns  octodns-dump --config-file=./config/production.yaml --output-dir=./tmp example.com. route53 --debug
2017-04-30T20:17:19  [139951851439872] INFO  Manager __init__: config_file=./config/production.yaml
2017-04-30T20:17:19  [139951851439872] DEBUG Manager __init__:   configuring providers
2017-04-30T20:17:19  [139951851439872] DEBUG Route53Provider[route53] __init__: id=route53, access_key_id=***, secret_access_key=***
2017-04-30T20:17:19  [139951851439872] DEBUG Route53Provider[route53] __init__: id=route53, apply_disabled=False
2017-04-30T20:17:19  [139951851439872] INFO  Manager dump: zone=example.com., sources=()
2017-04-30T20:17:19  [139951851439872] DEBUG YamlProvider[dump] __init__: id=dump, directory=./tmp, default_ttl=3600
2017-04-30T20:17:19  [139951851439872] DEBUG YamlProvider[dump] __init__: id=dump, apply_disabled=False
2017-04-30T20:17:19  [139951851439872] DEBUG Manager configured_sub_zones: subs=[]
2017-04-30T20:17:19  [139951851439872] DEBUG Zone __init__: zone=Zone<example.com.>, sub_zones=set([])
[...]

PR #111, opened 2017-09-28
Add support for included and excluded records

vanbroup:
`Included` and `Excluded` can be used to filter records for one or more specific provider(s). This can be extremely useful when certain record types are not supported by a provider and you want only that provider to receive an alternative record.

See also: https://github.com/github/octodns/issues/26

### DIFF
--- a/octodns/zone.py
+++ b/octodns/zone.py
@@ -110,9 +110,28 @@ class Zone(object):
         for record in filter(_is_eligible, self.records):
             if record.ignored:
                 continue
+            elif len(record.included) > 0 and \
+                    target.__class__.__name__ not in record.included:
+                self.log.debug('changes:  skipping record=%s %s - %s not'
+                               ' included ', record.fqdn, record._type,
+                               target.id)
+                continue
+            elif target.__class__.__name__ in record.excluded:
+                self.log.debug('changes:  skipping record=%s %s - %s '
+                               'excluded ', record.fqdn, record._type,
+                               target.id)
+                continue
             try:
                 desired_record = desired_records[record]
                 if desired_record.ignored:
+                    continue
+                elif len(record.included) > 0 and \
+                        target.__class__.__name__ not in record.included:
+                    self.log.debug('changes:  skipping record=%s %s - %s'
+                                   'not included ', record.fqdn, record._type,
+                                   target.id)
+                    continue
+                elif target.__class__.__name__ in record.excluded:
                     continue
             except KeyError:
                 if not target.supports(record):
@@ -141,6 +160,18 @@ class Zone(object):
         for record in filter(_is_eligible, desired.records - self.records):
             if record.ignored:
                 continue
+            elif len(record.included) > 0 and \
+                    target.__class__.__name__ not in record.included:
+                self.log.debug('changes:  skipping record=%s %s - %s not'
+                               ' included ', record.fqdn, record._type,
+                               target.id)
+                continue
+            elif target.__class__.__name__ in record.excluded:
+                self.log.debug('changes:  skipping record=%s %s - %s '
+                               'excluded ', record.fqdn, record._type,
+                               target.id)
+                continue
+
             if not target.supports(record):
                 self.log.debug('changes:  skipping record=%s %s - %s does not '
                                'support it', record.fqdn, record._type,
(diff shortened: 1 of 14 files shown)


octodns/zone.py
```diff
@@ -110,10 +110,29 @@ def changes(self, desired, target):
         for record in filter(_is_eligible, self.records):
             if record.ignored:
                 continue
+            elif len(record.included) > 0 and \
+                    target.__class__.__name__ not in record.included:
```
ross: What do you think about using `target.id`?

In cases where you have multiple instances of a single provider you'd be able to control which one(s) something should be included or excluded by with finer granularity. You could always get back to it being a provider-based thing by listing multiple ids (all the ids for the class.) I'd imagine most configs will just have one instance of a given provider class, but at least internally with PowerDNS we have many. It also seems a little more consistent to talk in terms of `id` since that's what's used in `sources` and `targets`.

vanbroup: Actually I did not like `target.__class__.__name__` but wasn't aware that we could use simply class.id :-). Thanks for the review, update on its way!

ross:
This generally looks good and seems close to ready to merge. Looks like there's a few uncovered branches in Zone tests:

screen shot 2017-10-11 at 12 37 10 pm

`./script/coverage` should create the files under `htmlcov/`

vanbroup:
Updated PR with with a fix and more tests to cover all code

vanbroup:
It did indeed :-)

mvermaes:
Is it possible to use the changes from this PR to manage a Route 53 zone that has an apex `ALIAS` like in the initial comment in #26 (ie. by having OctoDNS skip these records, but manage the remainder of the zone)?

I tried a few combinations of `excluded`/`ignored`, but now I'm thinking that I misunderstood what this PR is enabling.

vanbroup:
ALIAS records for route53 are not supported by OctoDNS as the implementation is completely different than how other providers have implemented. With R53 the target must be if the same type and for the Apex can't be a CNAME as the same rules apply. This makes the R53 ALIAS record useless for this pupose.

mvermaes:
Thanks @vanbroup, I see what you mean about the Route 53 ALIAS having a different meaning from the ALIAS record in the other providers.

Our use case is that we have a handful of non-OctoDNS managed Route 53 zones that already have apex ALIAS records. We'd like to manage the *non-apex* records for these zones with OctoDNS.

With the code from your PR, I can exclude/skip a non-apex record:

```
host:
  octodns:
    excluded:
    - route53
  type: xxx
  value: yyy
```

I had thought I would be able to do something similar to exclude the apex record, but I still receive the KeyError from https://github.com/github/octodns/issues/26#issue-225318225. I think this is because OctoDNS' route53 provider doesn't know how to interpret these records, which I realize now is not what this PR was designed to address.

I'll continue to use the workaround from https://github.com/github/octodns/pull/27 to handle skipping these records.

mvermaes:
I'm running from master, but was using 0.8.8 previously.

I had a look through the history, and from what I can understand when the changes were made to ignore unimplemented provider types (ce0b58d5 and 852c1013), the route53 provider wasn't updated to make use of this.

The route53 provider only ignores the SOA type - https://github.com/github/octodns/blob/master/octodns/provider/route53.py#L462. But even changing this line to:
```
if record_type not in self.SUPPORTS:
```
doesn't cause the apex ALIAS to be skipped, because the type for Route 53 ALIASes is 'A' unfortunately:
```
$ aws route53 list-resource-record-sets --hosted-zone-id=XYZ
{
    "ResourceRecordSets": [
        {
            "AliasTarget": {
            ...
            }, 
            "Type": "A", 
            "Name": "XYZ"
        },
        ...
```

I think this is also why I can't use your new `exclude` functionality to skip these records.

Now that I know this is already meant to be working, I should probably create a new issue(s) about it. Rather than continuing to bother you here :)

mvermaes:
PR submitted :)

I checked the docs [[1]](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html)[[2]](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-alias.html) - `ALIAS` is a property that another record can have, but not a separate record type.

Checking for presence of `AliasTarget` was the approach taken in #27 to address this. @michaelmcallister, are you interested in rebasing your PR, now that I think it has a good chance to be merged? If not, I'm happy to have a go at it myself.

michaelmcallister:
Thank you for the heads up! Leave it with me :-)

Issue #23, opened 2017-04-27
Add support for Google Cloud DNS


theckman:
Thank you for making such an awesome utility. I could definitely see this being put to good use at my organization where DNS sprawl is definitely a challenge. Unfortunately, we're using Google Cloud DNS for some zones so we wouldn't be able to apply this utility uniformly to all zones.

It seems that the Google Cloud Platform (GCP) is gaining more adoption in the industry, so I could see others needing to integrate with their DNS solution as well. Would there be any interest in having their hosted DNS supported by this utility?

ross:
> Would there be any interest in having their hosted DNS supported by this utility?

Definitely. It's probably not something I'd get around to adding in the near-term, but PRs would be welcome. I do have a TODO item for myself to put together a doc about the best way to go about adding a new provider that would likely help. Not 100% sure when I'll get to that, hopefully soon.

anthonysutardja:
I might be able to take this on next week as a PR

artooro:
I'd use this as well if it supported Google Cloud DNS

theckman:
@anthonysutardja If you have the spare time to tackle this, I'd be happy to help with a review / pairing session.

kornface13:
I would also use the heck out of this if it supported Google CloudDNS.  :)  Thanks a ton!!!

dolvany:
I would totally use this as well!

yzguy:
If this was merged to master from #113 can we close this issue out?

PR #113, opened 2017-10-02
Add proposed google cloud provider for #23

doddo:
This adds  a proposed google cloud provider for octodns.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ output/
 tmp/
 build/
 config/
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ output/
 tmp/
 build/
 config/
+.idea/
--- a/octodns/provider/googlecloud.py
+++ b/octodns/provider/googlecloud.py
@@ -1,0 +1,363 @@
+#
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import re
+import shlex
+import time
+from logging import getLogger
+
+from google.cloud import dns
+
+from .base import BaseProvider
+from ..record import Record
+
+
+class _GoogleCloudRecordSetMaker(object):
+    """Wrapper to make google cloud client resource record sets from OctoDNS
+    Records.
+
+        googlecloud.py:
+        class: octodns.provider.googlecloud._GoogleCloudRecordSetMaker
+        An _GoogleCloudRecordSetMaker creates google cloued client resource
+        records which can be used to update the Google Cloud DNS zones.
+    """
+
+    def __init__(self, gcloud_zone, record):
+        self.gcloud_zone = gcloud_zone
+        self.record = record
+
+        self._record_set_func = getattr(
+            self, '_record_set_from_{}'.format(record._type))
+
+    def get_record_set(self):
+        return self._record_set_func(self.record)
+
+    def _record_set_from_A(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, record.values)
+
+    _record_set_from_AAAA = _record_set_from_A
+
+    def _record_set_from_CAA(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, [
+                '{flags} {tag} {value}'.format(**record.data['value'])])
+
+    def _record_set_from_CNAME(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, [record.value])
+
+    def _record_set_from_MX(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, [
+                '{preference} {exchange}'.format(**v.data)
+                for v in record.values])
+
+    def _record_set_from_NAPTR(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, [
+                '{order} {preference} "{flags}" "{service}" '
+                '"{regexp}" {replacement}'
+                .format(**v.data) for v in record.values])
+
+    _record_set_from_NS = _record_set_from_A
+
+    _record_set_from_PTR = _record_set_from_CNAME
+
+    _record_set_from_SPF = _record_set_from_A
+
+    def _record_set_from_SRV(self, record):
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, [
+                '{priority} {weight} {port} {target}'
+                .format(**v.data) for v in record.values])
+
+    def _record_set_from_TXT(self, record):
+        if 'values' in record.data:
+            val = record.data['values']
+        else:
+            val = [record.data['value']]
+
+        return self.gcloud_zone.resource_record_set(
+            record.fqdn, record._type, record.ttl, val)
+
+
+class GoogleCloudProvider(BaseProvider):
+    """
+    Google Cloud DNS provider
+
+    google_cloud:
+        class: octodns.provider.googlecloud.GoogleCloudProvider
+        # Credentials file for a service_account or other account can be
+        # specified with the GOOGLE_APPLICATION_CREDENTIALS environment
+        # variable. (https://console.cloud.google.com/apis/credentials)
+        #
+        #  The project to work on (not required)
+        # project: foobar
+    """
+
+    SUPPORTS = set(('A', 'AAAA', 'CAA', 'CNAME', 'MX', 'NAPTR',
+                    'NS', 'PTR', 'SPF', 'SRV', 'TXT'))
+    SUPPORTS_GEO = False
+
+    def __init__(self, id, project=None, *args, **kwargs):
+
+        # Logger
+        self.log = getLogger('GoogleCloudProvider[{}]'.format(id))
+        self.id = id
+
+        super(GoogleCloudProvider, self).__init__(id, *args, **kwargs)
+        self.gcloud_client = dns.Client(project=project)
+
+    def _apply(self, plan):
+        """Required function of manager.py to actually apply a record change.
+
+            :param plan: Contains the zones and changes to be made
+            :type  plan: octodns.provider.base.Plan
+
+            :type return: void
+        """
+        desired = plan.desired
+        changes = plan.changes
+
+        self.log.debug('_apply: zone=%s, len(changes)=%d', desired.name,
+                       len(changes))
+
+        # Get gcloud zone, or create one if none existed before.
+        gcloud_zone = self._get_gcloud_zone(desired.name, create=True)
+
+        gcloud_changes = gcloud_zone.changes()
+
+        for change in changes:
+            class_name = change.__class__.__name__
+            if class_name in 'Create':
+                gcloud_changes.add_record_set(
+                    self._record_to_record_set(gcloud_zone, change.record))
+            elif class_name == 'Delete':
+                gcloud_changes.delete_record_set(
+                    self._record_to_record_set(gcloud_zone, change.record))
+            elif class_name == 'Update':
+                gcloud_changes.delete_record_set(
+                    self._record_to_record_set(gcloud_zone, change.existing))
+                gcloud_changes.add_record_set(
+                    self._record_to_record_set(gcloud_zone, change.new))
+            else:
+                raise RuntimeError('Change type "{}" for change "{!s}" '
+                                   'is none of "Create", "Delete" or "Update'
+                                   .format(class_name, change))
+
+        gcloud_changes.create()
+        i = 1
+        while gcloud_changes.status != 'done':
+            self.log.debug("Waiting for changes to complete")
+            time.sleep(i)
+            gcloud_changes.reload()
+            if i < 30:
+                i += 2
+
+    def _create_gcloud_zone(self, dns_name):
+        """Creates a google cloud ManagedZone with dns_name, and zone named
+            derived from it. calls .create() method and returns it.
+
+            :param dns_name: fqdn of zone to create
+            :type  dns_name: str
+
+            :type return: new google.cloud.dns.ManagedZone
+        """
+        # Zone name must begin with a letter, end with a letter or digit,
+        # and only contain lowercase letters, digits or dashes
+        zone_name = re.sub("[^a-z0-9-]", "",
+                           dns_name[:-1].replace('.', "-"))
+        # make sure that the end result did not end up wo leading letter
+        if re.match('[^a-z]', zone_name[0]):
+            # I cannot think of a situation where a zone name derived from
+            # a domain name would'nt start with leading letter and thereby
+            # violate the constraint, however if such a situation is
+            # encountered, add a leading "a" here.
+            zone_name = "a%s" % zone_name
+
+        gcloud_zone = self.gcloud_client.zone(
+            name=zone_name,
+            dns_name=dns_name
+        )
+        gcloud_zone.create(client=self.gcloud_client)
+
+        self.log.info("Created zone %s. Fqdn %s." %
+                      (zone_name, dns_name))
+
+        return gcloud_zone
+
+    def _get_gcloud_records(self, gcloud_zone, page_token=None):
+        """ Generator function which yields ResourceRecordSet for the managed
+            gcloud zone, until there are no more records to pull.
+
+            :param gcloud_zone: zone to pull records from
+            :type gcloud_zone: google.cloud.dns.ManagedZone
+            :param page_token: page token for the page to get
+
+            :return: a resource record set
+            :type return: google.cloud.dns.ResourceRecordSet
+        """
+        gcloud_iterator = gcloud_zone.list_resource_record_sets(
+            page_token=page_token)
+        for gcloud_record in gcloud_iterator:
+            yield gcloud_record
+        # This is to get results which may be on a "paged" page.
+        # (if more than max_results) entries.
+        if gcloud_iterator.next_page_token:
+            for gcloud_record in self._get_gcloud_records(
+                    gcloud_zone, gcloud_iterator.next_page_token):
+                # yield from is in python 3 only.
+                yield gcloud_record
+
+    def _get_gcloud_zone(self, dns_name, page_token=None, create=False):
+        """Return the ManagedZone which has has the matching dns_name, or
+            None if no such zone exist, unless create=True, then create a new
+            one and return it.
+
+            :param dns_name: fqdn of dns name for zone to get.
+            :type dns_name: str
+            :param page_token: page token for the page to get
+            :type page_token: str
+            :param create: if true, create ManagedZone if it does not exist
+                           already
+
+            :type return: new google.cloud.dns.ManagedZone
+        """
+        # Find the google name for the incoming zone
+        gcloud_zones = self.gcloud_client.list_zones(page_token=page_token)
+        for gcloud_zone in gcloud_zones:
+            if gcloud_zone.dns_name == dns_name:
+                return gcloud_zone
+        else:
+            # Zone not found. Check if there are more results which could be
+            # retrieved by checking "next_page_token".
+            if gcloud_zones.next_page_token:
+                return self._get_gcloud_zone(dns_name,
+                                             gcloud_zones.next_page_token)
+            else:
+                # Nothing found, either return None or else create zone and
+                # return that one (if create=True)
+                self.log.debug('_get_gcloud_zone: zone name=%s, '
+                               'was not found by %s.',
+                               dns_name, self.gcloud_client)
+                if create:
+                    return self._create_gcloud_zone(dns_name)
+
+    @staticmethod
+    def _record_to_record_set(gcloud_zone, record):
+        """create google.cloud.dns.ResourceRecordSet from ocdodns.Record
+
+            :param record: a record object
+            :type  record: ocdodns.Record
+            :param gcloud_zone: a google gcloud zone
+            :type gcloud_zone: google.cloud.dns.ManagedZone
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ output/
 tmp/
 build/
 config/
+.idea/
```
ross: Please remove this bit as it's editor specific. You might want to :eye: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,363 @@
...
+            record.fqdn, record._type, record.ttl, [
+                '{order} {preference} "{flags}" "{service}" '
+                '"{regexp}" {replacement}'
+                .format(**v.data) for v in record.values])
+
+    _record_set_from_NS = _record_set_from_A
+
+    _record_set_from_PTR = _record_set_from_CNAME
+
+    _record_set_from_SPF = _record_set_from_A
```
ross: In other providers TXT and SPF records use the same implementation as they're relatively similar records. Would that make sense here or is there a specific reason it's alised to A?

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,363 @@
...
+import time
+from logging import getLogger
+
+from google.cloud import dns
+
+from .base import BaseProvider
+from ..record import Record
+
+
+class _GoogleCloudRecordSetMaker(object):
```
ross: Fwiw, if this is modeled after the Route53Provider it's a little "extra" complicated b/c of the need to deal with geo records that combine/split into multiple rrsets. Looking at the code here I don't think that seems to be happening here so this may benefit from following the more straightforward providers, e.g. PowerDNS where there's a `_data_from_X` and `_params_for_X`. If so I think this case would just turn into something like `_rrset_for_A` methods on the provider itself. Feel free to lmk if I'm misreading something.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,363 @@
...
+
+class GoogleCloudProvider(BaseProvider):
+    """
+    Google Cloud DNS provider
+
+    google_cloud:
+        class: octodns.provider.googlecloud.GoogleCloudProvider
+        # Credentials file for a service_account or other account can be
+        # specified with the GOOGLE_APPLICATION_CREDENTIALS environment
+        # variable. (https://console.cloud.google.com/apis/credentials)
```
ross: If possible octoDNS prefers to provide credentials explicitly with config vars and params to clients rather than using them from the ENV. That generally allows for the management of records in multiple accounts in the same config/setup. E.g. if you worked at a company with multiple Google Cloud accounts you could configure something like:

```yaml
providers:
  google_cloud_us:
    ...
  google_cloud_eu:
    ...
```

.gitignore
doddo: Thanks, I did not know that. Reverted.

octodns/provider/googlecloud.py
doddo: I did use most of the others as role models for this one, mostly it was the Azure one. It's no particular reason, I guess it makes sense to move them all into the same class.

octodns/provider/googlecloud.py
doddo: yes. Updated.

octodns/provider/googlecloud.py
doddo: Yes. Updated this too.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+                gcloud_changes.add_record_set(
+                    _rrset_func(gcloud_zone, change.new))
+            else:
+                raise RuntimeError('Change type "{}" for change "{!s}" '
+                                   'is none of "Create", "Delete" or "Update'
+                                   .format(class_name, change))
+
+        gcloud_changes.create()
+        i = 1
+        while gcloud_changes.status != 'done':
```
ross: Cool that gcloud has a transactional API. Nothing to do with the provider, but do you know if it's implemented in an atomic/all or nothing matter?

This `while` loop could result in an infinite loop if there's any sort of error state since it would never hit the done state. It'd generally be safer to flip things around and do a `while gcloud_changes.status == 'whatever-the-in-progress-state-is':`

That said we'll want to eventually give up if we don't change state (something really went wrong) so I'd probably suggest reworking this to look something more like:

```python
for in in range(60):
    gcloud_changes.reload()
    if gcloud_changes.status != 'whatever-the-in-progress-state-is':
        # we're no longer in progress so break the loop
        break
    # better name, but nice to make this a constant so you can override it when 
    # testing this case to avoid delaying in a unit test
    sleep(self.CHANGE_LOOP_WAIT)
if gcloud_changes.status != 'done':
    # this is an error we should raise
# if we're here we're done
````

File: octodns/provider/googlecloud.py [hunk 6]
```diff
@@ -0,0 +1,343 @@
...
+        # and only contain lowercase letters, digits or dashes
+        zone_name = re.sub("[^a-z0-9-]", "",
+                           dns_name[:-1].replace('.', "-"))
+        # make sure that the end result did not end up wo leading letter
+        if re.match('[^a-z]', zone_name[0]):
+            # I cannot think of a situation where a zone name derived from
+            # a domain name would'nt start with leading letter and thereby
+            # violate the constraint, however if such a situation is
+            # encountered, add a leading "a" here.
+            zone_name = "a%s" % zone_name
```
ross: Hrm. Not 100% sure either as it's isn't something I've dug into. A quick search seems turned up https://stackoverflow.com/questions/7111881/what-are-the-allowed-characters-in-a-subdomain and https://en.wikipedia.org/wiki/Domain_name.

> Domain names may be formed from the set of alphanumeric ASCII characters (a-z, A-Z, 0-9), but characters are case-insensitive. In addition the hyphen is permitted if it is surrounded by characters, digits or hyphens, although it is not to start or end a label. Labels are always separated by the full stop (period) character in the textual name representation.

So I think all of this can be omitted/ignored and assumed to be well behaved. If we do want to enforce this it'd probably make sense to add as rules to the Zone class (in a separate future PR)

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+            :return: a resource record set
+            :type return: google.cloud.dns.ResourceRecordSet
+        """
+        gcloud_iterator = gcloud_zone.list_resource_record_sets(
+            page_token=page_token)
+        for gcloud_record in gcloud_iterator:
+            yield gcloud_record
+        # This is to get results which may be on a "paged" page.
+        # (if more than max_results) entries.
+        if gcloud_iterator.next_page_token:
```
ross: If I'm reading this correctly I believe this may only handle the case when there's 2 pages of results. Rough stab at what may help here if that's correct:

```python
while True:
    gcloud_iterator = gcloud_zone.list_resource_record_sets(page_token=page_token)
    for gcloud_record in gcloud_iterator:
        yield gcloud_record
    page_token = gcloud_iterator.next_page_token
    # This assumes that it's None when there's no more pages
    if page_token is None: 
        break
```

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+            :param dns_name: fqdn of dns name for zone to get.
+            :type dns_name: str
+            :param page_token: page token for the page to get
+            :type page_token: str
+            :param create: if true, create ManagedZone if it does not exist
+                           already
+
+            :type return: new google.cloud.dns.ManagedZone
+        """
+        # Find the google name for the incoming zone
```
ross: I think this may have the same 2-page problem as records. It'd also be O(n^2) since it'd walk all zones for each managed zone to find the one we're after. It'd probably be best to break up the process of getting the full list of zones from finding the one we're after. 

The method that gets all the zones could just store them in a _cached hash of zone -> object. There's several providers that do that already. Otherwise the code would look pretty much the same as the `while True` loop for records. 

`_get_gcloud_zone` would then just be able to look up the zone in the cached hash.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+
+        self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
+                       target, lenient)
+        before = len(zone.records)
+
+        gcloud_zone = self._get_gcloud_zone(zone.name)
+
+        _records = set()
+        if gcloud_zone:
+            for gcloud_record in self._get_gcloud_records(gcloud_zone):
```
ross: Nice use of generators.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+            :type return: void
+        """
+
+        self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
+                       target, lenient)
+        before = len(zone.records)
+
+        gcloud_zone = self._get_gcloud_zone(zone.name)
+
+        _records = set()
```
ross: Is there a specific need for the `set`? Is it filtering duplicates or something? If so it seems worth a comment describing the what's up.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+        if gcloud_zone:
+            for gcloud_record in self._get_gcloud_records(gcloud_zone):
+                if gcloud_record.record_type.upper() in self.SUPPORTS:
+                    _records.add(gcloud_record)
```
ross: If the set isn't being used to de-dup or something you can probably just change this to `not in` and `continue` to skip the record and otherwise get rid of `_records` and the second for each on the line below.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,343 @@
...
+                '{order} {preference} "{flags}" "{service}" '
+                '"{regexp}" {replacement}'
+                .format(**v.data) for v in record.values])
+
+    _rrset_for_NS = _rrset_for_A
+
+    _rrset_for_PTR = _rrset_for_CNAME
+
+    def _rrset_for_SPF(self, gcloud_zone, record):
+        if 'values' in record.data:
```
ross: `data` is essentially a `to_json` type method so it probably shouldn't be used in these. They should instead get at the values property `record.values`.

Specifically for TXT you'll probably want to check out https://github.com/github/octodns/pull/115 and determine whether gcloud automatically handles TXT values > 255 chars or if the provider needs to do `chunked_values`.

requirements.txt
```diff
@@ -9,6 +9,7 @@ dnspython==1.15.0
 docutils==0.14
 dyn==1.8.0
 futures==3.1.1
+google-cloud==0.27.0
```
ross: Wow. This pulls in a lot of other dependencies:

> Successfully installed cachetools-2.0.1 certifi-2017.7.27.1 chardet-3.0.4 dill-0.2.7.1 enum34-1.1.6 future-0.16.0 futures-3.1.1 gapic-google-cloud-datastore-v1-0.15.3 gapic-google-cloud-error-reporting-v1beta1-0.15.3 gapic-google-cloud-logging-v2-0.91.3 gapic-google-cloud-pubsub-v1-0.15.4 gapic-google-cloud-spanner-admin-database-v1-0.15.3 gapic-google-cloud-spanner-admin-instance-v1-0.15.3 gapic-google-cloud-spanner-v1-0.15.3 google-auth-1.1.1 google-cloud-0.27.0 google-cloud-bigquery-0.26.0 google-cloud-bigtable-0.26.0 google-cloud-core-0.26.0 google-cloud-datastore-1.2.0 google-cloud-dns-0.26.0 google-cloud-error-reporting-0.26.0 google-cloud-language-0.27.0 google-cloud-logging-1.2.0 google-cloud-monitoring-0.26.0 google-cloud-pubsub-0.27.0 google-cloud-resource-manager-0.26.0 google-cloud-runtimeconfig-0.26.0 google-cloud-spanner-0.26.0 google-cloud-speech-0.28.0 google-cloud-storage-1.3.2 google-cloud-translate-1.1.0 google-cloud-videointelligence-0.25.0 google-cloud-vision-0.26.0 google-gax-0.15.15 google-resumable-media-0.2.3 googleapis-common-protos-1.5.3 grpc-google-iam-v1-0.11.4 grpcio-1.6.3 httplib2-0.10.3 idna-2.6 monotonic-1.3 oauth2client-3.0.0 ply-3.8 proto-google-cloud-datastore-v1-0.90.4 proto-google-cloud-error-reporting-v1beta1-0.15.3 proto-google-cloud-logging-v2-0.91.3 proto-google-cloud-pubsub-v1-0.15.4 proto-google-cloud-spanner-admin-database-v1-0.15.3 proto-google-cloud-spanner-admin-instance-v1-0.15.3 proto-google-cloud-spanner-v1-0.15.3 protobuf-3.4.0 pyasn1-0.3.7 pyasn1-modules-0.1.4 requests-2.18.4 rsa-3.4.2 six-1.11.0 tenacity-4.4.0 urllib3-1.22

script/test
```diff
@@ -24,5 +24,6 @@ export DNSIMPLE_TOKEN=
 export DYN_CUSTOMER=
 export DYN_PASSWORD=
 export DYN_USERNAME=
+export GOOGLE_APPLICATION_CREDENTIALS=
```
ross: You'll want to do the same in `script/cibuild`

octodns/provider/googlecloud.py
doddo: Yes these changes are Atomic, and if one fails it'll apply nothing. I was was thinking the very same thing that you'd probably want some sort of eventual timeout. I will add something similar to what you are suggesting.

octodns/provider/googlecloud.py
doddo: OK makes sense.

octodns/provider/googlecloud.py
doddo: Aactually this function will simply call itself recursively until the iterator returned does not have the "next_page_token". I did update the unit test to fetch a result which is on a simulated "page 3" and it does work.

octodns/provider/googlecloud.py
doddo: Yes I was thinking this very same thing. This would also ensure that when a new zone is created, it's possible to check that the generated zone name is not already in use by some other zone.

octodns/provider/googlecloud.py
doddo: yes you are right I will do that.

octodns/provider/googlecloud.py
doddo: understood.

doddo: Curious. No google cloud does not handle >255 chars for quoted text values, so it would have to use `chunked_values` too. I will have a look at this as soon as I have handled the other stuff. Should be just to replace `values` with `chunked_values` though.

octodns/provider/googlecloud.py
ross: Hrm. I think I was confused about the yields. I guess that would involve recursively yielding as well? That's not something I've run across before. I don't know enough about the implementation details of python's `yield` to know if that's a problem, but it seems like it'd at least be sub-optimal. 

I feel like the `while True` walk pages and yield the records is a bit simpler and easier to understand. Thoughts?

doddo: Yes. It will recursively yield until there are no more results. With python 3.3 comes the `yield from` expression "(allowing a generator to delegate part of its operations to another generator)"  which I think has been working great for making yields from recursive functions. this is python27 though and it will yield the results it got by recursively calling itself. Anyway, in this scenario I don't think it's going to have any measurable performance differences compared to a `while True` walk solution proposed by you, and I don't know which is easier to read, but that can be because I know how the function I wrote works and may be biased. Also I have no strong opinions in favour of either solution, only that the one I have written is written already, and thus requires no more work on my end for now.

script/test
doddo: OK added to coverage file cause that's the file which  I think you ment.

requirements.txt
doddo: yes option 2 is using the REST API of google cloud DNS, it's well-documented, and they have got good documentation about various record types including gotchas and the like (https://cloud.google.com/dns/records/json-record) but I'd rather save that for a future version of this provider.

octodns/provider/googlecloud.py
doddo: OK so how should I implement this one? I propose that if this change gets merged before #115 then I will update #115 with fix for googlecloud provider, and if the other way  around I rebase this PR with master, or if both gets merged I can open a new PR with the chunked_values. What do you suggest?

ross: I'm hoping to get https://github.com/github/octodns/pull/115 reviewed & shipped shortly. If it goes first you can pull the changes in. If not I'll make the change in that PR and ping you to :eye: and test there.

requirements.txt
ross: No worries, just surprised me when I installed it. Shouldn't hurt anything.

doddo:
Are you waiting for some further action from me at this point?

ross:
> Are you waiting for some further action from me at this point?

Nothing specific. I just need to get a bit of time to sit down and do a final pass over things before clicking the approve & merge button. Busy week for me this week b/c of the GitHub Universe conference. Hopefully get to it tomorrow or saturday.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,348 @@
...
+            else:
+                raise RuntimeError('Change type "{}" for change "{!s}" '
+                                   'is none of "Create", "Delete" or "Update'
+                                   .format(class_name, change))
+
+        gcloud_changes.create()
+
+        for i in range(120):
+            gcloud_changes.reload()
+            self.log.debug("Waiting for changes to complete")
```
ross: This message will be printed twice per loop. This first one should be ✂️

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,348 @@
...
+            # https://cloud.google.com/dns/api/v1/changes#resource
+            # status can be one of either "pending" or "done"
+            if gcloud_changes.status != 'pending':
+                break
+            self.log.debug("Waiting for changes to complete")
+            time.sleep(self.CHANGE_LOOP_WAIT)
+
+        if gcloud_changes.status != 'done':
+            raise RuntimeError("Timeout reached after {} seconds".format(
+                i * self.CHANGE_LOOP_WAIT))
```
ross: Will this always be a timeout? Is it possible that `status == 'error'`?

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,348 @@
...
+        while zone_name in [z.name for z in self.gcloud_zones.values()]:
+            # If there is a zone in google cloud alredy, then try suffixing the
+            # name with a -i where i is a number which keeps increasing until
+            # a free name has been reached.
+            m = re.match("^(.+)-([0-9]+$)", zone_name)
+            if m:
+                i = int(m.group(2)) + 1
+                zone_name = "{}-{!s}".format(m.group(1), i)
+            else:
+                zone_name += "-2"
```
ross: What case is this having to handle with the appends? Wouldn't it be sufficient to make the name be dns_name.replace('.', '-')? It'd seem really confusing to have multiple things with the same `zone_name`, but different `dns_names`. Not sure how things would get into that situation.

octodns/provider/googlecloud.py
```diff
@@ -0,0 +1,348 @@
...
+            name=zone_name,
+            dns_name=dns_name
+        )
+        gcloud_zone.create(client=self.gcloud_client)
+
+        # add this new zone to the list of zones.
+        self._gcloud_zones[gcloud_zone.dns_name] = gcloud_zone
+
+        self.log.info("Created zone %s. Fqdn %s." %
+                      (zone_name, dns_name))
```
ross: This should use `.format`.

octodns/provider/googlecloud.py
doddo: status is one of either "done" or "pending". Errors raises exceptions.

octodns/provider/googlecloud.py
doddo: The rationale is that since `dns_names` may have both hyphens and dots in the name whereas `zone_name` can have only hyphens, there are situations in which dns_name.replace('.', '-') of two differet dns_names results the same zone_name (for example say forest-treck.example.com and forest.treck.example.com).

ross: OK. Weird that google cloud has that sort of display name and it doesn't just use the uniq dns_name. I assume leaking some sort of internal implementation detail or doing it for for consistency with something somewhere else. :slightly_smiling_face: 

Anyway, given the full context I think the best bet here is to just use a uuid on all names which it's safe to assume will be unique without having to repeatedly check, and it can just always do it. There's a similar bit in the DynProvider for [its `DSFRuleset.label`](https://github.com/github/octodns/blob/e0f357d9f9c041238b069b5fc4c1f4c663bf16a8/octodns/provider/dyn.py#L545). 

That'd get rid of all the looping and regexes and just look like:

```python
zone_name = '{}-{}'.format(dns_name[:-1].replace('.', '-'), uuid4().hex)
```

doddo: You may in google cloud create several dns zones with the same DNS name (but different zone names). They appear to put different NS records in the zone. I have yet to find out what is the benefit of this approach.

jacobleonw:
hey @ross, do have an ETA on a release with this in?

ross:
> hey @ross, do have an ETA on a release with this in?

No specific plans on timing. I was hoping to get a few things in last week before cutting a release, but didn't make it to them. I'll stick it on the todo list and cut one with things are they are when I get a chance.

ross:
Cut 0.8.8.

jacobleonw:
Thanks, @ross

PR #116, opened 2017-10-05
add selectel as octodns provider

nikita-b:
Selectel is one of the largest cloud providers in Russia.

API overview: https://blog.selectel.com/managing-domains-with-the-selectel-dns-api/

### DIFF
--- a/octodns/provider/selectel.py
+++ b/octodns/provider/selectel.py
@@ -1,0 +1,278 @@
+#
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from collections import defaultdict
+
+from logging import getLogger
+
+from requests import Session
+
+from ..record import Record
+from .base import BaseProvider
+
+
+class SelectelAuthenticationRequired(Exception):
+    def __init__(self, msg):
+        Exception.__init__(self,
+                           'Authorization failed. Invalid or empty token.')
+
+
+class SelectelProvider(BaseProvider):
+    SUPPORTS_GEO = False
+
+    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'MX', 'NS', 'TXT', 'SRV'))
+
+    MIN_TTL = 60
+    MAX_TTL = 604800
+
+    API_URL = 'https://api.selectel.ru/domains/v1'
+
+    def __init__(self, id, token, *args, **kwargs):
+        self.log = getLogger('SelectelProvider[{}]'.format(id))
+        self.log.debug('__init__: id=%s', id)
+        super(SelectelProvider, self).__init__(id, *args, **kwargs)
+
+        self._sess = Session()
+        self._sess.headers.update({
+            'X-Token': token,
+            'Content-Type': 'application/json',
+        })
+        self._zone_records = {}
+        self._domain_list = self.domain_list()
+
+    def _request(self, method, path, params=None, data=None):
+        self.log.debug('_request: method=%s, path=%s', method, path)
+
+        url = '{}{}'.format(self.API_URL, path)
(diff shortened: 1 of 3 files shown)


octodns/provider/selectel.py
```diff
@@ -0,0 +1,278 @@
...
+
+from requests import Session
+
+from ..record import Record
+from .base import BaseProvider
+
+
+class SelectelAuthenticationRequired(Exception):
+    def __init__(self, msg):
+        Exception.__init__(self,
```
ross: The use of `super` is preferred here e.g.

```python
super(SelectelAuthenticationRequired, self).__init__('Authorization failed. Invalid or empty token.')
```

octodns/provider/selectel.py
```diff
@@ -0,0 +1,278 @@
...
+    _params_for_A = _params_for_multiple
+    _params_for_AAAA = _params_for_multiple
+    _params_for_NS = _params_for_multiple
+    _params_for_TXT = _params_for_multiple
+
+    _params_for_CNAME = _params_for_single
+
+    def _data_for_A(self, _type, records):
+        return {
+            'ttl': records[0]['ttl'],
```
ross: Based on the MIN/MAX elsewhere I believe you'll run into the same stuff I hit with the Cloudflare provider when there's a TTL configured on a record below the min (or above the max in this case.)

There's an `_include_change` method that allows the provider to look things over at the correct point and ignore things that only mis-match b/c of an unsupported ttl. You can see how that works for Cloudflare [here](https://github.com/github/octodns/blob/c45b7a5fb0c4e75987a866667de242ad78858fe1/octodns/provider/cloudflare.py#L193-L200). You'll probably need something very similar for this provider.

octodns/provider/selectel.py
```diff
@@ -0,0 +1,278 @@
...
+                    record = Record.new(zone, name, data, source=self,
+                                        lenient=lenient)
+                    zone.add_record(record)
+        self.log.info('populate:   found %s records',
+                      len(zone.records) - before)
+
+    def domain_list(self):
+        path = '/'
+
+        resp = self._request('GET', path)
```
ross: There any sort of pagination support needed here?

octodns/provider/selectel.py
```diff
@@ -0,0 +1,278 @@
...
+        domain_dict = {}
+
+        for domain in resp:
+            domain_dict[domain['name']] = domain
+
+        return domain_dict
+
+    def zone_records(self, zone):
+        path = '/{}/records/'.format(zone.name[:-1])
+        resp = self._request('GET', path)
```
ross: Same pagination question.

octodns/provider/selectel.py
```diff
@@ -0,0 +1,278 @@
...
+        path = '/{}/records/'.format(domain_id)
+        return self._request('POST', path, data=data)
+
+    def delete_record(self, domain, _type, zone):
+        self.log.debug('Delete record. Domain: %s, Type: %s', domain, _type)
+
+        domain_id = self._domain_list[domain]['id']
+        path = '/{}/records/'.format(domain_id)
+
+        records = self._request('GET', path)
```
ross: This would be O(n^2) for deleting records which would likely run into trouble when deleting many records from a large zone. It may make sense to look at the caching mechanism that some of the other providers incorporate to avoid this problem.

yzguy: Something like this bit here: https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L71-L87

`self.zones.get(domain, False)` is called instead of doing an API call directly.

ross:
Completed in https://github.com/github/octodns/pull/383

PR #117, opened 2017-10-09
dyn provider: add ability to cleanup sessions

markrechler:
In long running processes using Dyn as a sync source it's desireable
to close sessions that will take a while to timeout and flush caches
so that changes are picked up on subsequent runs.

### DIFF
--- a/octodns/provider/dyn.py
+++ b/octodns/provider/dyn.py
@@ -96,6 +96,9 @@ class DynProvider(BaseProvider):
         # Whether or not to support TrafficDirectors and enable GeoDNS
         # (optional, default is false)
         traffic_directors_enabled: true
+        # Whether or not to close the Dyn session and flush cache after a run
+        # (optional, default is false)
+        cleanup_session: true
 
     Note: due to the way dyn.tm.session.DynectSession is managing things we can
     only really have a single DynProvider configured. When you create a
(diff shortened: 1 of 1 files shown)


octodns/provider/dyn.py
```diff
@@ -383,6 +396,10 @@ def populate(self, zone, target=False, lenient=False):
         self.log.info('populate:   found %s records',
                       len(zone.records) - before)
 
+        if self.cleanup_session:
+            _CachingDynZone.flush_zone(zone.name[:-1])
```
ross: There's 2 more "cache" vars on the provider object around traffic directors. They'd likely need to be cleared too if we go this route.

https://github.com/markrechler/octodns/blob/c45b7a5fb0c4e75987a866667de242ad78858fe1/octodns/provider/dyn.py#L154-L156

Those should probably go into a `_reset_caches` function or something like which is called from the bottom of `__init__` as well as any other point where things need to be reset.

octodns/provider/dyn.py
```diff
@@ -709,3 +726,6 @@ def _apply(self, plan):
         self._apply_regular(desired, changes, dyn_zone)
 
         dyn_zone.publish()
+
+        if self.cleanup_session:
+            self._close_dyn_sess()
```
ross: Wouldn't the caches need to be invalidate here too?

markrechler:
The use case I ran into where this was an issue was calling `manager.sync` repeatedly from a webapp. The workflow in the application is updating Dyn and then using `manager.sync` to push to other providers. After the first sync call the cache would stick around essentially never tracking further updates.

octodns/provider/dyn.py
markrechler: Was relying on `publish()` to flush the cache in this case:
https://github.com/markrechler/octodns/blob/dyn_sessions/octodns/provider/dyn.py#L81

ross:
Cool. What you're seeing makes more sense now. In general the providers are intended to be one-shot. E.g. run through one cycle and then go away. There's nothing that inherently prevents them from being re-used, as you've seen, but they haven't been tested/verified to work when long running situations, as you've also seen.

To take the safest/known to work route I'd suggest creating the Manager object each time you go to call sync. In general the process of spinning it up should be fairly lightweight and should for sure solve the issue you're seeing.

I'll have to think a bit about what could be done to ensure that the providers are stable when used as long-lived objects. To be honest I'm not sure where I'd land on that. It there's a clear way to ensure that I'll take that route, but absent that I may look at explicitly making them one-shot so that it's obvious things are in uncharted territory. 

One unrelated thing you'll probably want to think through is making sure that you either don't run multiple syncs at once or that you're specific uses won't run into problems. The former being the safest route to take, but the latter would be possible with great care to make sure that the way you're adding/removing records won't run into trouble if adds/removes happen in parallel.

markrechler:
@ross sounds good. Appreciate the input. Will close this out for now.

PR #120, opened 2017-10-10
add ability to configure update/delete thresholds

joewilliams:
Fixes https://github.com/github/octodns/issues/61

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -21,10 +21,14 @@ class Plan(object):
     MAX_SAFE_DELETE_PCENT = .3
     MIN_EXISTING_RECORDS = 10
 
-    def __init__(self, existing, desired, changes):
+    def __init__(self, existing, desired, changes,
+                 update_pcent_threshold=MAX_SAFE_UPDATE_PCENT,
+                 delete_pcent_threshold=MAX_SAFE_DELETE_PCENT):
         self.existing = existing
         self.desired = desired
         self.changes = changes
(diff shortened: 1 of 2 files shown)


joewilliams:
This PR should allow you to pass an update and/or delete threshold argument to the provider to override the default threshold of 30%. 

That would look something like:

```
  dyn:
    class: octodns.provider.dyn.DynProvider
    update_pcent_threshold: .4
    delete_pcent_threshold: .4
``

File: octodns/provider/base.py [hunk 1]
```diff
@@ -55,14 +59,19 @@ def raise_if_unsafe(self):
             update_pcent = self.change_counts['Update'] / existing_record_count
             delete_pcent = self.change_counts['Delete'] / existing_record_count
 
-            if update_pcent > self.MAX_SAFE_UPDATE_PCENT:
+            self.log.debug('raise_if_unsafe: update_pcent_threshold=%d, '
+                           'delete_pcent_threshold=%d',
+                           self.update_pcent_threshold,
+                           self.delete_pcent_threshold)
```
ross: I think it'd be preferable to log these when the provider is created, __init__ https://github.com/github/octodns/pull/120/files#diff-f1e7f60f6ff68ac72db8613d4f410dbbR95, since that's the level at which things will be configured. It'd be hard to trace back this log line to which provider emitted it at the plan point of things since the plan doesn't know it's source provider and there many plans, one per zone.

PR #124, opened 2017-10-16
use . instead of source [SC2039]

nikaro:
Just a little fix to make it more portable across distributions.
`source` doesn't work everywhere under `/bin/sh`

cf. https://github.com/koalaman/shellcheck/wiki/SC2039


ross:
Thanks @nikaro, looks like there are a bunch of places in `script/*` that do `.` as well. If you don't mind making a pass through those to `s/./source/` as well I'll get this 👍  and merged.

nikaro:
Hello @ross, actually doing `.` is good because it is more widely supported. `source` was not working for me. I think you read my message too quickly :-)

Issue #126, opened 2017-10-16
DYN: Traffic director services not found.


timhughes:
It appears that other people have a ':' in the label that is returned from DYN. I dont think necessarily always true. Acording to the docs it is a "string" "Name of the Traffic Director service." https://help.dyn.com/update-dsf-api/ 

This is the example response i think the tests are using.  
https://github.com/github/octodns/blob/7f2c292f5ac5f99808c843f68dcb47c7ed4dccf7/tests/fixtures/dyn-traffic-director-get.json#L4179

If i modify the code at: 
https://github.com/github/octodns/blob/7f2c292f5ac5f99808c843f68dcb47c7ed4dccf7/octodns/provider/dyn.py#L292

to be like  
```
            for td in get_all_dsf_services():
                print(td)
                try:
                    print(td.label)
                    fqdn, _type = td.label.split(':', 1)
                except ValueError as e:
                    print("Ahhhhhh: %s" % e)
                    continue
```

Then the output i get when trying to octodns-dump the domain I get

<TrafficDirector>: website-production, ID: XXXX-XXXXXXXXXXXXXXXXX
website-production
Ahhhhhh: need more than 1 value to unpack
<TrafficDirector>: asia-trader, ID: YYYY-YYYYYYYYYYYYYYYY
asia-trader
Ahhhhhh: need more than 1 value to unpack

ross:
> "label": "unit.tests.:A"

This is the format that octoDNS uses for TD records that it creates. If there are existing TD/geo records that don't match this format (I wouldn't expect anyone's records to match) octoDNS won't be able to import them cleanly. 

There's not really an existing fully supported process for importing a setup with advanced records. I mulled over trying to do something like that early on, but decided that there was so much possible variation in the way that Dyn allows things to be set up that the chance of it fulling working were slim. Especially the case since I don't have examples to work from.

If the dump process is throwing errors then it's definitely something I'd like to avoid. The end result would probably be to print or error out with a clear warning about what's going on. 

I don't have a complete/good story on what importing such zones would look like. It'd probably involve getting a dump of all of the standard records and then manually recreating the advanced stuff. That would result in most things being left untouched, but would still recreate the TD records using octoDNS's setup. I know it's not a great story, but I don't have any leads on a better one at the moment.

timhughes:
When I build a minimal script that replicates the call I get this.

```python
import os
from pprint import pprint
from dyn.tm.session import DynectSession
from dyn.tm.zones import Zone
[...]

PR #128, opened 2017-10-17
adds warning to dyn provider when it cannot load a trafficdirector

timhughes:
If a traffic director is not named correctly the Dyn provider will not load it. This patch adds feedback to the user to make them aware of this issue rather than just continuing with no notification.  Fixes some of #126

### DIFF
--- a/octodns/provider/dyn.py
+++ b/octodns/provider/dyn.py
@@ -291,7 +291,8 @@ class DynProvider(BaseProvider):
                 try:
                     fqdn, _type = td.label.split(':', 1)
                 except ValueError as e:
-                    self.log.warn("Failed to load TraficDirector '{}': {}".format(td.label, e))
+                    self.log.warn("Failed to load TrafficDirector %s: %s",
+                            td.label, e.message)
                     continue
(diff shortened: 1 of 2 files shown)


octodns/provider/dyn.py
```diff
@@ -290,7 +290,8 @@ def traffic_directors(self):
             for td in get_all_dsf_services():
                 try:
                     fqdn, _type = td.label.split(':', 1)
-                except ValueError:
+                except ValueError as e:
+                    self.log.warn("Failed to load TraficDirector '{}': {}".format(td.label, e))
```
ross: The logging functions can actually do their own formatting, so it's slightly preferable to do 

```
self.log.warn("Failed to load TraficDirector '%s': %s", td.label, e)
```

I realize there's a bit of inconsistency in octoDNS with this currently, but it's something I plan to address so would like to make sure that new additions go with the "preferred" route.

ross:
This was manually merged with the requested update and some minor lint fixes.

timhughes:
tks

PR #132, opened 2017-10-23
Improve Validations for None/Empty values, Ignore None/Empty values during dump

yzguy:
**Fixes**: https://github.com/github/octodns/issues/131

When dumping a zone, one tested provider (NS1) allows a user to create a record with no answers. A record with no answers is not valid.

# Dump

1. When a user dumps a zone, it will fail with a ValidationError, for a `missing value(s)` reason
2. When the user specifies `--lenient` during the dump, it will give them a warning, write the record to the file, and a `value` or `values` key will not be present

If a user attempts to apply changes from that file, it will fail normally if `value` or `values` is not specified.

* Previous Behavior:

	**Command**:
	
	```
	octodns-dump --config-file=config/production.yaml --output-dir=tmp/ yzguy.io. nsone
	```
	
	**Output**:
	
	```
	IndexError: list index out of range
	```

* New Behavior

	User does not specify `--lenient`

	**Command**:
	
	```
	octodns-dump --config-file=config/production.yaml --output-dir=tmp/ yzguy.io. nsone
	```
	
	**Output**:
	
	```
	octodns.record.ValidationError: Invalid record test.yzguy.io.
	  - missing value(s)
	```
	
	User does specify `--lenient`
	
	**Command**:

	```
	octodns-dump --config-file=config/production.yaml --output-dir=tmp/ yzguy.io. nsone --lenient
	```
	
	**Output**:
	

	```
	2017-10-22T22:54:58  [140737049330624] WARNING Record Invalid record test.yzguy.io.
	  - missing value(s)
	```
	
	**File**:
	
	```
	test:
	  type: A
	```

# Noop and Doit

 1. If a user applies a file where the `values` key is an empty list or nothing, it will show a ValidationError with `missing value(s)` reason.
 
To fix this the user should apply a value, values, or delete the record as it is not valid.

* Previous Behavior:

	**Command**:

	```
	octodns-sync --config-file=./config/production.yaml
	```
	
	**File**:
	
	```
	test:
	  type: A
	  values:
   ```
   
   **Output**:
   
   ```
   TypeError: 'NoneType' object is not iterable
   ```

* New Behavior

	**Output**:
		
	```
	octodns.record.ValidationError: Invalid record test.yzguy.io.
	- missing value(s)
	```

* Add tests for ValuesMixin (overs records that allow multiple values), ValueMixin (cover records that allow single value)
* Add tests for data method for ValueMixin and ValuesMixin

### DIFF
--- a/octodns/record.py
+++ b/octodns/record.py
@@ -211,9 +211,30 @@ class _ValuesMixin(object):
         values = []
         try:
             values = data['values']
+            if not values:
+                values = []
+                reasons.append('missing value(s)')
+            else:
+                # loop through copy of values
+                # remove invalid value from values
+                for value in list(values):
+                    if value is None:
+                        reasons.append('missing value(s)')
+                        values.remove(value)
+                    elif len(value) == 0:
+                        reasons.append('empty value')
+                        values.remove(value)
         except KeyError:
             try:
-                values = [data['value']]
+                value = data['value']
+                if value is None:
+                    reasons.append('missing value(s)')
+                    values = []
+                elif len(value) == 0:
+                    reasons.append('empty value')
+                    values = []
+                else:
+                    values = [value]
             except KeyError:
                 reasons.append('missing value(s)')
 
@@ -238,10 +259,16 @@ class _ValuesMixin(object):
     def _data(self):
         ret = super(_ValuesMixin, self)._data()
         if len(self.values) > 1:
-            ret['values'] = [getattr(v, 'data', v) for v in self.values]
-        else:
+            values = [getattr(v, 'data', v) for v in self.values if v]
+            if len(values) > 1:
+                ret['values'] = values
+            elif len(values) == 1:
+                ret['value'] = values[0]
+        elif len(self.values) == 1:
             v = self.values[0]
-            ret['value'] = getattr(v, 'data', v)
+            if v:
+                ret['value'] = getattr(v, 'data', v)
+
         return ret
 
     def __repr__(self):
@@ -349,6 +376,10 @@ class _ValueMixin(object):
         value = None
         try:
             value = data['value']
+            if value is None:
+                reasons.append('missing value')
+            elif value == '':
+                reasons.append('empty value')
         except KeyError:
             reasons.append('missing value')
         if value:
@@ -366,7 +397,8 @@ class _ValueMixin(object):
 
     def _data(self):
         ret = super(_ValueMixin, self)._data()
-        ret['value'] = getattr(self.value, 'data', self.value)
+        if self.value:
+            ret['value'] = getattr(self.value, 'data', self.value)
         return ret
 
     def __repr__(self):
--- a/tests/test_octodns_record.py
+++ b/tests/test_octodns_record.py
@@ -96,6 +96,57 @@ class TestRecord(TestCase):
 
             DummyRecord().__repr__()
 
+    def test_values_mixin_data(self):
+        # no values, no value or values in data
+        a = ARecord(self.zone, '', {
+            'type': 'A',
+            'ttl': 600,
+            'values': []
+        })
+        self.assertNotIn('values', a.data)
+
+        # empty value, no value or values in data
+        b = ARecord(self.zone, '', {
+            'type': 'A',
+            'ttl': 600,
+            'values': ['']
+        })
+        self.assertNotIn('value', b.data)
+
+        # empty/None values, no value or values in data
+        c = ARecord(self.zone, '', {
+            'type': 'A',
+            'ttl': 600,
+            'values': ['', None]
+        })
+        self.assertNotIn('values', c.data)
+
+        # empty/None values and valid, value in data
+        c = ARecord(self.zone, '', {
+            'type': 'A',
+            'ttl': 600,
+            'values': ['', None, '10.10.10.10']
+        })
+        self.assertNotIn('values', c.data)
+        self.assertEqual('10.10.10.10', c.data['value'])
+
+    def test_value_mixin_data(self):
+        # unspecified value, no value in data
+        a = AliasRecord(self.zone, '', {
+            'type': 'ALIAS',
+            'ttl': 600,
+            'value': None
+        })
+        self.assertNotIn('value', a.data)
+
+        # unspecified value, no value in data
+        a = AliasRecord(self.zone, '', {
+            'type': 'ALIAS',
+            'ttl': 600,
+            'value': ''
+        })
+        self.assertNotIn('value', a.data)
+
     def test_geo(self):
         geo_data = {'ttl': 42, 'values': ['5.2.3.4', '6.2.3.4'],
                     'geo': {'AF': ['1.1.1.1'],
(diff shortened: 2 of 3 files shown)


tests/test_octodns_record.py
```diff
@@ -718,6 +718,24 @@ def test_base(self):
...
         self.assertEquals(['invalid ttl'], ctx.exception.reasons)
 
+        # value, empty
+        with self.assertRaises(ValidationError) as ctx:
+            Record.new(self.zone, 'www', {
+                'type': 'CNAME',
+                'ttl': 600,
+                'value': ''
+            })
+        self.assertEquals(['missing value'], ctx.exception.reasons)
```
ross: Probably want to go ahead and add a `value: None` test here too just to cover that case. It shouldn't happen with the built-in providers, but it'd be possible to get there with someone's internal/custom built provider.

tests/fixtures/dnsimple-invalid-content.json
```diff
@@ -1,106 +0,0 @@
-{
-  "data": [
-    {
-      "id": 11189898,
-      "zone_id": "unit.tests",
-      "parent_id": null,
-      "name": "naptr",
-      "content": "",
```
ross: Hrm. Wonder if i managed to get empty content into/out of their provider or if I was just aggressively testing things. Either way most of these seem like cases that are now handled globally. Only one I'm not sure about the naptr. Looks like that was converting to an empty value which was just weird in general. All of that predated the validation improvements so it's probably fine for them to go away so long as the code coverage is still complete.

yzguy:
Test added. ValidationError thrown with `missing value` reason as expected. I was wondering if that was a good test, so I'm glad you pointed it out as well.

ross:
Just checked this out locally and looked around at it. I think the changes are generally in order, but there\'s a handful of branches/exceptions that no longer have test coverage. You can find them with `./script/coverage` and then opening up `htmlcov/index.html`.  

There were several bits of error handling in `DnsSimpleProvider` that aren\'t hit after those special cases were removed from it\'s tests. They looked to be things that were skipping over invalid data so it may be a case where the code itself needs to be updated to match the new lenient behavior (of including the record without a value when allowed.) 

The other place with missing coverage is in `_ValuesMixin` where some of the cases of handling None, zero-length, or len == 1 aren\'t being exercised.

screen shot 2017-10-24 at 2 09 45 pm

yzguy:
Fixed the one spot that showed up in coverage
I moved around the tests to better spots that are for testing the ValuesMixin and ValueMixin

Trying to figure out the other part that showed up. I think I was only able to trigger it while dumping the zone. Still tracing through to figure out how it would be called otherwise

Thanks for your help! Will hopefully have it figured out by tomorrow

octodns/record.py
```diff
@@ -207,9 +207,15 @@ def validate(cls, name, data):
         values = []
         try:
             values = data['values']
+            if values is None or len(values) == 0:
+                values = []
+                reasons.append('missing value(s)')
```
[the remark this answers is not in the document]
yzguy: So the `values is None` is specifically for NSOne, as I believe that's the only case where a provider allows records with no answers. You end up with `data['values'] = None`

As the individual answers, I do think it's a good point to check them individually check to invalidate bad answers

If we do something like
```
test:
  type: A
  values:
    - ''
    -
```

We would get
```
octodns.record.ValidationError: Invalid record test.yzguy.io.
  - empty value
  - empty value
```

Which is good, because neither are good.

octodns/record.py
```diff
@@ -207,9 +207,15 @@ def validate(cls, name, data):
...
         except KeyError:
             try:
                 values = [data['value']]
+                if values[0] is None or len(values) == 0:
```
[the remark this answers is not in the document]
yzguy: This makes sense. Will be adding this.

One question: If we know the value is empty, eg. `''`, why let it through? In the case of an A record, it will just get a `  - invalid ip address ""` reason added. My first reaction is it shouldn't be let through as you've already marked it as an empty value.

Would love to hear another opinion though.

yzguy:
Squashed/rebased commits onto master.

Everything is looking good, just need to finish the tests for the _data methods for `_ValuesMixin` and `_ValueMixin`

FWIW, I'm loving this back and forth. I'm learning a lot about testing/coverage as well. I just think OctoDNS is neat tool. I have a lot of experience with DNS so this is a very approachable project to contribute to.

Thanks for your help/patience

octodns/record.py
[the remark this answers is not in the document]
ross: > One question: If we know the value is empty, eg. '', why let it through? In the case of an A record, it will just get a - invalid ip address "" reason added.

My thinking was that A/AAAA would be caught by the IPAddress check, but other things that use `_ValueMixin` or `_ValuesMixin`, e.g. `TXT`, wouldn't check for empty. Some of them would get caught by other checks like no trailing `.`. In the end it seemed to be the safest & most consistent to check for empty explicitly. 

> In _ValueMixin, by doing if value it's not letting a None or empty value through, whereas in _ValuesMixin it would be on each individual value

👍 , `if value`'s check for truthy would be fine. A similar check in a foreach on `_ValuesMixing` would work there.

ross:
> FWIW, I'm loving this back and forth. Feedback is great and I'm learning a lot about testing/coverage as well. I just think OctoDNS is neat tool. I have a lot of experience with DNS administration so this is a very approachable project to contribute to.

Glad to hear it on all counts. I've always found coverage and linting to be worth it for things that matter. Most of of the bits they find are inconsequential, but the payoff is that one out of ten of them is something real and beyond that they just make you think about things a bit more deeply. 

> Thanks for your help/patience

Thanks for the same.

tests/test_octodns_record.py
```diff
@@ -755,17 +755,104 @@ def test_A_and_values_mixin(self):
...
+            'ttl': 600,
+            'values': [
+                '1.2.3.4',
                 '1.2.3.5',
             ]
         })
 
-        # missing value(s)
+        # no values, no value or values in data
+        a = ARecord(self.zone, '', {
```
yzguy: Not sure if this is an appropriate place for these test.

It's testing the _data method of ValuesMixin, but not actually a Validation. This is under TestValidation class

yzguy:
Alrighty, cleaned up those spots that only had partial coverage.

I've also added some stuff in _data for ValueMixin and ValuesMixin so that `value` or `values` is not added to the return value, this makes it so that it will not be output to the YAML file during a dump.

In the very rare/probably never going to happen case of having something like `['', '10.10.10.10']`, it will output the IP Address, but not the empty value. Thought being that if it were somehow working, and the IP address was being used, it would get rid of the bad value, but not the good one, that could possibly be used. Removing it might cause an issue/outage

tests/test_octodns_record.py
ross: Don't have strong opinions on whether they should live here or in `TestRecord`. If you don't feel like they fit with validations it'd probably make sense to move them up there. Probably worth sticking them in a dedicated function, `def test_data` or similar.

tests/fixtures/dnsimple-invalid-content.json
ross: Looks like there's some missing coverage in `DnsimpleProvider` now for some special cases where they allow invalid values to be submitted to their API for the complex record types (NAPTR, SSHFP, SRV.) Those snippets are similar to the special case needed for NS1 in your other PR. 

Played around with it a bit and it looks like they're covered correctly if this fixture and the tests below are restored and a few `lenient`s are added to the `populate` and record creations in the test code below.

```python
+            provider.populate(zone, lenient=True)
+            self.assertEquals(set([
+                Record.new(zone, '', {
+                    'ttl': 3600,
+                    'type': 'SSHFP',
+                    'values': []
+                }, lenient=True),
+                Record.new(zone, '_srv._tcp', {
+                    'ttl': 600,
+                    'type': 'SRV',
+                    'values': []
+                }, lenient=True),
+                Record.new(zone, 'naptr', {
+                    'ttl': 600,
+                    'type': 'NAPTR',
+                    'values': []
+                }, lenient=True),
+            ]), zone.records)
```

yzguy:
......and Done :)

Squashed it down to 1 commit, no sense in cluttering up the commit history.

tests/test_octodns_record.py
```diff
@@ -96,6 +96,57 @@ def __init__(self):
 
             DummyRecord().__repr__()
 
+    def test_values_mixin_data(self):
```
ross: 👍

ross:
Thanks!

Issue #114, opened 2017-10-02
OvhProvider: Support for DKIM record


webs86:
Ovh translate in his system TXT record with DKIM key with a dedicated record named DKIM and not TXT (if I check dns record via dig I view correct TXT record, but via API it is an DKIM record).

When I use `octodns-dump` to create YAML config file it fail because it does not find `_data_for_DKIM`

This is traceback:
~~~
Traceback (most recent call last):
  File "/home/webs86/Documenti/gitlab/samuelez/dns/env/bin/octodns-dump", line 9, in <module>
    load_entry_point('octodns==0.8.6', 'console_scripts', 'octodns-dump')()
  File "/home/webs86/Documenti/gitlab/samuelez/dns/env/local/lib/python2.7/site-packages/octodns/cmds/dump.py", line 31, in main
    manager.dump(args.zone, args.output_dir, args.lenient, *args.source)
  File "/home/webs86/Documenti/gitlab/samuelez/dns/env/local/lib/python2.7/site-packages/octodns/manager.py", line 362, in dump
    source.populate(zone, lenient=lenient)
  File "/home/webs86/Documenti/gitlab/samuelez/dns/env/local/lib/python2.7/site-packages/octodns/provider/ovh.py", line 65, in populate
    data_for = getattr(self, '_data_for_{}'.format(_type))
AttributeError: 'OvhProvider' object has no attribute '_data_for_DKIM'
~~~

ross:
Interesting. If you set up a TXT record with the DKIM info it in with OVH does everything work correctly? If so the cleanest path may be to add a `_data_for_DKIM` and have that return an appropriate TXT record. That could run in to interesting problems during the sync/apply so would need some checking and testing.

/cc @trnsnt as the author of the OVH provider

trnsnt:
Hello,
Sorry for the delay, I will look at it this week. It seems that adding _data_for_DKIM is a good solution.

webs86:
Hi @trnsnt ,
I have tried to implement it, but I have some problems with tests. You can view on my fork on ovh-dkim-record.

Thank you

ross:
> I have tried to implement it, but I have some problems with tests. You can view on my fork on ovh-dkim-record.

@webs86 took a quick look at your branch. Since `DKIM` is a specific use of TXT and not really a distinct record type (outside of OVH I guess) It's probably best to hide the implementation detail from the user. I can think of two ways to go about that:

[...]

PR #135, opened 2017-10-26
OVH: Add support of DKIM records

trnsnt:
Add support to DKIM for OVH #114 :

- If OVH record is DKIM, change his type to TXT
- For TXT records in yaml file, we do a check to verify if record is DKIM or TXT. This has to be done because TXT records are limited (and truncated) to 255 chars.

### DIFF
--- a/octodns/provider/ovh.py
+++ b/octodns/provider/ovh.py
@@ -5,6 +5,8 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
+import base64
+import binascii
 import logging
 from collections import defaultdict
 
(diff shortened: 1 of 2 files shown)


octodns/provider/ovh.py
```diff
@@ -32,8 +34,8 @@ class OvhProvider(BaseProvider):
 
     SUPPORTS_GEO = False
 
-    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'MX', 'NAPTR', 'NS', 'PTR', 'SPF',
-                    'SRV', 'SSHFP', 'TXT'))
+    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'DKIM', 'MX', 'NAPTR', 'NS', 'PTR',
```
ross: I think the `DKIM` isn't needed here.

trnsnt: I have put DKIM here, because I reuse SUPPORTS in populate:

```
if _type not in self.SUPPORTS:
    ****
    continue
```

ross: Ah. Didn't catch that. The main use-case for `SUPPORTS` is [in the base source/provider](https://github.com/github/octodns/blob/1a0055eedb6acc4c8b2319efd7208544d47bdb11/octodns/source/base.py#L38-L39). The secondary/related use is for octoDNS users to see what record types the provider can work with. That kind of puts this in a grey area since `DKIM` wouldn't apply to the first, but would to the second. 

I guess it makes sense to leave it in, but if you don't mind please add a comment above the `SUPPORTS` variable saying why it's there b/c it's a type specific to OVH that the provider supports by converting to `TXT`.

trnsnt: I have added a comment explaining that SUPPORTS is used in populate method

Issue #131, opened 2017-10-20
NS1: Single Value Records with no answer raises IndexError


yzguy:
So just dumping our zone, got this error. We had a PTR record in our forward zone in NS1, no idea why, but it also had no answers.

Response from NS1
```
PTR, {u'domain': u'asmith.net', u'short_answers': [], u'link': None, u'ttl': 3600, u'tier': 1, u'type': u'PTR', u'id': u'1b05cd7943dde9d3'}
```

Error
```
Traceback (most recent call last):
  File "/Users/asmith/as-octodns/venv/bin/octodns-dump", line 11, in <module>
    sys.exit(main())
  File "/Users/asmith/as-octodns/venv/lib/python2.7/site-packages/octodns/cmds/dump.py", line 31, in main
    manager.dump(args.zone, args.output_dir, args.lenient, *args.source)
  File "/Users/asmith/as-octodns/venv/lib/python2.7/site-packages/octodns/manager.py", line 362, in dump
    source.populate(zone, lenient=lenient)
  File "/Users/asmith/as-octodns/venv/lib/python2.7/site-packages/octodns/provider/ns1.py", line 156, in populate
    record = Record.new(zone, name, data_for(_type, record),
  File "/Users/asmith/as-octodns/venv/lib/python2.7/site-packages/octodns/provider/ns1.py", line 76, in _data_for_CNAME
    'value': record['short_answers'][0],
IndexError: list index out of range
```

The record doesn't make sense, plus why have a record with no answers. I believe this should probably throw a ValidationError with a "No Records" reason.

Also at least for NSOne, it's another convenience, you don't have to put the `.` in, and it will still put it when you query for it.

yzguy:
I'd like to tackle this one, would like some feedback on if it should be a validation error or not, etc.

Thanks

ross:
> I'd like to tackle this one, would like some feedback on if it should be a validation error or not, etc.

That's an interesting one. I don't think it is/should be valid to have a record with no values. I just tried to find a way in a couple other providers to add a "node" without any values and wasn't able to do so.

I think the best thing to do here is probably to handle that case in the NS1Provider's [`_data_for_CNAME`](https://github.com/github/octodns/blob/f39e1d28c810138daa62a86cafdc39492bcad670/octodns/provider/ns1.py#L71-L76). Happy to take a look at that and answer any questions if you run into any bumps in the road. The unit testing for that case will be the more interesting bit. 

[...]

PR #136, opened 2017-10-27
set default value for nsone cname to None, use first value if non-zer…

yzguy:
NS1 will allow you to create a record with no answer. You cannot query for it however, but it will come back in their API/UI (According to their support, this is a feature/expected behavior, there is no RFC saying it is incorrect)

Other record types have a default value of `values = []`. CNAME, ALIAS, and PTR are the special cases in which only 1 answer is valid.

This adds a default value of None, otherwise only set the value if there is one

This addresses one part of: #131

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -69,10 +69,14 @@ class Ns1Provider(BaseProvider):
         }
 
     def _data_for_CNAME(self, _type, record):
+        try:
+            value = record['short_answers'][0]
+        except IndexError:
+            value = None
         return {
             'ttl': record['ttl'],
             'type': _type,
-            'value': record['short_answers'][0],
+            'value': value,
         }
 
     _data_for_ALIAS = _data_for_CNAME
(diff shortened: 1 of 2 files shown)


yzguy:
I'm not sure if `None` or `''` is a better initial value.

`None` will produce a `null` in the YAML file, eg. `value: null`
`''` will produce a `''` value in the YAML file, eg. `value: ''`

In validations for None, zero-length values, it will not even write the value key to the YAML file. I think `None` is a good value to check in that case, rather than an empty string.

octodns/provider/ns1.py
```diff
@@ -69,10 +69,13 @@ def _data_for_CAA(self, _type, record):
         }
 
     def _data_for_CNAME(self, _type, record):
+        value = None
+        if len(record['short_answers']) == 1:
+            value = record['short_answers'][0]
```
ross: Slightly more pythonic way of writing that would be using try/except, thought that's starting to get into preference territory. 

```python
try:
    value = record['short_answers'][0]
except IndexError:
    value = None
```

yzguy:
Updated. I haven't seen that before (granted I don't look at a lot of python code), however I think it's a lot more clear about what it's doing than checking length.

Thanks

PR #137, opened 2017-10-31
Add Geo record zone tests

vanbroup:
@ross, I started to add some test geo records and it looks like the `A`, `AAAA` and the new `CNAME` all fail to detect the zone changes?

I don't see any code related to geo records in the `changes`  function of `zone.py` (line 99), have you left that out on purpose?

Related to: https://github.com/github/octodns/issues/62

### DIFF
--- a/tests/test_octodns_zone.py
+++ b/tests/test_octodns_zone.py
@@ -236,3 +236,71 @@ class TestZone(TestCase):
         zone.add_record(cname)
         with self.assertRaises(InvalidNodeException):
             zone.add_record(a)
+
+    def test_geo_records(self):
+        zone_a = Zone('unit.tests.', [])
+        zone_aaaa = Zone('unit.tests.', [])
+        zone_cname = Zone('unit.tests.', [])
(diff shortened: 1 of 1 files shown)


ross:
> I don't see any code related to geo records in the changes function of zone.py (line 99), have you left that out on purpose?

It was definitely WIP and it's been long enough that I don't recall the exact status of the code. I think it was able to create them from scratch, but there may have been issues with the health checks. I don't think I made it to the point of testing changes. If I had to guess I'd say there's probably a few days to a week of work if I was able to sit down and work on it un-interrupted to get it shipped. Unfortunately not something I'll be able to do soon.

vanbroup:
My question is not really about `CNAME` or the branch but this looks to be failing for all GeoIP record types? Is that something you are aware of?

ross:
> My question is not really about CNAME or the branch but this looks to be failing for all GeoIP record types? Is that something you are aware of?

I'm sure I ran the tests so probably was aware of it at the time. 😕  At the very least I'm not surprised. It definitely wasn't a finished product.

PR #139, opened 2017-11-03
Improve setuptools capabilities

4383:
Hey!

### Overview
Using `setup.cfg` to configure project packaging. functionality introduced by [setuptools 30.3.0](http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files).

### Features
- centralize `requirements*` management
- remove packaging configuration from python code
- using conf file for handling project configuration
- adapted/updated bootstrap
- reduce the number of files to maintains
- remove duplicate requirements declarations
- adding project classifiers

Have fun!


octodns/__init__.py
```diff
@@ -1,6 +1,25 @@
...
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
+import pkg_resources
+from os import path
+from setuptools.config import read_configuration
+
+
+def _extract_version(package_name):
+    try:
+        return pkg_resources.get_distribution(package_name).version
```
ross: Does the setup.cfg file get installed alongside the module? I assume not and if that's the case I think it'd be preferable to flip the order of things here, e.g. look for setup.cfg first and then fall back to the pkg_resources based version. 

It'd be pretty edge-case-y, but it seems like it'd be possible to have a system or pip installed version that isn't the one loaded into the python interpreter in which case you'd have code that doesn't match the installed version.

requirements.txt
```diff
@@ -1,23 +0,0 @@
...
-ipaddress==1.0.18
-jmespath==0.9.3
-msrestazure==0.4.10
-natsort==5.0.3
-nsone==0.9.14
-ovh==0.4.7
-python-dateutil==2.6.1
-requests==2.13.0
-s3transfer==0.1.10
-six==1.10.0
```
ross: This is the full/complete list of requirements for octoDNS if you used 100% of it's providers, not necessarily the ones that someone would install when running it and only using a subset. There was a separate minimal list in setup.py for this reason and the expectation was that users would explicitly install the subset of these needed for their providers. If all of this moves to a single list in setup.cfg that'd force everyone to install everything all the time. I don't want to do that since some of the providers, Google Cloud specifically, install dozens of indirect requirements. 

Not sure, but maybe that means there needs to be 3 lists in setup.cfg or maybe just 2? One with the base set that's always installed as hard-requirements, and then the dev set would get the 100% coverage provider requirements in addition to the dev tools stuff.

octodns/__init__.py
```diff
@@ -1,6 +1,25 @@
...
+        _conf = read_configuration(
+            path.join(
+                path.dirname(path.dirname(__file__)),
+                'setup.cfg'
+            )
+        )
+        return _conf['metadata']['version']
+
+
+__version__ = _extract_version('octodns')
```
ross: I think it'd be nice to avoid loading the version since it now involves some "work" in cases where it's not going to be used. Something like the following would do that, but there may be slightly cleaner ways to pull it off.

```python
#!/usr/bin/env python

class _VersionLoader(object):
    version = None

    @property
    def __version__(self):
        if self.version == None:
            # do the work to load it
            self.version = 42
        return self.version


__VERSION__ = _VersionLoader().__version__

if __name__ == "__main__":
    print(__VERSION__)
```

```bash
otter:~ ross$ python /tmp/foo.py
42
```

octodns/__init__.py
```diff
@@ -1,6 +1,25 @@
-'OctoDNS: DNS as code - Tools for managing DNS across multiple providers'
-
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
+import pkg_resources
+from os import path
```
ross: It's personal preference, but I like to do `from os.path import dirname, join` so that the invocations later are much less verbose. Seems like doing things that way here would prevent a lot of the line wrapping that's necessary.

README.md
```diff
@@ -21,6 +21,7 @@ $ cd dns
 $ virtualenv env
 ...
 $ source env/bin/activate
+$ pip install -U setuptools
 $ pip install octodns
```
ross: What's the error message people will get if they try and install octoDNS without updating setuptools? I'd imagine people familiar with python are just gonna type `pip install octodns` and I like to avoid that being an immediate failure. I guess that's my biggest reservation with making this update/change in general. 

Beyond that I think i'd prefer to have something like `pip install setuptools>=<min-required-version>` here so that it doesn't imply that you always have to be running the absolute latest. That should allow people in situations who already install a specific version of setuptools to know whether or not they have something new enough.

requirements.txt
ross: Also makes me realized that the providers should probably either list out their requirements or check for them and complain if they're not installed. Longer term I may break up the more involved providers into their own modules, but don't have specific plans/timeline for that yet.

4383:
@ross thanks for your comments :) changes done!

### Commits
I don't squash my commits to facilitate your proofreading. I can squash commits after if you want.

### Changes
In this review I have introduced 3 significant changes.

First, I've splitted the requirements, we have now 3 kinds of requirements:
- base requirements (`install_requires`)
- `test` requirements (coverage, etc...), 
- `dev` requirements (at start included in requirements.txt). Do you agree the `dev` name section?

Users can deal with these like this:
```shell
$ pip install octodns # only base
$ pip install octodns[test] # only base + test
$ pip install octodns[dev] # only base + dev
$ pip install octodns[dev,test] # full install 
```

Second, I reverse version management, `setup.cfg` retrieve information from `__init__.py` I think it's more elegant way than older version. This change:
- remove python code
- simplify management
- securizing retrieve (package installed or not already installed...)

Third, define a minimal setuptools version `30.3.0`.

### Security note

This changes can help users to have more confidence to your package.

Often, inside private companies projects, they realize static analyzes of dependencies or inspect code before using it in their projects [to make sure it's not a malicious dependencies](https://github.com/pypa/pypi-legacy/issues/644). [A good example](http://www.nbu.gov.sk/skcsirt-sa-20170909-pypi/) of this fact is the typo squatting.

You are welcome!

requirements.txt
4383: I agree your remarks, read me comment and my propositions below.

4383:
@ross I've make a mistake when I force users to update setuptools.
Your package is already build and uploaded to pypi so users doesn't have to use setuptools in latest version. 

Only you and your contributors have really the constraint to use setuptools on recent version. Also if your users install octodns from sources they have to use an up-to-date version of setuptools because they have to implicitly build the package `python setup.py install`.

I move setuptools dependencie minimal version to `CONTRIBUTING.md` instead of `README.md`

ross:
> I move setuptools dependencie minimal version to CONTRIBUTING.md instead of README.md

:cool: That was my biggest concern. Otherwise just looked over this and I think it's gtg. Will pull it all down and play around with it when I get a bit of time and :shipit: provided I don't run into anything unexpected.

4383:
You're welcome!
Do not hesitate to asking help to me if you need :)

setup.cfg
```diff
@@ -0,0 +1,68 @@
+[metadata]
+name = octodns
+description = "DNS as code - Tools for managing DNS across multiple providers"
+long_description = file: README.md
+version = attr: octodns.__VERSION__
```
ross: Cool. This seems like a good option.

4383: Yep! No code, only retrieve at build/install and on a specified file, it's more an elegant way.
You're welcome!

PR #140, opened 2017-11-07
Add DigitalOcean provider

yzguy:
Add Digitalocean provider (minus tests atm)

Almost all the code is the same as Dnsimple surprisingly, just a little bit different data structure back from API.

Figure I would push this part up to get feedback, then continue working on tests

One weird thing I noticed, not sure if it's a bug, but I cannot dump CAA records from Digitalocean, NS1, or Route53. I can however create them.

Will be working on the tests.

#### Feedback about creating new provider
The process I found pretty easy for creating a new provider was:
* Create Client/Wrapper for API (if necessary)
* Create Provider Class beginnings (Supports, Geo, Constructor)
* Dump Zone
    * `populate` method
    * `_data_for` methods
* Apply
    * Create zone if none exists
    * `_apply` methods
    * `_params_for` methods

For dumping, when you do it and don't have the `_data_for` methods, you get an error for which one you need, then you can write it, try again, until it works.

For applying, you will get an error for the `_params_for` methods, when you get an error write it, try again, until you have it working.

Then you can just attempt to dump a zone, delete the zone, try and reapply the dumped zone


yzguy:
For the "paging" functionality in the client. If there is more than 1 page, there will always be a `last`, unless you are on the `last` page. By checking the existence of `last` on every page, you can quickly determine if you have more pages to request.

Page 1:

```
	"links": {
		"pages": {
			"last": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=3",
			"next": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=2"
		}
	},
```

Page 2:

```
	"links": {
		"pages": {
			"first": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=1",
			"prev": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=1",
			"last": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=3",
			"next": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=3"
		}
	},
```

Page 3:

```
	"links": {
		"pages": {
			"first": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=1",
			"prev": "https://api.digitalocean.com/v2/domains/yzguy.io/records?page=2"
		}
	},
```

README.md
```diff
@@ -151,6 +151,7 @@ The above command pulled the existing data out of Route53 and placed the results
 |--|--|--|--|
 | [AzureProvider](/octodns/provider/azuredns.py) | A, AAAA, CNAME, MX, NS, PTR, SRV, TXT | No | |
 | [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, CAA, CNAME, MX, NS, SPF, TXT | No | CAA tags restricted |
+| [DigitalOceanProvider](/octodns/provider/digitalocean.py) | A, AAAA, CAA, CNAME, MX, NS, TXT, SRV | No | |
```
yzguy: Not sure what the CAA tags restricted means so I didn't put it.

ross:
> One weird thing I noticed, not sure if it's a bug, but I cannot dump CAA records from Digitalocean, NS1, or Route53. I can however create them.

Hrm. You might check to see if there are log messages (hopefully warnings, but maybe not) about those records being skipped during the dump. If so that'd definitely be a bug of some sort.

README.md
ross: I believe that was supposed to mean that they support the record type, but have restrictions on the values beyond what the spec lays out. The spec on CAA lays out a way to add non-standard/extra values and to flip a flag saying that if a certificate authority doesn't understand them it shouldn't issue certs for the zone. I believe these messages are saying the providers don't support all of that, they just support the base/initial set of laid out value types.

octodns/provider/digitalocean.py
```diff
@@ -0,0 +1,333 @@
...
+    def domain(self, name):
+        path = '/domains/{}'.format(name)
+        return self._request('GET', path).json()
+
+    def domain_create(self, name):
+        path = '/domains'
+
+        # Digitalocean requires an IP on zone creation
+        resp = self._request('POST', path, data={'name': name,
+                                                 'ip_address': '127.0.0.1'})
```
ross: Probably a good idea to use a non-routable/never used IP address here just in case, anything in `192.0.2.0/24` will work as it's a [block reserved for documentation](https://tools.ietf.org/html/rfc5737).

octodns/provider/digitalocean.py
```diff
@@ -0,0 +1,333 @@
...
+
+        # After the zone is created, immeadiately delete the record
+        records = self.records(name)
```
ross: Weird, is that what the IP address above is used for? It just stick an A-record on the top? Weird.

If we're sure there's only going to be a single record here it could probably short-cut things a bit and do something like:

```python
self.record_delete(name, records[0]['id'])
```

What's there is a little safer in the case that there's more than one so it's not a problem.

octodns/provider/digitalocean.py
```diff
@@ -0,0 +1,333 @@
...
+            # https://developers.digitalocean.com/documentation/v2/#links
+            # if pages exists, there are more pages
+            # if it doesn't, there is just a single page
+            try:
+                pages = links['pages']
+                # if last exists, there are more pages
+                # if it doesn't, you're on the last page
+                try:
+                    pages['last']
+                    page += 1
```
ross: Can these two levels be combined into one? Something like:

```python
try:
    links['pages']['last']
    page += 1
except KeyError:
    break
```

octodns/provider/digitalocean.py
```diff
@@ -0,0 +1,333 @@
...
+        resp = self._request('POST', path, data={'name': name,
+                                                 'ip_address': '127.0.0.1'})
+
+        # After the zone is created, immeadiately delete the record
+        records = self.records(name)
+        for record in records:
+            if record['name'] == '' and record['type'] == 'A':
+                self.record_delete(name, record['id'])
+
+        return resp
```
ross: Doesn't look like this return value is being used where this is called below so probably OK to drop it. If not it might make sense to do a `.json()` on it so that it's not returning the http response, but just the data.

octodns/provider/digitalocean.py
yzguy: DigitalOcean automatically put's in their NS records, so most likely the A record will be `records[3]['id']`, but you can't guarantee what order the apex A record will be at. Only sure way is to search for them and delete them when found.

It is a very weird thing to require, especially when it's not required in the UI.

yzguy:
Is being skipped expected?

NSOne:
`2017-11-07T10:20:48  [140736773477312] DEBUG Zone changes:  skipping record=yzguy.io. CAA - dump does not support it`

DigitalOcean:
`2017-11-07T10:23:02  [140736773477312] DEBUG Zone changes:  skipping record=yzguy.io. CAA - dump does not support it`

Route53:
`2017-11-07T18:34:29  [140736773477312] DEBUG Zone changes:  skipping record=yzguy.io. CAA - dump does not support it`

ross:
> Is being skipped expected?

Probably not. Mind filing a separate issue about it. I'll try and dig into it at some point tomorrow.

yzguy:
Done: https://github.com/github/octodns/issues/141

yzguy:
Finally finished this up! Opened the issue almost a month ago :P

PR #142, opened 2017-11-08
Add formal CAA support to YamlProvider

### LINKED ISSUE #141: Dump does not support CAA records on Multiple Providers
After creating CAA records in the UI/API, then trying to dump the zone. This is happening across the 3 providers I have access to.

NSOne:
`2017-11-07T10:20:48 [140736773477312] DEBUG Zone changes: skipping record=yzguy.io. CAA - dump does not support it`

DigitalOcean:
`2017-11-07T10:23:02 [140736773477312] DEBUG Zone changes: skipping record=yzguy.io. CAA - dump does not support it`

Route53:
2017-11-07T18:34:29 [140736773477312] DEBUG Zone changes: skipping record=yzguy.io. CAA - `dump does not support it`

ross:
An oversight when adding `CAA` support. `YamlProvider` is special when it comes to supports since it always supports everything that octoDNS does. When the `CAA` record type was added to octoDNS it wasn't included in the provider's `SUPPORTS` list thus anything that relied on that list (the provider itself doesn't) would think `CAA` isn't supported. This adds it to the list and updates the tests to match

/cc Fixes #141 
/cc @yzguy thanks

### DIFF
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -31,8 +31,8 @@ class YamlProvider(BaseProvider):
         enforce_order: True
     '''
     SUPPORTS_GEO = True
-    SUPPORTS = set(('A', 'AAAA', 'ALIAS', 'CNAME', 'MX', 'NAPTR', 'NS', 'PTR',
-                   'SSHFP', 'SPF', 'SRV', 'TXT'))
+    SUPPORTS = set(('A', 'AAAA', 'ALIAS', 'CAA', 'CNAME', 'MX', 'NAPTR', 'NS',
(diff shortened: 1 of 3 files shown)


yzguy:
Confirmed working for a dump from NS1 and Route53:

`2017-11-11T16:03:50  [140736773477312] DEBUG Zone changes: zone=Zone<yzguy.io.>, create record=<CaaRecord CAA 3600, yzguy.io., ['0 issue "letsencrypt.org"']>`

Thanks for this! We/I don't actually use CAA records yet, but someone might for sure.

PR #148, opened 2017-11-13
Send appropriate meta along for A and AAAA records

scoursen:
Set and send the geo meta data for the A & AAAA records.


octodns/provider/ns1.py
```diff
@@ -164,7 +164,16 @@ def populate(self, zone, target=False, lenient=False):
...
+        params = {'answers': record.values, 'ttl': record.ttl}
+        if record.geo:
+            # purposefully set non-geo answers to have an empty meta,
+            # so that we know we did this on purpose if/when troubleshooting
+            params['answers'] = [{"answer": x, "meta":{}} for x in record.values]
+            for iso_region, target in record.geo.items():
+                params['answers'].append({'answer': target.values,
+                                          'meta': {'iso_region_code': [iso_region]},
+                })
+        return params
```
ross: Nice. Was this functionality always available or newly added to make this use-case easy? Definitely ideal API-wise. Would be interested to see what the backing implementation details look like to see if the behaviors match up with Dyn and Route's octoDNS implementation in terms of fallbacks etc. 

There'll need to be something similar on the way out in `_data_for_A` so that octoDNS can see the state of things to make sure they match up with how things are configured. 

Regardless happy to chat about the details if that's useful.

scoursen: There's some functionality change in the API server to support this.  Previously, NS1 used a different way of identifying geo regions internally, but an upcoming change will support using ISO region codes.  This PR addresses that the API will be able to accept this new data shortly.

scoursen:
Changes need to be made based on internal discussions.  Closing this PR for now, will open a new one soon.

PR #162, opened 2017-12-27
Major reworking of Cloudflare record update

ross:
CloudflareProvider's update path wasn't working correctly in many cases. This should get it to a point where it correctly functions, but it's still not perfect/ideal. It's doing about as good as it reasonably can given the API doesn't provide a way to update the multiple pieces of a single record atomically. More details in the comments. 

This is not yet finished. It'll need a pass of more thorough testing to get all of the bits covered and validated more thoroughly. 

- [x] Additional test coverage of new paths & logic.

/cc Fixes https://github.com/github/octodns/issues/144
/cc @jr-RXcode @NicolaiSchmid

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import, division, print_function, \
 
 from collections import defaultdict
 from logging import getLogger
+from json import dumps
 from requests import Session
 
 from ..record import Record, Update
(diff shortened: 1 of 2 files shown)


ross:
OK. Pushed up a pass at adding detailed tests for the new update code. This should be ready to go.

ross:
Since this was previously non-functional this PR surely can't be worse right :-) going to merge.

vanbroup:
Looks like this PR introduced some new issues with the Cloudflare plugin, the dump works fine but when you sync information back records with multiple values get deleted (only one remains).

ross:
> Looks like this PR introduced some new issues with the Cloudflare plugin, the dump works fine but when you sync information back records with multiple values get deleted (only one remains).

Odd. Its at this point that I'm glad octoDNS tells you what it's going to do before doing it. Will take a look at this, but it'll probably take me a couple days to get to it atm.

ross:
> Looks like this PR introduced some new issues with the Cloudflare plugin, the dump works fine but when you sync information back records with multiple values get deleted (only one remains).

@vanbroup do you have more details or repo steps for this? I just tried dumping a couple zones and then syncing them back and get `No changes were planned`. 

Checked both records with multiple values and names with multiple record types and didn't turn up any problems, but obviously it's hard to be exhaustive.

vanbroup:
When everything is in sync it works fine, I think the problem is caused by a combination of [L334](https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L334).and [L348](https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L348).

When one of (for example) three values is updated, it will update the first record that matches the name and type (line 348). The other two lines get deleted by line [371](https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L371).

ross:
Sorry @vanbroup, but I'm still unable to reproduce. If you have something easily reproducible can you drop the details in here or email them to me.

ross:
OK. That should be enough for me to dig in. I was previously looking for problems around the [process of dumping & then syncing](https://github.com/github/octodns/pull/162#issuecomment-362214435) and didn't turn anything up.

ross:
I couldn't get this to reproduce even with an exact copy of your case until I turned on proxying. 

Turns out it's a diff in the dump'd json between the TTLs. You can't hit it if you're running normal records b/c CF has the same min 120s TTL for all records so everything matches up and the 2 records are left alone. If instead you turn on the proxying stuff the API returns back answers with 1s TTLs which don't match up with the minimum TTL of 120s for the provider so the records aren't considered the same at that point. 

Ultimately I think this falls into the same bucket as https://github.com/github/octodns/issues/202, managing zones with proxied records just isn't supported currently. I'll get a PR going to update the README to call that out for Cloudflare to at least make that explicit. 

In general I don't have the use-case for CF protection and won't have the bandwidth to dig into it "for fun" any time soon so it'll probably have to find an owner to pick it up and work out the details/kinks.

vanbroup:
In my test case I had this problem even without any proxied records. Also as stated in comment https://github.com/github/octodns/pull/162#issuecomment-363337216 I think this also looks to be a logical behavior of the code if the records are not returned in fixed order.

ross:
> In my test case I had this problem even without any proxied records.

Putting a TTL of 1 in the config is completely ignored by the CF provider and raised up to 120s. The only way I've found to recreate the output you get is if I go into the CF UI and click automatic on the records so that they return 1s TTL which can't be matched by the provider that ensures a minimum of 1. I'll play around with it a bit more, but since the "existing" record version says 1s I'm pretty sure that means automatic was turned on for the record unless you have something enabled in your account that I don't.

> I think this also looks to be a logical behavior of the code if the records are not returned in fixed order.

I'm not following that. `existing_contents` and `new_contents` are hashes so ordering shouldn't be of consequence. I currently think it's running into problems because all of the contents are out of whack: existing have 1s ttl as returned by CF and new have 120s as bumped up from 1s by the provider.

I'm still trying, but I can't find a way to get a 1s TTL back from the API without having automatic enabled and having 1s in the config will always be ignored. https://github.com/github/octodns/blob/master/octodns/provider/cloudflare.py#L325

ross:
Filed https://github.com/github/octodns/pull/205 which should address the TTL mis-matches, but the 1s (Automatic) TTL stuff still isn't supported in general. More info in the issue. We should move any further discussion there.

PR #164, opened 2017-12-29
Add geo support for NS1 provider

scoursen:
Support geo information for A and AAAA records using the NS1 provider.  The zone needs to be created through other calls (either the NS1 portal or directly to the NS1 API) and geo filter chain needs to be created (again, either through the NS1 portal or direct via API).  

If the geo filter chain is created, records with iso region codes will be used as appropriate.

Any records created outside of octodns (via NS1 portal or NS1 API) will be overwritten when using `octodns-sync` as the yml files are considered the source of truth if using octodns.


markmpeterson:
@yzguy @ross or other project collaborators, could you assist us with review/merging of this pr? Just wanted to raise it to your attention. Thank you!

octodns/provider/ns1.py
```diff
@@ -146,25 +175,46 @@ def populate(self, zone, target=False, lenient=False):
...
                 raise
             records = []
+            geo_records = []
 
         before = len(zone.records)
-        for record in records:
+        # geo information isn't returned from the main endpoint, so we need
+        # to query for all records with geo information
+        zone_hash = {}
+        for record in chain(records, geo_records):
```
ross: Hadn't run across `chain` before. That's handy.

octodns/provider/ns1.py
```diff
@@ -146,25 +175,46 @@ def populate(self, zone, target=False, lenient=False):
         try:
             nsone_zone = self._client.loadZone(zone.name[:-1])
             records = nsone_zone.data['records']
+            geo_records = nsone_zone.search(has_geo=True)
```
ross: Would things with `has_geo=True` not show up in `.data['records']` or is this relying on the order of processing below to replace the plain record with the geo version?

octodns/provider/ns1.py
```diff
@@ -146,25 +175,46 @@ def populate(self, zone, target=False, lenient=False):
...
+            for iso_region, target in record.geo.items():
+                key = 'iso_region_code'
+                value = iso_region
+                params['answers'].append(
+                    {
+                        'answer': target.values,
+                        'meta': {key: [value]},
+                    },
+                )
+        self.log.info("params for A: %s", params)
```
ross: Probably should be `debug` level.

octodns/provider/ns1.py
```diff
@@ -35,11 +37,38 @@ def __init__(self, id, api_key, *args, **kwargs):
...
         }
+        values, codes = [], []
+        if 'answers' not in record:
+            values = record['short_answers']
+        for answer in record.get('answers', []):
+            meta = answer.get('meta', {})
+            if meta:
+                country = meta.get('country', [])
+                us_state = meta.get('us_state', [])
+                ca_province = meta.get('ca_province', [])
```
ross: Is it possible for multiple of these to be non-default? E.g.

```json
{
  "country": "US",
  "us_state": "WA"
}
```

If so it seems like this would set both `geo['US']` and `geo['US-WA']`. If only `us_state` would be set and not  country for cases where a US state was targeted then this would work, in which case a comment to that effect would probably make sense.

octodns/provider/ns1.py
scoursen: things with has_geo=True do show up in .data["records"], but not with their geo information.  Doing the `nsone_zone.search(has_geo=True)` gets us the records, along with their geo information.

octodns/provider/ns1.py
scoursen: It's possible, but we collapse things, so if we receive both country of US and state of WA, we only consider the more specific state of WA.  will add comment clarifying that.

scoursen:
> What would the update workflow look like? It sounds like corresponding changes need to be made outside of the octoDNS sync run?

That is no longer true.  There was an addition to this pull request that simplifies the workflow.  Specifying a ISO code in your config YAML file will create all the appropriate settings with the NS1 API

ross:
> That is no longer true. There was an addition to this pull request that simplifies the workflow. Specifying a ISO code in your config YAML file will create all the appropriate settings with the NS1 API

Oh, awesome. I poked around in the UI a bit to try and figure out how that'd work, but couldn't suss it out so figure I was missing something. In that case I like how this turned out :-), definitely makes things clean/easy for octoDNS and imo for anyone else working with the API.

I'll stick this on my TODO list. Probably won't get to it this weekend, but 🤞 early next week.

octodns/provider/ns1.py
ross: Looks like `.search` was added recently so `setup.cfg` will likely need to update the required version of `nsone`. I was getting test failures until I bumped it.

octodns/provider/ns1.py
```diff
@@ -35,11 +37,38 @@ def __init__(self, id, api_key, *args, **kwargs):
...
+        # record meta (which would include geo information is only
+        # returned when getting a record's detail, not from zone detail
+        geo = {}
+        data = {
             'ttl': record['ttl'],
             'type': _type,
-            'values': record['short_answers'],
         }
+        values, codes = [], []
+        if 'answers' not in record:
```
ross: Most of this section of code doesn't have code coverage in the tests. `./script/coverage` will run the tests with coverage and drop the results in `./htmlcov`.

scoursen: added test mocks to get coverage back up.

ross:
> added test mocks to get coverage back up.

:cool: Looks like there's a couple of lint (pep8) warnings and CI will be happy. 

Any thoughts on what's up with the `ResourceException: server error` bit? That should be the last thing prior to merge.

scoursen:
Fixed the serialization in `_params_for_A` when dealing with multiple answers for an A or AAAA record.  This was the cause of the `ResourceException`

octodns/provider/ns1.py
```diff
@@ -35,11 +37,42 @@ def __init__(self, id, api_key, *args, **kwargs):
...
+                for cntry in country:
+                    cn = transformations.cc_to_cn(cntry)
+                    con = transformations.cn_to_ctca2(cn)
+                    geo['{}-{}'.format(con, cntry)] = answer['answer']
+                for state in us_state:
+                    geo['NA-US-{}'.format(state)] = answer['answer']
+                for province in ca_province:
+                    geo['NA-CA-{}'.format(province)] = answer['answer']
+                for code in meta.get('iso_region_code', []):
+                    geo[code] = answer['answer']
```
ross: Bit more testing has pointed out an issue in the code here. Each each value comes back from the system as a seperate answer. Because of that we get multiple trips around this loop for anything with multiple values, with the subsequent rounds replacing the previous value.

scoursen:
I pushed the serializing change (when sending to the API), but had forgotten to commit/push the corresponding fix to deserializing the API responses.

octodns/provider/ns1.py
```diff
@@ -35,11 +38,62 @@ def __init__(self, id, api_key, *args, **kwargs):
...
+                # be redundant since the country would supercede in when
+                # resolving the record.  it is syntactically valid, however.
+                country = meta.get('country', [])
+                us_state = meta.get('us_state', [])
+                ca_province = meta.get('ca_province', [])
+                for cntry in country:
+                    cn = transformations.cc_to_cn(cntry)
+                    con = transformations.cn_to_ctca2(cn)
+                    key = '{}-{}'.format(con, cntry)
+                    if key not in geo:
```
ross: You could avoid having to do the `if key not in geo ... else ...` stuff in each block here by using `geo = defaultdict(list)` from `collections`.  then the `geo[key].extend(answer['answer']) should be enough.

scoursen: d'oh.  i forgot about defaultdict(list).

ross:
:cool: Will poke around and test a bit more in the morning and get this merged assuming that goes smoothly.

PR #165, opened 2018-01-05
Add RackSpace as a provider

terrence2:
Add RackSpace to the list of supported providers. This is based on RackSpace's version 1 API. We have been using this in production daily for over three months now and feel confident that it is ready for general use.


octodns/provider/rackspace.py
```diff
@@ -0,0 +1,385 @@
...
+    return string.replace(s, '\;', ';')
+
+
+class RackspaceProvider(BaseProvider):
+    SUPPORTS_GEO = False
+    SUPPORTS = set(('A', 'AAAA', 'ALIAS', 'CNAME', 'MX', 'NS', 'PTR', 'SPF',
+                    'TXT'))
+    TIMEOUT = 5
+
+    def __init__(self, id, username, api_key, ratelimit_delay, *args,
```
ross: Should `ratelimit_delay` have a default? The example config below doesn't specify a value for it, but it appears to be a required param here.

octodns/provider/rackspace.py
```diff
@@ -0,0 +1,385 @@
...
+                                                       "apiKey": api_key}}},
+                   )
+        cloud_dns_endpoint = \
+            [x for x in ret.json()['access']['serviceCatalog'] if
+             x['name'] == 'cloudDNS'][0]['endpoints'][0]['publicURL']
+        return ret.json()['access']['token']['id'], cloud_dns_endpoint
+
+    def _get_zone_id_for(self, zone):
+        ret = self._request('GET', 'domains', pagination_key='domains')
+        time.sleep(self.ratelimit_delay)
```
ross: It'd probably be more robust to move `time.sleep(self.ratelimit_delay)` into `_request` so that all requests would be covered by the single invocation rather than having it copied around to each of the various ways requests are made.

README.md
```diff
@@ -158,6 +158,7 @@ The above command pulled the existing data out of Route53 and placed the results
 | [Ns1Provider](/octodns/provider/ns1.py) | All | No | |
 | [OVH](/octodns/provider/ovh.py) | A, AAAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT | No | |
 | [PowerDnsProvider](/octodns/provider/powerdns.py) | All | No | |
+| [Rackspace](/octodns/provider/rackspace.py) | A, AAAA, ALIAS, CNAME, MX, NS, PTR, SPF, TXT | No |  |
```
ross: This cover everything they support or a partial list based on what the provider has built out?

octodns/provider/rackspace.py
```diff
@@ -0,0 +1,385 @@
...
+            return [record.value]
+
+    def _mod_Create(self, change):
+        return self._create_given_change_values(change,
+                                                self._get_values(change.new))
+
+    def _create_given_change_values(self, change, values):
+        out = []
+        for value in values:
+            transformer = getattr(self,
```
ross: Not a huge deal, but this line can come out of the for loop so that it only happens once for each change.

octodns/provider/rackspace.py
```diff
@@ -0,0 +1,385 @@
...
+
+        if updates:
+            data = {"records": sorted(updates, key=lambda v: v['name'])}
+            self._put('domains/{}/records'.format(domain_id), data=data)
+
+        if creates:
+            data = {"records": sorted(creates, key=lambda v: v['type'] +
+                                      v['name'] +
+                                      v.get('data', ''))}
+            self._post('domains/{}/records'.format(domain_id), data=data)
```
ross: Are the creates, updates, and deletes broken out into 3 requests due to constraints or could they all be sent in a single request? 

It can be safer to make all the changes in one go if the API supports it since you could be deleting an A record to replace it with a CNAME and if the delete succeeds, but the update doesn't things would be in a "broken" state where neither record would exist. This is a pretty common flaw in the DNS APIs that either don't support batch updates (doesn't appear to be applicable here) or don't correctly do validations on the result of the whole batch. Just want to make sure this is doing things as safely as reasonably possible. I've been slogging through the same mess with the Cloudflare provider in https://github.com/github/octodns/pull/162. Some providers are "safer" than others in this respect.

Separately are there any (obvious/published) size limits to the request data size?

README.md
terrence2: This is just what we are confident we support fully. RackSpace supports more record types than this, however, we found enough idiosyncrasies in these that we didn't want to try adding any record types we didn't have extensive testing for. It should be easy to add more support in future if someone has a need and enough good test data to check their assumptions.

octodns/provider/rackspace.py
terrence2: RackSpace has a REST API that treats individual record-set entries as resources. To say that this makes for an absolutely *terrible* DNS API is a dramatic understatement. In this specific case, please note that delete, update, and create require DELETE, PUT, and POST HTTP methods, respectively, so simply cannot be done in a single request. Combine this with the one-address-per-"record" rule and the fact that they want their internal `id` for most operations instead of the record data itself and you end up with the nightmare fuel  in _mod_Update. I think this went through 4 or 5 refactors before we finally got all of the constraints addressed well enough for update to work reliably.

Speaking of safety, I'd settle for merely correct. Take a look at the uses of {add,remove}_trailing_dot. The RackSpace API does not allow you to have a trailing dot in the request at all for most (but not all) record types, however internally treats most records as if they had the trailing dot.

It took a lot of trial and error to figure out what their API even wanted in most cases. I think we ended up doing fairly exhaustive testing around the various supported record types -- most of the test cases were built from actual failures we encountered in our own records or in tests we built to explore these edge cases.

ross: Sorry, I didn't catch the delete/put/post bit. Makes sense given the constraints and is inline with some of the other API's. If you don't mind just drop a comment in before the 3 calls with a quick bit about why it's that way, how it's not particularly safe, but making the best of things etc.  Mostly so that if others see it in the future or run into problems it'll be called out in place.

terrence2: Also, there do not appear to be size limits on individual requests; all of their throttling appears to be around number and frequency of requests.

terrence2:
:+1: Thanks for all of the fast and thoughtful feedback!

brianeclow:
🎊 Thank you.

PR #169, opened 2018-01-09
RateLimitException.period is coming back as str now

ross:
Found this when testing a very large/complicated zone and hitting rate limit exceptions. 

/cc @scoursen for 👍  / 👎  since I can't explicitly request one from you


scoursen:
:+1:

ross:
Thanks.

PR #170, opened 2018-01-09
Set Ns1Provider to "Yes" for GeoDNS Support

scoursen:


ross:
Actually I think we need to add a caveat comment to this line about health checking and fallback support since it doesn't currently match the Route53 & Dyn behaviors.  I started looking around at the filter chain stuff last night, but don't yet understand it enough to spell out what it would look like. For now a comment should suffice, I didn't want to extend the Geo PR itself further to try and work it all out, but happy to chat about what the "standard" behavior looks like and help figure out what it'd look like with filter chains.

PR #171, opened 2018-01-09
Comment that NS1 geo health checks aren't yet supported

ross:
/cc @scoursen 
/cc #170


scoursen:
NS1 supports health checks, but its orthogonal to geo dns and yes would require some discussion about what the "standard" behavior should be.

ross:
> NS1 supports health checks, but its orthogonal to geo dns and yes would require some discussion about what the "standard" behavior should be.

Yeah looked around a bit and see the "up" filter, but it wasn't immediately clear if/how that would be configured in terms of what to check. Feel free to email me (addr on profile) if you want to set up a time to chat about it and see what we can figure out.

Issue #150, opened 2017-11-16
Add support for DNS Made Easy


davidmmiller:
Can you please look into supporting DNS Made Easy DNS?

http://www.dnsmadeeasy.com/integration/restapi/

yzguy:
Will need to look into it. The sandbox thing might be adequate for building it, or maybe a 30 day trial.

It “should” be relatively easy if the API is decent.

yzguy:
Totally do-able. I already started work on it. They have a pretty weird authentication scheme, but figured that out.

I think initially will probably just be records, then after Geo can be added later to the provider.

oryagel:
I would love to see that as well. I'm in the process of migrating from DME just because they don't have OctoDNS provider.

yzguy:
I have the "provider" relatively complete. DME has some quirks to deal with (Relative values, Cannot make NS record without there being an A record first, etc.) that I'm trying to deal with.

Then the rest would just be adding tests.

**Couple questions:**
Do you use ANAME records?
Do you use relative values?

oryagel:
We don't have any ANAME records.
By relative values, do you mean domain records that points to the same domain without specifying the domain part? We try not to have these. In my opinion, you can ignore this case.

yzguy:
Great. I'm probably not going to add in ANAME records because I think they are only specific to DNSMadeEasy.

As for relative, yes you're right on with your description. I'm handling it with dumping relative values, then making them absolute, so they will be absolute in the YAML. If you re-apply the YAML it will delete/create the record again with the absolute value. 

That delete/create behavior could be dicey, as there is a period where the record wouldn't exist.

yzguy:
@ross Would like to bounce a couple questions/thoughts off you

1. DNSMadeEasy allows you to specify relative values, it returns them from the API as relative. Currently, I'm handling it with dumping relative values via the DnsMadeEasyClient, then making them absolute before returning them from the caller, so they will be absolute in the YAML. If you re-apply the YAML it will delete the relative record from the provider and create the record again with the absolute value.

This delete/create behavior could be dicey, as there is a period where the record wouldn't exist. Wondering if you would consider that acceptable and if a note just saying the behavior of the provider, so users are aware.

2. DNSMadeEasy will not allow you to create a NS record with a relative value if there is no A record already created for it. It will however allow you to create a NS record with an absolute value of a record in the same zone regardless if the A record exists. The latter is the behavior I have written in the client to force absolute values.

[...]

PR #175, opened 2018-01-21
Adds a new field Plan.exists to indicate whether the zone exists

ross:
Uses the field to indicate whether the zone will be created when doing plan outputs. As of right now it just includes a new row Saying "Create Zone<...>" for each provider where it would be created. If that row isn't there it already exists. 

The new behavior for `populate` has been implemented for all providers except OVH which doesn't seem to have any indication as to whether the zone exists or not, looking at the code I'm a little worried the case of a zone not exists isn't handled by the provider, unless the API just returns an empty record in that case. @trnsnt would you be able to take a look at that. Feel free to PR off of this branch and I'll merge it in if that lands soon. Otherwise we can go forward without it.

Fixes https://github.com/github/octodns/issues/174
/cc @yzguy

PR #178, opened 2018-01-23
Dnsmadeeasy provider

yzguy:
Figure I'd get this up here to get some feedback while I work on the tests (~73% coverage right now)

Add DNSMadeEasy provider

* No support for Geo or ANAME records
* Allows user to toggle sandbox environment
* Added zone creation detection (#175)

I have notes for investigating paging, I added 2500 records and I could not get a response that had `totalPages > 1`. It was always 1 response, with all the records. I'm not sure what the limit is where it will start paging. Opening a support ticket with DnsMadeEasy to find out if there is pagination at all.

Fixes: #150


yzguy:
Response from their support:

```
This feature was added based on our users request. By default we will return all records, but you do have the ability to paginate the responses as well in smaller chunks if necessary.

Example:
https://api.dnsmadeeasy.com/V2.0/dns/managed/<domain_id>/records/?rows=5&page=4
https://api.dnsmadeeasy.com/V2.0/dns/managed/<domain_id>/records/?rows=5&page=3
```

So is it worth implementing pagination now to avoid an issue due to large response size, etc down the road.

octodns/provider/dnsmadeeasy.py
```diff
@@ -0,0 +1,396 @@
...
+    PRODUCTION = 'https://api.dnsmadeeasy.com/V2.0/dns/managed'
+    SANDBOX = 'https://api.sandbox.dnsmadeeasy.com/V2.0/dns/managed'
+
+    def __init__(self, api_key, secret_key, sandbox=False,
+                 ratelimit_delay=0.0):
+        self.api_key = api_key
+        self.secret_key = secret_key
+        self._base = self.PRODUCTION
+        if sandbox:
+            self._base = self.SANDBOX
```
ross: These 3 lines can optionally be combined into `self._base = self.SANDBOX if sandbox else self.PRODUCTION`

octodns/provider/dnsmadeeasy.py
```diff
@@ -0,0 +1,396 @@
...
+    def _request(self, method, path, params=None, data=None):
+        now = self._current_time()
+        hmac_hash = self._hmac_hash(now)
+
+        headers = {
+            'x-dnsme-apiKey': self.api_key,
+            'x-dnsme-hmac': hmac_hash,
+            'x-dnsme-requestDate': now
+        }
+        self._sess.headers.update(headers)
```
ross: Some of these look like they're session-level and others request specific. 

You can probably do `self._sess.headers.update({'x-dnsme-apiKey': api_key})` up in the `__init__` and avoid having to keep it around as an object property. 

The the `hmac` and `requestDate` can then be passed as `headers=headers` down in the `self._sess.request` call.

yzguy:
Make changes :) Thanks for the feedback. Just working on tests now.

What are your thoughts on implementing pagination mentioned above, rather than just a "give me all the records". Do you think there could be any issue in the future with large zones + huge response from API?

ross:
> What are your thoughts on implementing pagination mentioned above, rather than just a "give me all the records". Do you think there could be any issue in the future with large zones + huge response from API?

So there's pagination support, but the default page size is really large or something and you haven't hit it? Don't have a strong opinion in general. People are pretty good at hitting edge cases and have definitely found pagination problems in other providers so there's at least some chance that'll happen here, but if their API doesn't give any indication as to what it's size/behaviors are it might be fine to wait and see if/when someone hits it.

yzguy:
That's kinda what I'm thinking, it should be easy enough to add it in around where its grabbing the records on the first response if it becomes an issue. It's nice having 1 request for all of them, because there is rate limiting on their API.

yzguy:
After months of procrastinating, I have finally finished this first iteration of the DnsMadeEasy provider.

Tests are finished, all passing, and 100% test coverage.

Limitations currently are:
* No ANAME support
* No Geo Support

ross:
> After months of procrastinating, I have finally finished this first iteration of the DnsMadeEasy provider.

😁 Cool. Will take a look at things and review it this evening.

ross:
This was the last thing I was holding off for to land before cutting a new release. Will look to do that this week at some point.

yzguy:
Thank you for your help :) Really glad I'm able to contribute, definitely a project I've very interested in.

PR #182, opened 2018-01-31
Option to handle Cloudflare proxied records

vanbroup:
This change imports records that are marked as proxied so that they can be synced to other DNS providers as described in [this support acticle](https://support.cloudflare.com/hc/en-us/articles/115000830351-How-to-configure-DNS-for-CNAME-partial-setup-when-managing-DNS-externally). Records that use this functionality will be ignored by this provider and not be synced back to Cloudflare as we don't know the origin record values that would be required.

This change does not allow you to enable, disable or configure the CDN itself as that would require a lot of metadata to be handled by OctoDNS. 

The intention of this change is to allow users to run a multi-DNS provider setup without sending any traffic to their origin directly.

Resolves https://github.com/github/octodns/issues/65

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -14,14 +14,14 @@ from ..record import Record, Update
 from .base import BaseProvider
 
 
-class CloudflareAuthenticationError(Exception):
+class CloudflareError(Exception):
 
     def __init__(self, data):
         try:
             message = data['errors'][0]['message']
         except (IndexError, KeyError):
             message = 'Authentication error'
-        super(CloudflareAuthenticationError, self).__init__(message)
+        super(CloudflareError, self).__init__(message)
 
 
 class CloudflareProvider(BaseProvider):
@@ -34,6 +34,12 @@ class CloudflareProvider(BaseProvider):
         email: dns-manager@example.com
         # The api key (required)
         token: foo
+        # Import CDN enabled records as CNAME to {}.cdn.cloudflare.net. Records
+        # ending at .cdn.cloudflare.net. will be ignored when this provider is
+        # not used as the source and the cdn option is enabled.
+        #
+        # See: https://support.cloudflare.com/hc/en-us/articles/115000830351
+        cdn: false
     '''
     SUPPORTS_GEO = False
     # TODO: support SRV
@@ -43,9 +49,10 @@ class CloudflareProvider(BaseProvider):
     MIN_TTL = 120
     TIMEOUT = 15
 
-    def __init__(self, id, email, token, *args, **kwargs):
+    def __init__(self, id, email, token, cdn=False, *args, **kwargs):
         self.log = getLogger('CloudflareProvider[{}]'.format(id))
-        self.log.debug('__init__: id=%s, email=%s, token=***', id, email)
+        self.log.debug('__init__: id=%s, email=%s, token=***, cdn=%s', id,
+                       email, cdn)
         super(CloudflareProvider, self).__init__(id, *args, **kwargs)
 
         sess = Session()
@@ -53,6 +60,7 @@ class CloudflareProvider(BaseProvider):
             'X-Auth-Email': email,
             'X-Auth-Key': token,
         })
+        self.cdn = cdn
         self._sess = sess
 
         self._zones = None
@@ -65,8 +73,8 @@ class CloudflareProvider(BaseProvider):
         resp = self._sess.request(method, url, params=params, json=data,
                                   timeout=self.TIMEOUT)
         self.log.debug('_request:   status=%d', resp.status_code)
-        if resp.status_code == 403:
-            raise CloudflareAuthenticationError(resp.json())
+        if resp.status_code == 400 or resp.status_code == 403:
+            raise CloudflareError(resp.json())
         resp.raise_for_status()
         return resp.json()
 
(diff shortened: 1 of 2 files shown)


vanbroup:
@ross any comments on this PR?

octodns/provider/cloudflare.py
```diff
@@ -14,14 +14,14 @@
...
-class CloudflareAuthenticationError(Exception):
+class CloudflareError(Exception):
 
     def __init__(self, data):
         try:
             message = data['errors'][0]['message']
         except (IndexError, KeyError):
             message = 'Authentication error'
-        super(CloudflareAuthenticationError, self).__init__(message)
+        super(CloudflareError, self).__init__(message)
```
ross: Since removing it would be an API regressions I think I'd prefer to have `CloudflareAuthenticationError ` be renamed `CloudflareError` (as you've done) and then have a child class `CloudflareAuthenticationError(CloudflareError)`. You could then just treat 403 and 400 distinctly down below.

octodns/provider/cloudflare.py
```diff
@@ -170,8 +190,8 @@ def zone_records(self, zone):
         return self._zone_records[zone.name]
 
     def populate(self, zone, target=False, lenient=False):
-        self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
-                       target, lenient)
+        self.log.debug('populate: name=%s, cdn=%s, target=%s, lenient=%s',
```
ross: Don't think it makes sense to have `cdn` in here, these generally just print the params directly to the function. Whether or not it's enabled show be provided by the `__init__` debug line.

octodns/provider/cloudflare.py
```diff
@@ -250,6 +276,12 @@ def _gen_contents(self, record):
         if _type == 'ALIAS':
             _type = 'CNAME'
 
+        # If this is a record to enable to Cloudflare CDN don't update as we
+        # don't know the original values.
+        if (self.cdn and _type == 'CNAME' and
+           record.value.endswith('.cdn.cloudflare.net.')):
+            raise StopIteration
```
ross: There's an existing provider mechanism called `_include_change` that's may be more suited to do this sort of omission of specific things.

https://github.com/github/octodns/blob/75403e39f9d721b23fc5cc0ca8243b37543f56a2/octodns/provider/base.py#L29-L34

In fact `CloudflareProvider` actually uses it already to handle the case of their minimum `ttl` being `120s`. 

https://github.com/github/octodns/blob/75403e39f9d721b23fc5cc0ca8243b37543f56a2/octodns/provider/cloudflare.py#L216-L223

You could probably expand the checks there to include the bits you're looking for here and `return False` if there's a match. Both the new case (and probably the existing TTL one) could do with a `self.log.info` to tell people it's happening. Thoughts?

vanbroup:
@ross Updated PR according to your comments, let me know if there are any further changes required.

octodns/provider/cloudflare.py
```diff
@@ -186,14 +213,29 @@ def populate(self, zone, target=False, lenient=False):
...
 
-                    data_for = getattr(self, '_data_for_{}'.format(_type))
-                    data = data_for(_type, records)
                     record = Record.new(zone, name, data, source=self,
                                         lenient=lenient)
+
+                    # only one rewrite is needed for names where the proxy is
+                    # enabled at multiple records with a different type but
+                    # the same name
+                    if (self.cdn and records[0]['proxied'] and
```
ross: Interesting case.

octodns/provider/cloudflare.py
```diff
@@ -65,8 +77,11 @@ def _request(self, method, path, params=None, data=None):
         resp = self._sess.request(method, url, params=params, json=data,
                                   timeout=self.TIMEOUT)
         self.log.debug('_request:   status=%d', resp.status_code)
+        if resp.status_code == 400:
+            raise CloudflareError(resp.json())
```
ross: There doesn't appear to be any test cases running into this 400 case. Should be relatively simple to get one in place and then this will be good to go.

vanbroup:
@ross added a new test and updated the existing ones

octodns/provider/cloudflare.py
```diff
@@ -206,6 +248,20 @@ def _include_change(self, change):
...
+        if (hasattr(change.new, '_type') and
+            (change.new._type == 'CNAME' or
+             change.new._type == 'ALIAS') and
+                change.new.value.endswith('.cdn.cloudflare.net.')):
+            return False
+        if (hasattr(change.existing, '_type') and
+            (change.existing._type == 'CNAME' or
+             change.existing._type == 'ALIAS') and
+                change.existing.value.endswith('.cdn.cloudflare.net.')):
+            return False
```
ross: Sorry didn't really catch this in the last round. I assume the `hasattr` stuff here is around trying to handle changes that don't have `new` or `existing` by testing to see if `_type` is there, which it won't be on `None`?

If so it can be simplified to something like:

```python
if (change.record._type in ('ALIAS', 'CNAME') and
    change.record.value.endswith('.cdn.cloudflare.net.')):
    return False
```

The `.record` property was added to the `Change` object to make cases like this easier to deal with by always returning a record from the change object regardless of what type of change it is.

Otherwise this looks gtg. Soon as that comes through I'll :shipit:

vanbroup:
@ross good point, this is much cleaner, thanks!

Code updated.

PR #185, opened 2018-02-02
Add SRV support to Cloudflare provider

vanbroup:
Cloudflare provider was missing SRV support, this PR adds support for this record type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The above command pulled the existing data out of Route53 and placed the results
 | Provider | Record Support | GeoDNS Support | Notes |
 |--|--|--|--|
 | [AzureProvider](/octodns/provider/azuredns.py) | A, AAAA, CNAME, MX, NS, PTR, SRV, TXT | No | |
-| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, ALIAS, CAA, CNAME, MX, NS, SPF, TXT | No | CAA tags restricted |
+| [CloudflareProvider](/octodns/provider/cloudflare.py) | A, AAAA, ALIAS, CAA, CNAME, MX, NS, SPF, SRV, TXT | No | CAA tags restricted |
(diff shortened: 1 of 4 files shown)


ross:
Fwiw I had originally run into problems with Cloudflare being more stringent than other dns providers on what they would allow for the name of a SRV record. I think maybe they require it to conform to the `_service._protocol...` name format, but everyone else would allow arbitrary names. 

I'm 👍  on adding support, I just hadn't gotten around to figuring out how we should handle the case of other providers allowing the record, but CloudFlare rejecting it. Possibilities I'd thought about:

* Adding octoDNS validations that require the "standards" be adhered to (need to make sure this can be ignored)
* Adding code specifically to the Cloudflare provider during the plan phase that would detect that there's going to be a problem and blow up
* Nuanced SRV support for CloudFlare that would sync things that conform and ignore things that don't

I'd probably lean towards the first option there as it seems to align with octoDNS general encouragement to follow the standards. If we want to move forward with this prior to getting validations in place it'd probably make sense to add a comment to the Cloudflare provider's notes section in the README. Thoughts?

vanbroup:
I'm absolutely for following the standards, Cloudflare looks to follow RFC 2782 here too. I agree that option one to add a validator is the best option.

vanbroup:
@ross I was just looking at the validator for SRV records and it looks like there is already some validation on the name format.

https://github.com/github/octodns/blob/master/octodns/record.py#L843
`re.compile(r'^_[^\.]+\.[^\.]+')`

While I think we can change this to: `re.compile(r'^_[^\.]+\._[^\.]+')` as the RFC states:

> The symbolic name of the desired protocol, with an underscore
>         (_) prepended to prevent collisions with DNS labels that occur
>         in nature.

These looks to be no validation of the record data but that should make no difference between providers or prevent this PR to be merged.

I think it's fine to merge this PR and create an issue to address SRV content value validation.

What do you think?

ross:
> What do you think?

Doh. I guess I added that in the big validation pass and it didn't occur to me to go back and update CloudflareProvider since that would have resolved the original issue. This looks good. Will poke at it a bit now and get it merged.

PR #191, opened 2018-02-12
Support for SSHFP ECDSA (3) and SHA-256 (2)

ross:
RFC 6594 & RFC 7479

Fixes https://github.com/github/octodns/issues/180
/cc @dschaper please take a look and test it out


dschaper:
LGTM, thanks!

PR #194, opened 2018-02-13
Escape unescaped semicolons coming out of Google Cloud DNS

mochipon:
Google Cloud DNS seems to reply `;` not `\;` just like Route53 and NS1. Therefore, escaped records are always forced to updated.

This PR tries to solve the problem in the same way as #59.


nunnun:
Hi, Could you please review this PR? I have same issue and I think this PR should resolve.

ross:
Totally missed this coming through. I'll put it on my TODO list and hopefully get the chance to look into it soon.

PR #195, opened 2018-02-13
Properly set filter rules for geo answers

scoursen:
Only set filter rules when geo target information is configured.  If the answer has no geo target, don't set the rules.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -234,10 +234,9 @@ class Ns1Provider(BaseProvider):
                     )
             params['filters'] = []
             if has_country:
-                if len(params['answers']) > 1:
-                    params['filters'].append(
-                        {"filter": "shuffle", "config": {}}
-                    )
+                params['filters'].append(
+                    {"filter": "shuffle", "config": {}}
+                )
                 params['filters'].append(
                     {"filter": "geotarget_country", "config": {}}
                 )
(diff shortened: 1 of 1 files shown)


vanbroup:
@ross, verified and confirmed working ok

octodns/provider/ns1.py
```diff
@@ -233,11 +233,11 @@ def _params_for_A(self, record):
                         },
                     )
             params['filters'] = []
-            if len(params['answers']) > 1:
-                params['filters'].append(
-                    {"filter": "shuffle", "config": {}}
-                )
             if has_country:
+                if len(params['answers']) > 1:
```
ross: Looks like this change probably needs a corresponding tweak in the tests as this `if` always evaluates to true.

ross: Would removing it still apply a "shuffle" filter to records with a single answer? Not sure that'd be a problem, but I assume that was the original intent behind the `if > 1`

scoursen:
From my understanding, the shuffle filter needs to be present if there is
any geotarget involved.

scoursen:
The original intent was to handle shuffle when there was multiple answers
(regardless of geotarget).  That seems to have been a misunderstanding of
our internal api

ross:
> The original intent was to handle shuffle when there was multiple answers
> (regardless of geotarget).  That seems to have been a misunderstanding of
> our internal api

OK.

PR #197, opened 2018-02-14
Check Route 53 records against all supported types

mvermaes:
As discussed at https://github.com/github/octodns/pull/111#issuecomment-365614180, a small fix to check Route 53 records are of a type contained in the `SUPPORTS` set.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -461,7 +461,7 @@ class Route53Provider(BaseProvider):
                 record_name = zone.hostname_from_fqdn(rrset['Name'])
                 record_name = _octal_replace(record_name)
                 record_type = rrset['Type']
-                if record_type == 'SOA':
+                if record_type not in self.SUPPORTS:
                     continue
                 data = getattr(self, '_data_for_{}'.format(record_type))(rrset)
                 records[record_name][record_type].append(data)


octodns/provider/route53.py
```diff
@@ -459,7 +459,7 @@ def populate(self, zone, target=False, lenient=False):
                 record_name = zone.hostname_from_fqdn(rrset['Name'])
                 record_name = _octal_replace(record_name)
                 record_type = rrset['Type']
-                if record_type == 'SOA':
+                if record_type not in self.SUPPORTS:
```
ross: Does it make sense to try and detect the Route53 alias records here to and ignore them?

mvermaes: Sorry, yeah it does. I wanted to give the original PR submitter a chance to do it first, since he already did that work in #27. I'll update this if he doesn't get a chance to.

ross: Checking back in on this. Would be good to get this cleared up and tested.

mvermaes: @michaelmcallister, are you still interested/do you have time to take a look at adding your fix from #27 here?

Please let me know, I'll probably have a look at incorporating your stuff towards the end of the week if you don't get a chance to, so this PR can be closed. I think I can add you as a co-author on those commits so you would still get credit.

michaelmcallister: I'll work on it now @mvermaes  - sorry for the delay!

michaelmcallister: Done see: https://github.com/github/octodns/pull/216

mvermaes: Great, thanks for that!

@ross let us know if you're OK to merge these two PRs as-is, otherwise I don't mind if @michaelmcallister incorporates my trivial edit into #216.

ross:
Thanks @mvermaes & @michaelmcallister

PR #200, opened 2018-02-18
Changes recommended structure of config files

ad-m:
Keeping everything in one directory made it difficult for me to understand what the expected directory structure is. Hence, it proposes to separate the basic configuration of the environment from the detailed configuration of the zones.


README.md
```diff
@@ -79,7 +79,7 @@ Further information can be found in [Records Documentation](/docs/records.md).
 We're ready to do a dry-run with our new setup to see what changes it would make. Since we're pretending here we'll act like there are no existing records for `example.com.` in our accounts on either provider.
 
 ```
-$ octodns-sync --config-file=./config/production.yaml
+$ octodns-sync --config-file=./env/production.yaml
```
ross: `env` would be a common case for how to break these down, but not the only one. E.g. we might separate the public important zones, from the very static and seldom touched vanity redirects like gtihub.com -> github.com, and separate the constantly changing internal zones into yet another config. All of these would be "production", but there'd be (in this case) 3 different setups.

README.md
```diff
@@ -60,7 +60,7 @@ Further information can be found in the `docstring` of each source and provider
 
 Now that we have something to tell OctoDNS about our providers & zones we need to tell it about or records. We'll keep it simple for now and just create a single `A` record at the top-level of the domain.
 
-`config/example.com.yaml`
+`zones/example.com.yaml`
```
ross: The zones change makes sense.

ad-m:
I - at the time of README reading - at some point I got the impression that we need to create a nested ``config/config`` directory. Hence the will to change the name of ``config`` directory. It may be worth explaining it in a different way.

README.md
ad-m: @ross , maybe there is another name that will actually describe the content? `` config`` matches everything. Maybe ```inventory```?

ross: I think I'd like to stick to `config/` here since it's "config" for octoDNS itself. Inventory doesn't really seem to line up to me and I've not come up with anything else that seems to encompass all the cases.

ad-m:
Is there anything what I can do to help merge that PR?

ross:
> Is there anything what I can do to help merge that PR?

As mentioned in the review. The zones change makes sense, but I'd rather than recommend the `env/` change as the config files aren't always tied to the environment (prod/staging/dev.) Sometimes they're split up other ways, e.g. `config/production.yaml` and  `config/secondary.yaml` where secondary has things that essentially never change, other TLDs, type-os, mis-spellings, ...

PR #204, opened 2018-02-19
No Cloudflare Automatic message

ross:
Call out that "automatic" records are not supported and will cause problems for the time being. 

/cc https://github.com/github/octodns/issues/202 & https://github.com/github/octodns/pull/162#issuecomment-366736328
/cc @vanbroup @ad-m


ross:
Closing in favor of https://github.com/github/octodns/pull/210.

PR #205, opened 2018-02-19
Extract _record_for from populate, use round trip in _apply_Update

ross:
This will ensure that we have exactly the same logic/behavior across the
board when turning records into content and prevent the :-( hack that was in
here before. It was missing the max(min, ttl) bit we throw everything else
through and this makes that consistent. Most importantly it'll prevent us
from having to fix bugs or make improvements in multiple code paths in the
future.

@vanbroup this should prevent what you were seeing in https://github.com/github/octodns/pull/162#issuecomment-366767533 b/c it more consistently deals with TTL in the partial update case. 

That said the only way to hit the problem would have been to have a TTL < 120s, which I guess comes from the "Automatic" option which is turned on with proxied records, but can be selected manually in the UI. octoDNS won't support that and will remove/clamp it to 120s the first time it mucks with any of the values for a record.

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -210,6 +210,20 @@ class CloudflareProvider(BaseProvider):
 
         return self._zone_records[zone.name]
 
+    def _record_for(self, zone, name, _type, records, lenient):
+        # rewrite Cloudflare proxied records
+        if self.cdn and records[0]['proxied']:
+            data = self._data_for_cdn(name, _type, records)
+        else:
+            # Cloudflare supports ALIAS semantics with root CNAMEs
+            if _type == 'CNAME' and name == '':
+                _type = 'ALIAS'
+
+            data_for = getattr(self, '_data_for_{}'.format(_type))
+            data = data_for(_type, records)
+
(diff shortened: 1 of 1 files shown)


vanbroup:
I used a slightly different approach by adding support for a TTL of 1 or higher than 120:
https://github.com/github/octodns/pull/206

ross:
> As the TTL=1 is the Cloudflare default for proxied and non-proxied records I think it's key to support it.

I looked around at that for this PR and decided against it b/c I don't like the idea of putting a 1 in the config for Cloudflare provider and it meaning something completely different for other providers (an actual TTL of 1s). 

Since this sort of "automatic" unknown default is specific to CF I landed on it just not being a concept that octoDNS should support b/c it can't do so elsewhere. The line of thinking comes from octoDNS's focus on getting the same behavior out of multiple providers and not just allowing full-featured managed config for each provider.

ross:
Going to ship this. I still plan to circle back and see if there's something reasonable to do with "automatic" records, based on https://github.com/github/octodns/pull/206#issuecomment-367018436 it might be as simple as converting 1s to 300s or just translating them during population.

PR #206, opened 2018-02-19
Allow TTL=1 (auto) in Cloudflare provider

vanbroup:
This solved the deletion/update of non-relevant records and allows to reset the TTL to 1 (the values that Cloudflare uses for automatic TTL selection).

https://github.com/github/octodns/pull/162

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -259,7 +259,7 @@ class CloudflareProvider(BaseProvider):
         if isinstance(change, Update):
             existing = change.existing.data
             new = change.new.data
-            new['ttl'] = max(120, new['ttl'])
+            new['ttl'] = self._min_ttl(new['ttl'])
             if new == existing:
                 return False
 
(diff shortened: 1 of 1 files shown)


ross:
Dropped my thoughts on the 1s == automatic over in https://github.com/github/octodns/pull/205#issuecomment-366812079, but they're probably just as relivant on this PR/thread so I'll crosspost them here:

> I looked around at that for this PR and decided against it b/c I don't like the idea of putting a 1 in the config for Cloudflare provider and it meaning something completely different for other providers (an actual TTL of 1s).

> Since this sort of "automatic" unknown default is specific to CF I landed on it just not being a concept that octoDNS should support b/c it can't do so elsewhere. The line of thinking comes from octoDNS's focus on getting the same behavior out of multiple providers and not just allowing full-featured managed config for each provider.

vanbroup:
Fair enough. Probably some logic would be desirable that for proxied records the TTL forced to 1s or whatever Cloudflare accepts. I will leave it to you as I have no interest in sending updates to Cloudflare anyway.

ross:
> Probably some logic would be desirable that for proxied records the TTL forced to 1s or whatever Cloudflare accepts.

Yeah. I'm still bouncing around ways to handle it a bit better for Cloudflare itself. What it does now isn't horrible.. leave it alone unless it's otherwise changing the record, then clamp it to 120s, but I don't have a clue what the actual "automatic" behavior is atm so it's hard to figure out what's right :-)

vanbroup:
> The automatic TTL value means that your DNS changes with Cloudflare will push out in approximately 5 minutes  (300 seconds).

https://support.cloudflare.com/hc/en-us/articles/200168866-What-does-the-Automatic-TTL-value-mean-

ross:
>> The automatic TTL value means that your DNS changes with Cloudflare will push out in approximately 5 minutes (300 seconds).
 
> https://support.cloudflare.com/hc/en-us/articles/200168866-What-does-the-Automatic-TTL-value-mean-

Heh. OK. That makes it seem like the right thing to do may be `1 => 300`, but would have to look at the details/scenarios to make sure that it keeps things clean.

PR #210, opened 2018-02-24
CloudflareProvider: Map TTL=1 to 300

ross:
This will map the Cloudflare ttl of 1 to 300s which based on [this support thread](https://support.cloudflare.com/hc/en-us/articles/200168866-What-does-the-Automatic-TTL-value-mean-) is the behavior users would be getting there. The result of this will be that automatic ttl things dumped from CF would have 300s ttls in the resulting output which would once synced to CF and other places result in matching/preserving behaviors. It won't change anything that's already 1 in CF unless the record otherwise is being modified.

I think this is our best option for handling the automatic TTLs with the mindset of trying to get multiple providers to behavior the same way from a single config (octoDNS's mindset.)

/cc Fixes https://github.com/github/octodns/issues/202
/cc Replaces https://github.com/github/octodns/pull/204
/cc @vanbroup @ad-m for thoughts

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -102,6 +102,15 @@ class CloudflareProvider(BaseProvider):
 
         return self._zones
 
+    def _ttl_data(self, ttl):
+        # Apparently CF's ttl=1 (automatic) is just an alias for 300 :shrug:
+        # https://support.cloudflare.com/hc/en-us/articles/
+        #   200168866-What-does-the-Automatic-TTL-value-mean-
+        # We'll map it to what it actually does so that dumps and syncs to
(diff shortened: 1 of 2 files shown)


OrRosenblatt:
Why not change 
```
new['ttl'] = max(self.MIN_TTL, new['ttl'])
```
to
```
if not new['ttl'] == 1
  new['ttl'] = max(self.MIN_TTL, new['ttl'])
```
?

It will keep Cloudflare API default behavior.

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-cloudflare/. Feel free to rework/reopen this PR there.

PR #211, opened 2018-02-25
Implement, document, and test --cautious mode

ross:
Normal plan:

```
(env)otter:octodns ross$ PYTHONPATH=. ./octodns/cmds/sync.py --config-file=config/dev.yaml
2018-02-24T16:18:12  [140736612635584] INFO  Manager __init__: config_file=config/dev.yaml
2018-02-24T16:18:12  [140736612635584] INFO  Manager __init__:   max_workers=4
2018-02-24T16:18:12  [140736612635584] INFO  Manager __init__:   include_meta=True
2018-02-24T16:18:12  [140736612635584] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=True, force=False, cautious=False
2018-02-24T16:18:12  [140736612635584] INFO  Manager sync:   zone=githubtest.net.
2018-02-24T16:18:12  [140736612635584] INFO  Manager sync:   sources=['config'] -> targets=['route53']
2018-02-24T16:18:12  [123145418510336] INFO  YamlProvider[config] populate:   found 17 records
2018-02-24T16:18:12  [123145418510336] INFO  Route53Provider[route53] plan: desired=githubtest.net.
2018-02-24T16:18:13  [123145418510336] INFO  Route53Provider[route53] populate:   found 18 records
2018-02-24T16:18:13  [123145418510336] INFO  Route53Provider[route53] plan:   Creates=1, Updates=1, Deletes=1, Existing Records=18
2018-02-24T16:18:13  [140736612635584] INFO  Manager
* githubtest.net.
* route53 (Route53Provider)
*   Update
*     <ARecord A 1, test.githubtest.net., ['1.1.1.1', '1.1.1.2', '1.1.1.4']> ->
*     <ARecord A 600, test.githubtest.net., ['1.1.1.1', '1.1.1.2', '1.1.1.3']> (config)
*   Delete <ARecord A 3600, cx.githubtest.net., ['1.2.3.4']>
*   Create <ARecord A 3600, cx2.githubtest.net., ['1.2.3.4']> (config)
*   Summary: Creates=1, Updates=1, Deletes=1, Existing Records=18
```

Cautious plan

```
(env)otter:octodns ross$ PYTHONPATH=. ./octodns/cmds/sync.py --config-file=config/dev.yaml --cautious
2018-02-24T16:18:29  [140736612635584] INFO  Manager __init__: config_file=config/dev.yaml
2018-02-24T16:18:29  [140736612635584] INFO  Manager __init__:   max_workers=4
2018-02-24T16:18:29  [140736612635584] INFO  Manager __init__:   include_meta=True
2018-02-24T16:18:29  [140736612635584] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=True, force=False, cautious=True
2018-02-24T16:18:29  [140736612635584] INFO  Manager sync:   zone=githubtest.net.
2018-02-24T16:18:29  [140736612635584] INFO  Manager sync:   sources=['config'] -> targets=['route53']
2018-02-24T16:18:29  [123145376927744] INFO  YamlProvider[config] populate:   found 17 records
2018-02-24T16:18:29  [123145376927744] INFO  Route53Provider[route53] plan: desired=githubtest.net.
2018-02-24T16:18:30  [123145376927744] INFO  Route53Provider[route53] populate:   found 18 records
2018-02-24T16:18:30  [123145376927744] INFO  Route53Provider[route53] plan:   Creates=1, Updates=1, Deletes=1, Existing Records=18
2018-02-24T16:18:30  [140736612635584] INFO  Manager
* githubtest.net.
* route53 (Route53Provider)
*   Update
*     <ARecord A 1, test.githubtest.net., ['1.1.1.1', '1.1.1.2', '1.1.1.4']> ->
*     <ARecord A 60, test.githubtest.net., ['1.1.1.1', '1.1.1.2', '1.1.1.3']> (config)
*   Delete <ARecord A 3600, cx.githubtest.net., ['1.2.3.4']>
*   Create <ARecord A 60, cx2.githubtest.net., ['1.2.3.4']> (config)
*   Summary: Creates=1, Updates=1, Deletes=1, Existing Records=18
```

This automates something we've done a couple times now to "cautiously" roll out changes in a way where we have a quick path to revert. We've been doing this with two rounds of changes were we manually set the TTLs on the 😨 bits to 60s in the first round and then to the real value we want in the subsequent. This just builds that concept into octoDNS itself.

It seems useful, but feedback/thoughts solicated.

/cc @theojulienne @joewilliams @yzguy @vanbroup

### DIFF
--- a/octodns/provider/plan.py
+++ b/octodns/provider/plan.py
@@ -50,6 +50,11 @@ class Plan(object):
                        self.change_counts['Update'],
                        self.change_counts['Delete'], existing_n)
 
+    def make_cautious(self):
+        for change in self.changes:
+            if change.new:
+                change.new.ttl = 60
+
     def raise_if_unsafe(self):
         # TODO: what is safe really?
         if self.existing and \
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Now that we have something to tell OctoDNS about our providers & zones we need t
 ```yaml
 ---
 '':
-  ttl: 60
+  ttl: 300
   type: A
   values:
     - 1.2.3.4
@@ -85,10 +85,10 @@ $ octodns-sync --config-file=./config/production.yaml
 * example.com.
 ********************************************************************************
 * route53 (Route53Provider)
-*   Create <ARecord A 60, example.com., [u'1.2.3.4', '1.2.3.5']>
+*   Create <ARecord A 300, example.com., [u'1.2.3.4', '1.2.3.5']>
 *   Summary: Creates=1, Updates=0, Deletes=0, Existing Records=0
 * dyn (DynProvider)
-*   Create <ARecord A 60, example.com., [u'1.2.3.4', '1.2.3.5']>
+*   Create <ARecord A 300, example.com., [u'1.2.3.4', '1.2.3.5']>
 *   Summary: Creates=1, Updates=0, Deletes=0, Existing Records=0
 ********************************************************************************
 ...
@@ -108,6 +108,36 @@ $ octodns-sync --config-file=./config/production.yaml --doit
 ```
 
 The output here would be the same as before with a few more log lines at the end as it makes the actual changes. After which the config in Route53 and Dyn should match what's in the yaml file.
+
+#### Cautious Mode
+
+If changing sensitive records where the repercussions of problems may be substantial octoDNS sync offers a flag `--cautious` that can be added to the sync command to lower the TTL of any modified or added records down to 60s. Deletes are essentially cautious by default since they will become available immediately once you add them back.
+
+While this doesn't help address any problems with the changes themselves what it does is provide a reasonably quick path to a revert. If a change was being made to a record whose TTL is normally 3600, 1 hour, reverting changes to that record will take up to an hour. This is due to the need to wait for caches to expire the problematic value. Using `--cautious` the change would be made with a 60s TTL, so that the problematic value would be flushed quickly. The expected flow using cautious would look something like the following.
+
+
+```
+$ octodns-sync --config-file=./config/production.yaml
+# Review the changes as they would be made normally
+$ octodns-sync --config-file=./config/production.yaml --cautious
+# Review the cautious version of th changes, should be the same, just with 60s TTLs
+$ octodns-sync --config-file=./config/production.yaml --cautious --doit
+# Make the cautious version of the changes, verify they're working as expected,
+# allowing original records to expire and the new to take effect
+$ octodns-sync --config-file=./config/production.yaml --doit
+# Make the normal version of the changes active, should only modify TTLs
+```
+
+If there were problems after the `--cautious --doit` step you can quickly jump back and deploy out the previous version.
+
+```
+$ git checkout master
+# Go back to a known good state
+$ octodns-sync --config-file=./config/production.yaml
+# Make sure the revert is doing what you expect/need
+$ octodns-sync --config-file=./config/production.yaml --doit
+# Apply it, after 60s it should take effect
+```
 
 ### Workflow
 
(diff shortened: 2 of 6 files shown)


yzguy:
I believe this would be a helpful feature for sure. Some thoughts.

I think we can identity at least two workflows for changes

1. Lower TTL, wait time enough for ~all clients to be using new TTL, change value

Workflow 1 I think is common, but maybe is more of a candidate for the two round rollout rather than a "cautious" rollout. It gives you a shorter time to convergence across all clients and quick rollback. However at the same time, a larger percentage of traffic is taking the change in a shorter amount of time, so if it's a bad change then the impact is bigger.

2. Lower TTL, change value, wait time enough for ~all clients to be using new TTL + value.

Workflow 2 is the one that the "cautious" rollout targets, this would be a smaller percentage of traffic moving in a longer time period, however a quicker rollback for the clients that took the change, resulting in lower impact.


One thing that comes to mind that caused a big issue using Terraform that would be similar, is that when you do "cautious", and apply, you have changed the state on the provider, but you haven't changed the state locally.

Two cases I can think of that would become an issue:

1. You have pushed up the changes, someone else goes to make a change, they apply your "normal record", which removes the  "cautious record", effectively increasing the time it takes to rollback should you need to.

2. You don't push up the changes, someone else goes to make a change, and deletes/reverts your record because they don't have your changes, effectively causing an outage in the deletion case and an unknown recovery time as resolver behavior around NXDOMAIN records differs. In the revert case, the results vary, but there's some risk of impact.

I know in the case of Github, you're pushing up changes, there's a noop process on a branch, review, and then a deploy of the branch, then merge. In the case of the cautious, there is a period where the branch does not reflect the true state. If someone else made a branch, did a noop, I guess the "removing cautious" would hopefully be caught during review, before their deploy went out. However, if you're banking on having some time before your record was updated to "normal" to protect your rollback, you effectively have to halt all DNS deploys until you've either rolled back or went forward with bringing it to "normal". That period could be a while, especially if you're waiting for traffic to shift, and you could have a issue that only comes up when you're under load.

I think as far as the feature, it's definitely a great addition to OctoDNS, the above points I believe are process problems, not a tool problem. I'm gonna pull down in the morning and give it a test!

vanbroup:
Thanks @ross, this is a great feature!

I agree with @yzguy that it comes with a workflow challenge. I was thinking that an intermediate branch could push a "cautious" state and the master branch would set the normal state. But this would require another (manual) merge and you don't want the records in a "cautious" state for too long either.

octodns/provider/plan.py
```diff
@@ -50,6 +50,11 @@ def __init__(self, existing, desired, changes,
                        self.change_counts['Update'],
                        self.change_counts['Delete'], existing_n)
 
+    def make_cautious(self):
+        for change in self.changes:
+            if change.new:
+                change.new.ttl = 60
```
joewilliams: My only nit, could the TTL be a global or maybe configurable?

ross:
> Lower TTL, change value, wait time enough for ~all clients to be using new TTL + value.

This seems like  it describes the use-case this was intended for. Changes are rolled out with a short TTL so that if something goes wrong and you revert the reversion will take effect quickly, thus "cautious."

> Lower TTL, wait time enough for ~all clients to be using new TTL, change value

I think I'd call this one aggressive more than cautious. Lowering the TTL before making a change would cause all the clients to pick it up quickly so if there was a problem, everyone would see it before you'd have a chance to revert. You could at least still revert quickly assuming you left the lower TTL in place when the value change happened. 

> Two cases I can think of that would become an issue:

> You have pushed up the changes, someone else goes to make a change, they apply your "normal record", which removes the "cautious record", effectively increasing the time it takes to rollback should you need to.

> ...

Our deployment system ensures that only one branch deploy is active at a time. In general i think there are a lot of problems you could run into without "doit" locking of some sort. Terraform's serial numbers and saved plans style setup could help with that, but that was a direction we consciously didn't end up taking to avoid the need for state of some sort (and the extra complexity.)

PR #214, opened 2018-02-26
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/tests/test_octodns_provider_cloudflare.py
+++ b/tests/test_octodns_provider_cloudflare.py
@@ -599,7 +599,8 @@ class TestCloudflareProvider(TestCase):
         zone = Zone('unit.tests.', [])
         provider.populate(zone)
 
-        # the two A records get merged into one CNAME record poining to the CDN
+        # the two A records get merged into one CNAME record pointing to
+        # the CDN.
         self.assertEquals(3, len(zone.records))
 
         record = list(zone.records)[0]
--- a/tests/test_octodns_provider_cloudflare.py
+++ b/tests/test_octodns_provider_cloudflare.py
@@ -621,7 +621,7 @@ class TestCloudflareProvider(TestCase):
         self.assertEquals('a.unit.tests.cdn.cloudflare.net.', record.value)
 
         # CDN enabled records can't be updated, we don't know the real values
-        # never point a Cloudflare record to itsself.
+        # never point a Cloudflare record to itself.
         wanted = Zone('unit.tests.', [])
         wanted.add_record(Record.new(wanted, 'cname', {
             'ttl': 300,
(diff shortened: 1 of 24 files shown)


tests/test_octodns_provider_dyn.py
```diff
@@ -932,14 +932,14 @@ def test_mod_geo_create(self, mock):
         provider = DynProvider('test', 'cust', 'user', 'pass',
                                traffic_directors_enabled=True)
 
-        # will be tested seperately
+        # will be tested seperatelyly
```
yzguy: `separately` would be the correct spelling for this one.

tests/test_octodns_provider_cloudflare.py
```diff
@@ -599,7 +599,7 @@ def test_cdn(self):
         zone = Zone('unit.tests.', [])
         provider.populate(zone)
 
-        # the two A records get merged into one CNAME record poining to the CDN
+        # the two A records get merged into one CNAME record pointing to the CDN
```
yzguy: `tests/test_octodns_provider_cloudflare.py:602:80: E501 line too long (80 > 79 characters)`
Error causing the test failure

tests/test_octodns_provider_googlecloud.py
```diff
@@ -401,8 +401,8 @@ def test__get_gcloud_zone(self):
         provider.gcloud_client.list_zones = Mock(
             return_value=DummyIterator([]))
 
-        self.assertIsNone(provider.gcloud_zones.get("nonexistant.xone"),
-                          msg="Check that nonexistant zones return None when"
+        self.assertIsNone(provider.gcloud_zones.get("nonexistent.xone"),
```
yzguy: We might want to take this change to make this `nonexistent.zone`, instead of `.xone`

yzguy:
After fixing the issue caught by the linter, the tests all pass. Just a couple comments inline

tests/test_octodns_provider_googlecloud.py
jsoref: Will do

jsoref:
I'll try to get to this sometime this week...

yzguy:
Thanks! Great little tool you have there for fixing up simple spelling mistakes!

tests/test_octodns_provider_dyn.py
jsoref: This was me trying to do a fix by hand instead of letting my browser suggest the spelling fix. Doh.

yzguy:
@ross looks good to me. Do you think it’s worth squashing down to 1 commit for all the changes?

jsoref:
FWIW, I leave them as individual commits while developing because it makes addressing errors, dropping things which people want to keep, and dealing w/ merge conflicts during rebase much easier.

ross:
> @ross looks good to me. Do you think it’s worth squashing down to 1 commit for all the changes?

Yeah. The history will remain here if anyone is interested in it and that'll avoid a load of small commits in `git log` etc.

If you're both happy with the state of it. I'll merge.

jsoref:
Note that I tend to delete my branches, so the individual commits won't last. If you need me not to, I can...

ross:
> Note that I tend to delete my branches, so the individual commits won't last. If you need me not to, I can...

Pretty sure they'll stick around anyway (thanks to behind the scenes GH magic) so it should be fine to delete.

yzguy:
Loooks good to me!

PR #219, opened 2018-03-08
Sorting feature

ydesloges88:
Added sorting function and args.
Adding --sort to cmd line will automatically sort the zone file before sending to validation.

### DIFF
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -2,10 +2,9 @@
 <project version="4">
   <component name="ChangeListManager">
     <list default="true" id="5dba8fff-ec86-4dc6-83f5-ecd72c871d3a" name="Default" comment="">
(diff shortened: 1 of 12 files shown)


yzguy:
@ydesloges88 I see in some of the commits there is an UltraDNS provider? Looks to be only in MindGeek's Fork of OctoDNS.

Is that something you/MindGeek would contribute back to upstream? If not, I would not mind taking what is there and finishing it, writing tests, etc.

asvinours:
@yzguy I just pushed more stuff in our UltraDNS provider (SiteBacker management) but I haven't started writing tests.
We have it in our plans to write all the tests and then ask for our features to be merged but I can't tell you when it will happen. If you want to help and write tests for it, it would be amazing.

We also did not add all the ultraDNS feature to our ultradns provider. We only added features we use (Sitebacker, Directionnal dns and most of the records types).

PR #222, opened 2018-03-24
Make sure to insert new rulesets at the "end"

ross:
Dyn will still match them, even while they're empty before we've had a
chance to add the respons pools to them which is BAD and caused a
medium-size outage (thankfully not as bad as it could have been.) Ideally
we'd use publish=False and stage things, but that seems broken in the client
lib, there's no way to send publish=N. I also tried sending the
response_pool_ids as part of the create calls and response pool config if
one didn't exist, but neither of those routes worked :-(

### DIFF
--- a/octodns/provider/dyn.py
+++ b/octodns/provider/dyn.py
@@ -547,10 +547,22 @@ class DynProvider(BaseProvider):
 
         # Rulesets
 
+        # We need to make sure and insert the new rules after any existing
+        # rules so they won't take effect before we've had a chance to add
+        # response pools to them. I've tried both publish=False (which is
+        # completely broken in the client) and creating the rulesets with
+        # response_pool_ids neither of which appear to work from the client
+        # library. If there are no existing rulesets fallback to 0
+        insert_at = max([
+            int(r._ordering)
+            for r in existing_rulesets
+        ] + [-1]) + 1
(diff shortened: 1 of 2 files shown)


ross:
One potential problem with this would be if there's a (especially low) limit on index/ordering. The max could hit it and we'd lose the ability to insert where we want. I don't have a clue how we'd fix that without doing an "unsafe" ruleset insert to lower the values. Going to make sure that it's not super low (byte, char, etc.) and if not just 🏈 on trying to solve it.

ross:
Starting at `2147483647` works fine and running a few times after that still works.  Starting at `4294967295` failed so it seems like that field is probably an unsigned int. Definitely not going to deal with that and I'll give the first person who hits it without cheating a $1.

ross:
> Less critical and definitely fine for a follow up PR, but it might make sense to make sure we delete the old rulesets starting with the old default, so we never have a situation where there's a default with no geo records above it, even for a split second.

It was a simple one-line fix to take care of this case. Basically removes the old rulessets in reverse _ordering. That ensures that the old default is removed first. Seems like that should be optimal in general as we'll be removing them from least to most specific. That should let things fall through as the old rules go away into the new rules rather than potentially matching the wider nets from the bottom of the old.

PR #236, opened 2018-04-24
Fix minor typo in contributing doc, layout

mvermaes:


yzguy:
Looks good! Thank you :)

CONTRIBUTING.md
```diff
@@ -40,8 +40,10 @@ Here are a few things you can do that will increase the likelihood of your pull
 
 ## Development setup
 
-pip instal -r requirements.txt
-pip instal -r requirements-dev.txt
+```
+$ pip install -r requirements.txt
+$ pip install -r requirements-dev.txt
```
ross: The code/pre block is a good idea here, but if you don't mind remove the $ so that they can be copy-n-pasted without editing. Otherwise LGTM.

mvermaes: Makes sense, updated.

ross:
Thanks.

PR #249, opened 2018-06-20
Support DS record delegation for subzone

nunnun:
Currently delegating a NS record to subzone is allowed.
But I think DS record is also needed for subzone delegation.
So I've changed the code to allow a DS record as well as a NS record.


ross:
It looks like DS records are part of the DNSSEC bits of DNS, https://www.godaddy.com/help/what-is-a-ds-record-6148. There's a bit of discussion and background on why octoDNS doesn't support DNSSEC over in https://github.com/github/octodns/issues/176. It'd be nice if it octoDNS had support, but a quick look though things at that point seemed to indicate that provider support wasn't really there/ready for it.

yonosemanana:
Hello,

I see three years passed since last comments here in 2018. AWS Route53, Google Cloud DNS and other DNS providers, like Akamai, CloudFlare, Neustar, Dyn, NS1, etc. support DNSSEC.

Do you have plans to add DNSSEC support to OctoDNS? Any ETA, if you have the plans?

ross:
> Do you have plans to add DNSSEC support to OctoDNS? Any ETA, if you have the plans?

I don't have any specific use cases where DNSEC adds something meaningful and in consulting with security people (who know a lot more about the space than I) in the past the response was generally no real reason to use it. That's not to say no, just more than I'm not motivated to figure it out and if/how it'd work with [split authority](https://github.blog/2017-04-27-enabling-split-authority-dns-with-octodns/) which would be important to adding support in octoDNS given that is it's focus. 

If someone else takes that on, gets everything in order & implemented I'd be happy to review, but it's not something I'm likely to dig into short of taking on a contract for it or something.

yonosemanana:
Hi Ross,
Thank you for the quick response!
I personally agree with you that DNSSEC doesn't bring much value. But sometimes we have to enable it because of regulations / compliance. :-)  
Noted that DNSSEC is not a priority for now, thanks!

ross:
> But sometimes we have to enable it because of regulations / compliance. :-)

Yup. I have worked at places in the past that faced the same external requirement. Afaik we pushed back/filed variances or whatever that's called. We were likely aided by the fact that all of the exposed services (IPs & ports) were TLS protected and thus had their identity validated by certificates. 

Just to get it in the discussion/record are you all using multiple providers or just a single provider?

yonosemanana:
Looking to use many DNS providers. If possible, it would be cool to have DNSSEC signing under control and then push signed records to many providers.

Issue #239, opened 2018-05-03
AXFR Source


yzguy:
So I was thinking about this today, it would be useful to be able to bootstrap your zone file from any provider in as generic of a way as possible, ie AXFR

In this case I'm thinking maybe like a Bind DNS server, etc. Something that does not have an API to call to get records.

In this way, you could get your initial zone file in OctoDNS, then that would allow you to then switch which provider you push it to, or even allow you to do a sync from Bind out to an external provider.

I know many providers allow you to import a bind compatible zone file to start your zone, so that would be one way around it, but I'm curious if there is a use case/advantage to doing it this way through OctoDNS

I'm not talking about a Zone File Provider to template out a Bind Compatible zone file though. However, maybe you could also parse a zone file into OctoDNS zone file as an alternative to a AXFR

Thoughts?

It would be interesting as you couldn't do a YAML -> the provider, but you could do Provider -> YAML

ross:
> ... It would be interesting as you couldn't do a YAML -> the provider, but you could do Provider -> YAML

That's the idea behind `Source`s. They're able to populate zones, but not plan and apply since they're one-way.

I definitely like the idea of a AXFR `Source` though I haven't looked into the details of what it'd take to do it. I could have sworn this came up somewhere before, but there's nothing about it in the repo history so it must have been in person with someone or in chat. 

There's an existing [`Source` for TinyDNS](https://github.com/github/octodns/blob/master/octodns/source/tinydns.py) which can read it's config file and convert it to octoDNS's format. We used that as part of the internal DNS infrastructure migration. I'd imagine a similar setup for AXFR wouldn't be TOO bad to pull off.

yzguy:
Ha, I should read more 

> An AxfrSource could be really interesting as well. Another case where a source may make sense is if you'd like to export data from a legacy service that you have no plans to push changes back into.

So internally I've written a python script that will zone transfer two providers and compare the zone results to each other and tell you what is missing, what is not matching, etc

That is using the dnspython package, which after doing the zone transfer you get basically a hash of hashes that represent all your records. That data structure would be very similar to what you get back from any provider we are already getting data from.

https://gist.github.com/yzguy/e648209d332520956ecabe6e64db24cb

ross:
`octodns-compare` does similar comparison type stuff, though it works at the provider level, `octodns-report` does actual DNS queries against servers using octoDNS to get the list of things to check. 

```
[...]

PR #253, opened 2018-07-09
add AXFR source to OctoDNS

yzguy:
Adds a new source requested in #239. This source allows a user to pull data from
a legacy system (Bind9, etc.) that does not have an API/existing provider via
AXFR Zone Transfer.

I tested it against NS1 via AXFR with a zone with ~6500 records and it worked like a dream, might be missing some record types, but it's a good start. Figured I'd get it up here for some review while I tweak/add tests.

Ideally, it could be highly configurable:
* TSIG
* Different Port
* Timeout


README.md
```diff
@@ -162,7 +162,8 @@ The above command pulled the existing data out of Route53 and placed the results
 | [PowerDnsProvider](/octodns/provider/powerdns.py) | | All | No | |
 | [Rackspace](/octodns/provider/rackspace.py) | | A, AAAA, ALIAS, CNAME, MX, NS, PTR, SPF, TXT | No |  |
 | [Route53](/octodns/provider/route53.py) | boto3 | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, TXT | Yes | |
-| [TinyDNSSource](/octodns/source/tinydns.py) | | A, CNAME, MX, NS, PTR | No | read-only |
+| [TinyDnsFileSource](/octodns/source/tinydns.py) | | A, CNAME, MX, NS, PTR | No | read-only |
+| [AxfrSource](/octodns/source/axfr.py) | | A, AAAA, CNAME, MX, NS, PTR, SPF, SRV, TXT | No | read-only |
```
ross: The rest of these are in alphabetical order. Probably good to move this up to the top. We could break things out into two tables, providers and sources, if you'd prefer to group them.

octodns/source/axfr.py
```diff
@@ -0,0 +1,157 @@
...
+                                                   relativize=False),
+                                     relativize=False)
+        except DNSException:
+            raise AxfrSourceZoneTransferFailed()
+
+        records = []
+
+        for (name, ttl, rdata) in zone.iterate_rdatas():
+            rdtype = dns.rdatatype.to_text(rdata.rdtype)
+            if rdtype != 'SOA':
```
ross: Should this use `SUPPORTS` to filter things to only the supported record types?

yzguy: I was thinking about this today, I think I would end up removing it here, and adding it within the populate method.

Ideally this `zone_records` method is just the thing that performs the zone transfer and transforms the response into a more sane data structure, it shouldn't care about filtering records by type.

However `populate` should care about what record types get let through.

I will make the changes.

ross:
Just checking that you're ready for this to :ship: before I merge...

yzguy:
Need to write the tests for it, it'll be a little confusing because I have to probably mock the response from the zone transfer, which is a bunch of dnspython objects, etc.

I'm not sure if this is the best spot to put it, but I added a ZoneFileSource as well, as the only difference with loading from a zone file or a zone transfer with dnspython is a method, so it was a quick win to add a source for a zone file.

It's a little confusing it being under the axfr module, but I'm not sure what would be the best place to put it. I'm open to any suggestions on better placement/organization.

yzguy:
Added test, instead of creating a bunch of dnspython objects, I created a small zone file, then load it via dnspython's `from_file` method, and then just return that when `from_xfr` is called. Both methods generate a dns.zone.Zone object.

PR #264, opened 2018-09-01
Add ability to manage Cloudflare proxy flag

### LINKED ISSUE #65: Cloudflare and "shield" records
Cloudflare lets you "shield" (cdn/dns) A and CNAME records. This enables their cdn functionality. Seeing how the feature is embedded into the way DNS records is handled on CloudFlare, I thought I should at least bring up whether or not this belongs in octodns.

(if nothing else, a closed issue might be useful if others are thinking of the same thing)

begincalendar:
# Description

Fixes #65, i.e. adds the ability to enable/disable Cloudflare's proxy/CDN flag for **A**, **AAAA** and **CNAME** records.

# TODO

* [x] Support **CNAME** records.
* [x] Support **A** records.
* [x] Support **AAAA** records.
* [x] Add/Alter tests
* [x] Final clean-up and/or refactor
* [x] Update documentation.
* [x] Wait for response from Cloudflare on a relevant support ticket
* [x] Wait for resolution to #269

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -47,6 +47,16 @@ class CloudflareProvider(BaseProvider):
         #
         # See: https://support.cloudflare.com/hc/en-us/articles/115000830351
         cdn: false
+
+    Note: The "proxied" flag of "A", "AAAA" and "CNAME" records can be managed
+          via the YAML provider like so:
+              name:
+                octodons:
+                  cloudflare:
+                    proxied: true
+                ttl: 120
+                type: A
+                value: 1.2.3.4
     '''
     SUPPORTS_GEO = False
     SUPPORTS = set(('ALIAS', 'A', 'AAAA', 'CAA', 'CNAME', 'MX', 'NS', 'SRV',
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
 from collections import defaultdict
+from copy import deepcopy
 from logging import getLogger
 from requests import Session
 
@@ -25,6 +26,9 @@ class CloudflareError(Exception):
 class CloudflareAuthenticationError(CloudflareError):
     def __init__(self, data):
         CloudflareError.__init__(self, data)
+
+
+_PROXIABLE_RECORD_TYPES = {'A', 'AAAA', 'CNAME'}
 
 
 class CloudflareProvider(BaseProvider):
@@ -221,7 +225,14 @@ class CloudflareProvider(BaseProvider):
             data_for = getattr(self, '_data_for_{}'.format(_type))
             data = data_for(_type, records)
 
-        return Record.new(zone, name, data, source=self, lenient=lenient)
+        record = Record.new(zone, name, data, source=self, lenient=lenient)
+
+        if _type in _PROXIABLE_RECORD_TYPES:
+            record._octodns['cloudflare'] = {
+                'proxied': records[0].get('proxied', False)
+            }
+
+        return record
 
     def populate(self, zone, target=False, lenient=False):
         self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
@@ -260,8 +271,18 @@ class CloudflareProvider(BaseProvider):
 
     def _include_change(self, change):
         if isinstance(change, Update):
-            existing = change.existing.data
             new = change.new.data
+
+            # Cloudflare manages TTL of proxied records, so we should exclude
+            # TTL from the comparison (to prevent false-positives).
+            if self._record_is_proxied(change.existing):
+                existing = deepcopy(change.existing.data)
+                existing.update({
+                    'ttl': new['ttl']
+                })
+            else:
+                existing = change.existing.data
+
             new['ttl'] = max(self.MIN_TTL, new['ttl'])
             if new == existing:
                 return False
@@ -322,6 +343,12 @@ class CloudflareProvider(BaseProvider):
                 }
             }
 
+    def _record_is_proxied(self, record):
+        return (
+            not self.cdn and
+            record._octodns.get('cloudflare', {}).get('proxied', False)
+        )
+
     def _gen_data(self, record):
         name = record.fqdn[:-1]
         _type = record._type
@@ -338,6 +365,12 @@ class CloudflareProvider(BaseProvider):
                 'type': _type,
                 'ttl': ttl,
             })
+
+            if _type in _PROXIABLE_RECORD_TYPES:
+                content.update({
+                    'proxied': self._record_is_proxied(record)
+                })
+
             yield content
 
     def _gen_key(self, data):
@@ -512,3 +545,23 @@ class CloudflareProvider(BaseProvider):
 
         # clear the cache
         self._zone_records.pop(name, None)
+
+    def _extra_changes(self, existing, desired, changes):
+        extra_changes = []
+
+        existing_records = {r: r for r in existing.records}
+        changed_records = {c.record for c in changes}
+
+        for desired_record in desired.records:
+            if desired_record not in existing.records:  # Will be created
+                continue
+            elif desired_record in changed_records:  # Already being updated
+                continue
+
+            existing_record = existing_records[desired_record]
+
+            if (self._record_is_proxied(existing_record) !=
+                    self._record_is_proxied(desired_record)):
+                extra_changes.append(Update(existing_record, desired_record))
+
+        return extra_changes
--- a/tests/test_octodns_provider_cloudflare.py
+++ b/tests/test_octodns_provider_cloudflare.py
@@ -18,6 +18,17 @@ from octodns.provider.yaml import YamlProvider
 from octodns.zone import Zone
 
 
+def set_record_proxied_flag(record, proxied):
+    try:
+        record._octodns['cloudflare']['proxied'] = proxied
+    except KeyError:
+        record._octodns['cloudflare'] = {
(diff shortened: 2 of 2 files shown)


begincalendar:
I completed **CNAME** first because Cloudflare allows **A** and **AAAA** records with the same name (e.g. "www") but different values (e.g. "1.2.3.4" and "1.2.3.5"), to have independent proxy/CDN flag settings and that will involve more invasive changes to the Cloudflare provider class.

@ross I know this is not ready to be merged in yet, but am I on the right track?

begincalendar:
@ross 

Perhaps the YAML config should be as follows for **A** and **AAAA** records:
```yaml
foo:
  octodns:
    cloudflare:
      proxied:
        - 1.2.3.5
  type: A
  values:
    - 1.2.3.4
    - 1.2.3.5
```
...where "proxied" is a subset of the "values" that are to have the proxy flag enabled.

A list over a map (IP address keys and proxy flag values), to reduce duplication/verbosity (in the cases where values are not in "proxied" or "proxied" does not exist at all).

For **CNAME** records, a simple boolean could be used (as you suggested).

How does that sound?

ross:
> I completed CNAME first because Cloudflare allows A and AAAA records with the same name (e.g. "www") but different values (e.g. "1.2.3.4" and "1.2.3.5"), to have independent proxy/CDN flag settings and that will involve more invasive changes to the Cloudflare provider class.

Is that something people make use of? Having a hard time imagining a use-case for it.

> A list over a map (IP address keys and proxy flag values), to reduce duplication/verbosity (in the cases where values are not in "proxied" or "proxied" does not exist at all).

I can't think of anything cleaner/simpler if individual values can have it enabled/disabled. 

> For CNAME records, a simple boolean could be used (as you suggested).

I'd maybe stick with the value even for CNAMEs just for consistency sake, but don't feel strongly.

begincalendar:
>> I completed CNAME first because Cloudflare allows A and AAAA records with the same name (e.g. "www") but different values (e.g. "1.2.3.4" and "1.2.3.5"), to have independent proxy/CDN flag settings and that will involve more invasive changes to the Cloudflare provider class.
>
> Is that something people make use of? Having a hard time imagining a use-case for it.

I've never used it myself and agree that there are no obvious use-cases, but Cloudflare provides the ability and if we can implement it cleanly, I don't see why we shouldn't.
I'll see what I can come up with and if it gets out of control, I guess we can revert to a boolean that covers all **A** and **AAAA** values.

>> For CNAME records, a simple boolean could be used (as you suggested).
>
> I'd maybe stick with the value even for CNAMEs just for consistency sake, but don't feel strongly.

In a list, a single value, or allow for both?

I.e. This:
```yaml
foo:
  octodns:
    cloudflare:
      proxied:
        - some.otherdomain.com
  type: CNAME
  value: some.otherdomain.com
```

...or this:

```yaml
foo:
  octodns:
    cloudflare:
      proxied: some.otherdomain.com
  type: CNAME
  value: some.otherdomain.com
```

...or either?

begincalendar:
On second thought...

The existing Cloudflare provider doesn't allow for **A** and **AAAA** records with the same name but different values, to have independent TTLs (even though, Cloudflare allows it).

I'll keep things simple and similar to the TTL handling, by just having a boolean "proxied" YAML config flag for **A**, **AAAA** and **CNAME**, where **A** and **AAAA** "proxied* applies to all values.

parkr:
I use Cloudflare because it's free and it's already setup. 😄 I would _definitely_ love to manage this proxy flag. Anything I can do here to help?

begincalendar:
@parkr Nothing from me at this stage. 

Please see #269 (if you haven't already).

This could be merged prior to addressing that, but I figured it would be prudent to resolve the issue first.

octodns/provider/cloudflare.py
```diff
@@ -44,6 +48,16 @@ class CloudflareProvider(BaseProvider):
         #
         # See: https://support.cloudflare.com/hc/en-us/articles/115000830351
         cdn: false
+
+    Note: The "proxied" flag of "A", "AAAA" and "CNAME" records can be managed
+          via the YAML provider like so:
+              subdomain:
```
ross: I think `name` would be a little more consistent with the octoDNS terminology elsewhere.

octodns/provider/cloudflare.py
```diff
@@ -222,7 +236,14 @@ def _record_for(self, zone, name, _type, records, lenient):
             data_for = getattr(self, '_data_for_{}'.format(_type))
             data = data_for(_type, records)
 
-        return Record.new(zone, name, data, source=self, lenient=lenient)
+        record = Record.new(zone, name, data, source=self, lenient=lenient)
+
+        if records[0].get('proxiable', False):
```
ross: What's the `proxiable` flag here? Would this run into problems if Cloudflare added a new type that had this true, but `_PROXIABLE_RECORD_TYPES` didn't include it?

ross:
> * [ ]  Wait for resolution to #269

Proposed fix for this over in https://github.com/github/octodns/pull/277.

begincalendar:
> Does this make things any worse regarding #269 than they were already?

it depends on how #277 ends up I guess.

> I think resolving that issue will be a bit more involved

I agree.

> ...so it doesn't make sense to block this unless we need to.

It wasn't that I thought one would be more complicated than the other.
It was that I thought there was the potential for the solution to #269 to require changes here.

Yet it looks like that might not be the case, so we should be clear to merge this first.

ross:
Looks like this has minor conflicts with the changes made in https://github.com/github/octodns/pull/277. LMK if you want me to try and resolve them.

begincalendar:
Should be ready for review again.

ross:
Sorry, lost this is the shuffle of moving and other stuff that came up last week. Thanks!

begincalendar:
@ross Can we expect a new release (e.g. to PyPI) that includes the recent changes?
Or are releases done according to some internal schedule?

ross:
> @ross Can we expect a new release (e.g. to PyPI) that includes the recent changes?
> Or are releases done according to some internal schedule?

There's no set schedule. I just do them periodically or when someone asks. I added a TODO item for myself to cut one. Probably be a few days until I make it there :grin:

ross:
0.9.3 released.

parkr:
Is there a way the metadata can be printed? 

```text
2018-10-30T23:08:48  [4642821568] INFO  YamlProvider[config] populate:   found 10 records, exists=False
2018-10-30T23:08:48  [4642821568] INFO  CloudflareProvider[cloudflare] plan: desired=parkermoore.de.
2018-10-30T23:08:48  [4642821568] INFO  CloudflareProvider[cloudflare] populate:   found 10 records, exists=True
2018-10-30T23:08:48  [4642821568] INFO  CloudflareProvider[cloudflare] plan:   extra changes
  Update
    <CnameRecord CNAME 1, email.dev.parkermoore.de., mailgun.org.> ->
    <CnameRecord CNAME 1, email.dev.parkermoore.de., mailgun.org.> (config)
  Update
    <CnameRecord CNAME 1, blog.parkermoore.de., parkr.github.io.> ->
    <CnameRecord CNAME 1, blog.parkermoore.de., parkr.github.io.> (config)
  Update
    <CnameRecord CNAME 1, jekyll.parkermoore.de., parkermoore.de.> ->
    <CnameRecord CNAME 1, jekyll.parkermoore.de., parkermoore.de.> (config)
  Update
    <CnameRecord CNAME 1, bookmarks.parkermoore.de., parkr.github.io.> ->
    <CnameRecord CNAME 1, bookmarks.parkermoore.de., parkr.github.io.> (config)
  Update
    <CnameRecord CNAME 1, jekyllrb.parkermoore.de., parkermoore.de.> ->
    <CnameRecord CNAME 1, jekyllrb.parkermoore.de., parkermoore.de.> (config)
2018-10-30T23:08:48  [4642821568] INFO  CloudflareProvider[cloudflare] plan:   Creates=0, Updates=5, Deletes=0, Existing Records=10
```

begincalendar:
Simplest way I can see is to modify the `_ValuesMixin.__repr__()` and `_ValueMixin.__repr__()` methods to include the `_octodns` attribute, but not sure if that is desirable.

Curious to see what @ross thinks.

ross:
> Curious to see what @ross thinks.

No opposed to the idea so long as the output is reasonably clean & sensible.

begincalendar:
Sorry for the confusion @ross, but I meant I was curious about what you thought about addressing @parkr's request (not my proposed solution).

I thought about the request too during the implementation of this PR, but didn't pursue it because I couldn't see a way to make it work based on the internal APIs.

I can't see how my trivial proposed solution would produce clean/sensible output, so I'm not sure what you have in mind.

ross:
> Sorry for the confusion @ross, but I meant I was curious about what you thought about addressing @parkr's request (not my proposed solution).

Your proposed solution was the first thing that came to mind for me as well. I can't really think of anywhere else off-hand that it'd fit in. That said it's pretty hard to read long records, if you've ever messed with `geo` values you can quickly see that the before and after are tough to read.

If we wanted to do something less verbose we could "checksum" the `_octodns` attribute and just show that. It'd let you see that's what changed, but not how it changed. I think it'd probably be preferable to just output the full text.

> I can't see how my trivial proposed solution would produce clean/sensible output, so I'm not sure what you have in mind.

Yeah, it wouldn't be all that great, but it would match up with how the more involved `geo` records currently work and is the only quick-fix I can think of. 

Bigger picture options would be to rework the [PlanLogger (and friends)](https://github.com/github/octodns/blob/6ec5bd4da2218c28359276ba68c28e02296a306c/octodns/provider/plan.py#L91-L145) to do a more intelligent/involved diff output that could show things with more information.

PR #274, opened 2018-10-06
[route53] Set custom endpoint for route-53 compatibile provider

ad-m:
I work for a [HyperOne](http://www.hyperone.com/) that provides DNS-as-a-service. We plan to publish our Route53-compatible API soon. We want to enable our users to use OctoDNS. 

This change allows the use of any vendor in OctoDNS, which provides API compatible with Route53.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -221,6 +221,8 @@ class Route53Provider(BaseProvider):
         access_key_id:
         # The AWS secret access key
         secret_access_key:
+        # URL of endpoint of AWS-compatible API (optional)
+        endpoint_url: "http://localhost:3000",
 
     Alternatively, you may leave out access_key_id and secret_access_key,
     this will result in boto3 deciding authentication dynamically.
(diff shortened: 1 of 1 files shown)


ross:
Nice. Do you all support the same GeoDNS setup as Route53? If not this will probably need to add a param that makes that optional. 

Also note that I probably have a pretty major reworking of octoDNS's Geo support coming at some point in the not too distant future to allow it to be an order of magnitude more powerful/flexible. That's assuming I get the time to focus on it.

ad-m:
@ross Currently, we do not have any plans to support GeoDNS at all. Our service is simpler. Therefore, we ignore any parameters related to GeoDNS.

ross:
> @ross Currently, we do not have any plans to support GeoDNS at all.

:cool: I believe that may result in problems with an inability to get zones with `geo` records to think they're synced with this PR as-is. 

One option here would be to create a base class that does the simple non-geo stuff and two child classes, Route53Provider which would extend things to support geo and HyperOneProvider which would just inherit and not extend.

Other option would be to make SUPPORTS_GEO driven off of a parameter.

Issue #254, opened 2018-07-15
Feature Request: ZoneFileProvider


yzguy:
Add a ZoneFileProvider, this will be very similar to AxfrSource, but have the ability to read a Bind Compatible zone file and be able to write one out.

dnspython package has methods to read/write zone files, it shouldn't be too difficult.

yzguy:
Wondering if we will need to allow specifying the SOA/NS records, retry, expiry, etc. These will be necessary to create a zone file. 

Alternatively, the zone file may just be a source rather than a provider, and would be trivial to add as it’s a line change from AxfrSource.

ross:
> Alternatively, the zone file may just be a source rather than a provider, and would be trivial to add as it’s a line change from AxfrSource.

Interesting question. I guess if people wanted to run bind off of it it might be nice to allow it to be full provider, but I'd be fine if it's just a source (to start.) If nothing else the SOA/root NS stuff could be specified as part of the provider config. 

There was a stab at managing root NS records a long time ago with https://github.com/github/octodns/pull/53, but it kind of stalled out.

doddo:
I wrote a bind zone file provider and found that templating a bind zone file is much easier than parsing a zone file, because there are so many corner cases and such things.  But the octodns and bind server would have to share the file system (for validation),  so therefore I concluded that for octodns, in my opinion, reading the records from a zone transfer, and writing the data witn DNS Dynamic update would be best approach.  Maybe that's something to consider?

barnumbirr:
Hello,

I've seen that the `Zonefile source` has been added in #276 but was wondering if anyone had started to work on a `Zonefile provider` as I'd be interested in managing Bind zones with octodns.

Cheers.

kompetenzbolzen:
https://github.com/kompetenzbolzen/octodns-custom-provider/blob/master/octodns-custom-providers/provider/zonefile.py

I implemented this for me, would that be something that could be integrated here?

yzguy:
@kompetenzbolzen yeah absolutely, I was actually playing around with your code the other night to see how feasible it would be to integrate, I was having decent success, but then some stuff came up and I haven't picked it back up.

https://github.com/yzguy/octodns/tree/zonefileprovider

kompetenzbolzen:
@yzguy Oh cool! I fixed up some of my code this morning, hence the comment now. I added a config option for the SOA, that works for me. It's a bit of a hackjob and needs a newer version of dnspython though.

[...]

PR #276, opened 2018-10-14
add Zone File source, reads Bind compatible zone files

yzguy:
Addresses half of #254

The dnspython package is a bit confusing on creating a zone file, and I don't really have the use case to do it anymore. I believe a ZoneFileSource is still useful as a way for people to move their data into OctoDNS

As most of the code is similar/same as AxfrSource, I have included it in the same file, but I'm not sure if it is a bit of a anti-pattern.


ross:
> As most of the code is similar/same as AxfrSource, I have included it in the same file, but I'm not sure if it is a bit of a anti-pattern.

It seems reasonable/OK to do so since they're so similar. We can always break things up if/when it gets more involved.

PR #280, opened 2018-10-25
AzureProvider: Absorb CloudError exceptions

koendc:
CloudError exceptions are about not supported values. These exceptions
are now caught and a warning message is produced.
The apply phase will now continue with other changes instead of bailing
out with an exception.


ross:
Sorry, I somehow missed this one coming through and only saw it when I was looking through the backlog.

I believe the `CloudError`s being raised in this situation make sense. octoDNS generally throws errors if there are problems when things fail to apply. If those problems are due to bad data being configured in the zone file (that the provider doesn't accept) then we should look at handling that earlier on in the process, likely during planning.

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-azure/. Feel free to rework/reopen this PR there.

Issue #284, opened 2018-11-01
octodns-dump does not print cloud flare proxy flag


parkr:
Hey! 

This is a follow-up to https://github.com/github/octodns/pull/264 by @begincalendar.

My dns records can be found at [parkr/dns](https://github.com/parkr/dns). In that repo, I have a script called `script/dump` which I use to dump new zones when I create them. I upgraded to octodns 0.9.3 and re-ran `script/dump` hoping to see the new `proxied` field in my config files. Alas, my YAML output was unchanged.

Is it possible to add this provider-specific flag to octodns-dump? 

Here's my incantation:

```bash
for zone in "${ZONES[@]}"; do
    echo "Processing zone $zone."
    ./vendor/bin/octodns-dump \
      --config-file config/production.yaml \
      --output-dir config \
      "$zone." cloudflare \
      "$@"
done
```

(Also, I tried doing `octodns-dump --version`, and it would have no part of it. Think we could add a global --version flag to make checking versions easier?)

begincalendar:
@ross Would something like this cause any issues:
```
diff --git octodns/provider/yaml.py octodns/provider/yaml.py
index 287fd3b..176cf42 100644
--- octodns/provider/yaml.py
+++ octodns/provider/yaml.py
@@ -86,6 +86,8 @@ class YamlProvider(BaseProvider):
         for record in records:
             d = record.data
             d['type'] = record._type
+            if record._octodns:
+                d['octodns'] = record._octodns
             if record.ttl == self.default_ttl:
                 # ttl is the default, we don't need to store it
                 del d['ttl']
```
?

ross:
> @ross Would something like this cause any issues:

@begincalendar yeah that looks like the right approach. Feel free to PR something down that path.

PR #293, opened 2018-11-03
WIP: Add (non-empty) Record._octodns dict to YAML file output

begincalendar:
## Before

```python
# record: Record

record._octodns['cloudflare'] = {
    'proxied': True
}
```

-- (via the YAML provider) -->

```yaml
name:
  ttl: 300
  type: A
  value: 1.2.3.4
```

## After

```python
# record: Record

record._octodns['cloudflare'] = {
    'proxied': True
}
```

-- (via the YAML provider) -->

```yaml
name:
  octodns:
    cloudflare:
      proxied: true
  ttl: 300
  type: A
  value: 1.2.3.4
```

//cc #284 
//cc @parkr

### DIFF
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -89,6 +89,8 @@ class YamlProvider(BaseProvider):
             if record.ttl == self.default_ttl:
                 # ttl is the default, we don't need to store it
                 del d['ttl']
+            if record._octodns:
+                d['octodns'] = record._octodns
             data[record.name].append(d)
 
         # Flatten single element lists


begincalendar:
`YamlProvider._apply()` doesn't have dependency injection hooks for unit testing, but is it worth doing some monkey patching?

ross:
> `YamlProvider._apply()` doesn't have dependency injection hooks for unit testing, but is it worth doing some monkey patching?

You should be able to make use of some `octodns` bits already in the `YamlProvider` [test config](https://github.com/github/octodns/blob/88c3effe9957405010c4d1360316adfaf7354608/tests/config/unit.tests.yaml#L70-L75)

```yaml
included:
  octodns:
    included:
    - test
  type: CNAME
  value: unit.tests.
```

That'll be loaded up and accessible in the `reloaded` zone around [this point](https://github.com/github/octodns/blob/88c3effe9957405010c4d1360316adfaf7354608/tests/test_octodns_provider_yaml.py#L64)

ross:
> Is that what you were alluding to?

Yup.

parkr:
I love this ❤️ Thank you for merging!

PR #298, opened 2018-11-12
Added option to set the AWS session token to the Route53Provider

zBart:
Besides `access_key_id` and `secret_access_key` sometimes AWS also needs a `session_token` to have a complete IAM session.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -221,7 +221,8 @@ class Route53Provider(BaseProvider):
         access_key_id:
         # The AWS secret access key
         secret_access_key:
-        # The AWS session token
+        # The AWS session token (optional)
+        # Only needed if using temporary security credentials
         session_token:
 
     Alternatively, you may leave out access_key_id, secret_access_key
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -246,8 +246,7 @@ class Route53Provider(BaseProvider):
         _msg = 'access_key_id={}, secret_access_key=***, ' \
                'session_token=***'.format(access_key_id)
         use_fallback_auth = access_key_id is None and \
-                            secret_access_key is None and \
-                            session_token is None
+            secret_access_key is None and session_token is None
         if use_fallback_auth:
             _msg = 'auth=fallback'
         self.log = logging.getLogger('Route53Provider[{}]'.format(id))
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -240,8 +240,8 @@ class Route53Provider(BaseProvider):
     HEALTH_CHECK_VERSION = '0001'
 
     def __init__(self, id, access_key_id=None, secret_access_key=None,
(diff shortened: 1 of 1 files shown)


ross:
> Besides `access_key_id` and `secret_access_key` sometimes AWS also needs a `session_token` to have a complete IAM session.

Not opposed to the idea, but session tokens seem ephemeral. What's the workflow that would be used to get one set into the environment? I guess more generally what's the use case that's being solved here? Assume role?

zBart:
When using temporary security credentials you get the session token (besides the access key and secret).

Currently we have two cases where I know of this:
* If you're using [AWS SSO](https://aws.amazon.com/blogs/security/aws-single-sign-on-now-enables-command-line-interface-access-for-aws-accounts-using-corporate-credentials/), you can request temporary CLI keys that work like this.
* Another case we bump into this is in our build environment where temporary keys are created that also work like this.

Currently you are already able to do this by setting the three through environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY and AWS_SESSION_TOKEN) and [boto3 will pick it up](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables). 

However, we have DNS zones spread over several AWS accounts which means we can't do that.

octodns/provider/route53.py
```diff
@@ -221,9 +221,12 @@ class Route53Provider(BaseProvider):
         access_key_id:
         # The AWS secret access key
         secret_access_key:
+        # The AWS session token
```
ross: Probably a good idea to indicate that this one is optional or otherwise call out that `session_token` is only required if you need it. Mainly don't want to confuse people and make them think they have to come up with one to use the provider.

zBart:
I added a note, but I'm not sure if this makes it much clearer.

ross:
Looks like after fixing a minor lint issue CI is failing with this change. I haven't dug into what's up yet, but we'll need to get that sorted before this can merge.

zBart:
@ross sorry for poking you, was wondering if you saw my previous message.

Build is failing because this call: https://github.com/zBart/octodns/blob/patch-2/tests/test_octodns_provider_route53.py#L1282 assumes that the 4th argument is the max_changes one.

Should I move the argument to the end of the list or change the test to use keyword arguments?

ross:
> Should I move the argument to the end of the list or change the test to use keyword arguments?

I'd say both really. Probably best to preserve backwards compatibility by moving the new session param to the end of the list in the constructor in case people are passing things positionally like the test is.

It's also probably a good idea to use `param=value` in the invocation, but up to you whether you want to make that "fix" since it's not related to your change.

Thanks.

zBart:
@ross it only took 8 commits, but the build succeeds now ;-). So much for thinking this would be a quick fix.

I didn't change the test, figured it wouldn't have broken otherwise and then we wouldn't have thought that adding a parameter in the argument list isn't a great idea.

Issue #304, opened 2018-12-06
Add A-Name support to DNSMadeEasy Provider


fahanson:
DNS Made Easy has a type of Alias Record called an "A-Name". Some of our infrastructure is still hosted in Heroku which requires A-Names for custom domain names. This would be a huge win for us so we can manage all our domains with OctoDNS.

ross:
It should be possible to wire things up so that a octoDNS `ALIAS` record is mapped to an "A-name" in DNSMadeEasy. There are a few other providers that do similar types of mapping already, e.g. Cloudflare's is just a CNAME on the root. I won't have the bandwidth to look into it in the near term, but would gladly look over an PR if someone gets a chance to look into it.

fahanson:
My Python is decent enough that I could probably take at least a first pass at something for you to look at and suggest changes on.

fahanson:
@yzguy If you could add this that would be awesome!

yzguy:
Sure thing @fahanson, let me check it out.

PR #306, opened 2018-12-10
add ANAME support to DnsMadeEasy (ALIAS == ANAME)

yzguy:
Adds requested feature from #304 

This maps an OctoDNS ALIAS record to DnsMadeEasy's ANAME record. In looking at their ANAME support, they allow IP addresses as values for ANAME records, which seems silly as then it's just an A record at that point, so I did not handle that case, if it's common enough I guess we could.

**Example OctoDNS config**

```
---
? ''
: - ttl: 1800
    type: ALIAS
    value: www.example.com.
  - type: MX
```

OctoDNS will dump all ANAME records to being ALIAS records
OctoDNS will apply all ALIAS records as being ANAME records

/cc @fahanson


yzguy:
Hold on, there is a big during updates.

yzguy:
Refactored, fixed the update issue.

PR #314, opened 2019-01-13
Adding more details to OctoDNS docs

volym3ad:
- Add description to `default_ttl` and `enforce_order` keys in config
- Replace `priority/value` to `preference/exchange` in MX example. As far as I see from code and my personal experience with OctoDNS this is the right syntax.

I've added these commits since I think they'll be useful for newbies when it comes to first interaction with OctoDNS. Personally it could help me a lot when I had started.

Hopes it helps. Ready to listen your thoughts around this pr. Thanks in advance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ providers:
   config:
     class: octodns.provider.yaml.YamlProvider
     directory: ./config
+    default_ttl: 3600
+    enforce_order: True
   dyn:
     class: octodns.provider.dyn.DynProvider
     customer: 1234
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,6 +58,10 @@ zones:
 
 `class` is a special key that tells OctoDNS what python class should be loaded. Any other keys will be passed as configuration values to that provider. In general any sensitive or frequently rotated values should come from environmental variables. When OctoDNS sees a value that starts with `env/` it will look for that value in the process's environment and pass the result along.
 
+`enforce_order` is a special key that tells OctoDNS to enforce sorting order on the yaml config. By default it is set to True.
+
+If you don't want to set same TTL each time for your records, you can optionally use `default_ttl` key - it will be used in records when not specified in the data. By default it is set to 3600 seconds.
+
```
ross: `enforce_order` and `default_ttl` are both specific to `YamlProvider` and not part of the stuff being documented here (the manager config file.) It seems fine/reasonable to include them in the example up above where `config` is used, but documenting the provider stuff here is probably a bit out of scope. `YamlProvider` does have a class docstring that includes info about them. Not opposed to having separate markdown documentation for the providers, but don't think it belongs here.

docs/records.md
```diff
@@ -120,10 +120,10 @@ In the above example each name had a single record, but there are cases where a
...
   - type: MX
     values:
-      - priority: 10
-        value: mx1.example.com.
-      - priority: 10
-        value: mx2.example.com.
+      - exchange: mx1.example.com.
+        preference: 10
+      - exchange: mx2.example.com.
+        preference: 10
```
ross: Good call. `preference` is the preferred key.

README.md
volym3ad: Ok, I'll remove `enforce_order` and `default_ttl` description for now. Maybe it's really better to create separate doc for `YamlProvider` and others.

PR #315, opened 2019-01-14
Add Lenience Flag to Sync

oliverisaac:
This PR:

- Adds a lenience flag to the sync command
- When enabled this flag:
    - Allows non-spec SRV records (e.g.: dig SRV mongodb.example.org )
    - Disables check_origin on AXFR import
- Makes the SRV record "invalid name" more descriptive

I created this PR so we can do AXFR's against CoreDNS in a kubernetes cluster. CoreDNS creates non-spec SRV records and does not include the required NS records in an AXFR.

All together, this PR allow us to have CoreDNS create and manage DNS records in the cluster but have Google Cloud DNS actually serve the DNS records.

### DIFF
--- a/octodns/cmds/sync.py
+++ b/octodns/cmds/sync.py
@@ -22,6 +22,9 @@ def main():
                         help='Acknowledge that significant changes are being '
                         'made and do them')
 
+    parser.add_argument('--lenient', action='store_true', default=False,
+                        help='Do not strictly follow DNS standard.')
+
     parser.add_argument('zone', nargs='*', default=[],
                         help='Limit sync to the specified zone(s)')
 
@@ -36,7 +39,7 @@ def main():
 
     manager = Manager(args.config_file)
     manager.sync(eligible_zones=args.zone, eligible_targets=args.target,
-                 dry_run=not args.doit, force=args.force)
+                 dry_run=not args.doit, force=args.force, lenient=args.lenient)
 
 
 if __name__ == '__main__':
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -217,7 +217,8 @@ class Manager(object):
 
     def _populate_and_plan(self, zone_name, sources, targets, lenient=False):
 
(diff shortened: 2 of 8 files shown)


ross:
> All together, this PR allow us to have CoreDNS create and manage DNS records in the cluster but have Google Cloud DNS actually serve the DNS records.

:wave: @oliverisaac. That sounds cool. 

Sorry I missed this PR come through when you originally filed it. In general it seems reasonable/on track. I haven't had a chance to dig into the PR deeply, but noticed that `lenient` had to be added to a couple providers's calls to `Record.new` and that we already have it in a few other places. Did you go through and audit the providers as part of the addition or just add it in axfr & google cloud since you were using them. It looks like a couple others don't have it and I was curious to know if you looked at those and skipped them for a reason or just hadn't inspected them.

oliverisaac:
I didn't audit the other providers or sources. My goal with this was to be backwards compatible so that if someone is not using the lenience flag they shouldn't see a difference. To that end, I made as few changes as possible to get this to work in my context (AXFR and Google DNS)

octodns/cmds/sync.py
```diff
@@ -22,6 +22,9 @@ def main():
                         help='Acknowledge that significant changes are being '
                         'made and do them')
 
+    parser.add_argument('--lenient', action='store_true', default=False,
+                        help='Do not strictly follow DNS standard.')
```
ross: My general worry with this is that `lenient` is also passed to `Zone.add_record` and it allows more than just adding records that don't follow the standards.

https://github.com/github/octodns/blob/7bf4914a41042cd283cf8cecdec88a9e37f2dbb1/octodns/zone.py#L59-L93

This text probably needs to be made more broad. Something to the effect of:

> Allow records that do not conform to DNS standards to be synced as well as records that below in sub-zones

This actually may be a positive thing in terms of allowing the break out of sub-zones (though providers behaviors around this are super wonky/inconsistent.)

danielmittelman:
Hi @oliverisaac, any progress with this PR? Need assistance completing it?

We're waiting for this feature and would love to see it released!

PR #292, opened 2018-11-03
Include "special" _octodns values in Record.__repr__

ross:
This includes the `octodns` bits in `Record.__repr__`. The upside of this is that you can see the special bits that may be pushed to your providers, downside is that it could be noisy and not all special things apply to all providers and things pulled out of providers will only include the things that provider knows about.

## Example of the "good" case, CF proxied

```yaml
'*.sub':
  octodns:
    cloudflare:
      proxied: True
  ttl: 144
  type: A
  values: 
  - 2.2.3.3
  - 3.3.4.4
  - 4.4.5.5
```

```
********************************************************************************
* githubtest.net.
********************************************************************************
* cloudflare (CloudflareProvider)
*   Update
*     <ARecord A 144, *.sub.githubtest.net., ['2.2.3.3', '3.3.4.4', '4.4.5.5'], cloudflare=cloudflare.proxied=False> ->
*     <ARecord A 144, *.sub.githubtest.net., ['2.2.3.3', '3.3.4.4', '4.4.5.5'], cloudflare=cloudflare.proxied=True> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=20
********************************************************************************
```

## Example of the "bad" case, where things aren't super clear

```yaml
'*.sub':
  octodns:
    cloudflare:
      other: False
      proxied: True
    something: True
  ttl: 144
  type: A
  values: 
  - 2.2.3.3
  - 3.3.4.4
  - 4.4.5.5
```

```
********************************************************************************
* githubtest.net.
********************************************************************************
* cloudflare (CloudflareProvider)
*   Update
*     <ARecord A 144, *.sub.githubtest.net., ['2.2.3.3', '3.3.4.4', '4.4.5.5'], cloudflare=cloudflare.proxied=False> ->
*     <ARecord A 144, *.sub.githubtest.net., ['2.2.3.3', '3.3.4.4', '4.4.5.5'], cloudflare=cloudflare.other=False, cloudflare.proxied=True, something=True> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=20
********************************************************************************
```

/cc https://github.com/github/octodns/issues/284
/cc @parkr @begincalendar

PR #316, opened 2019-01-28
Version bump to 0.9.4 & CHANGELOG.md update

ross:
Cutting a release as a bunch of small things have changed since the last round and it's still going to be a bit before the `dynamic` stuff is fully ready to land.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,11 @@
-## v0.9.4 - ....-..-.. - Dynamic Records Beta
+## v0.9.4 - 2019-01-28 - The one with a bunch of stuff, before the big one
 
-* Dynamic Records (more powerful/flexible replacement for geo)
-   * Will support A, AAAA, and CNAME out the gate and include the ability to
-     weight records. It should provide a foundation for further suppport
-     if/when needed.
-   * Major refactoring and improvements to validation of the Record hierarchy,
-     things are much more consisntely implemented now and error messages should
-     be more actionable/clear. Both the base values and dynamic values use the
-     same validatio logic.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,13 +1,11 @@
...
-   * Major refactoring and improvements to validation of the Record hierarchy,
-     things are much more consisntely implemented now and error messages should
-     be more actionable/clear. Both the base values and dynamic values use the
-     same validatio logic.
+## v0.9.4 - 2019-01-28 - The one with a bunch of stuff, before the big one
+
+* A bunch of "dynamic" stuff that'll be detailed in the next release when
+  providers actually support it :grin:
+* Route53Provider adds support for using session tokens
+* Added support for proxying Cloudflare ALIAS records
```
parkr: I think I can set proxying for all Cloudflare records!

parkr: Speaking of which, there may be a tiny bit of confusion on this flag until https://github.com/github/octodns/pull/292 is shipped

ross: > Speaking of which, there may be a tiny bit of confusion on this flag until #292 is shipped

Yeah. I don't think it's going to be any worse than it already was as a result of this :ship: since proxy and similar stuff existed before, but I agree. I'd like to return to #292 or something broader that does better formatting in general as it'd help a lot with that and with the more complex `geo`/`dynamic` records as well.

ross: Oh yeah https://github.com/github/octodns/pull/307 has bits and pieces of that, but it turned into a mess that I didn't have the space to sort out at the time. I'll probably pick it back up eventually.

PR #326, opened 2019-03-04
Add healthcheck option 'request_interval' for Route53 provider

jrunkel:
Route53 allows to specify an interval for its health checks.
To maintain backward compatibility, the default for this option
when ommited is 10 (fast check).


octodns/provider/route53.py
```diff
@@ -550,14 +550,21 @@ def _healthcheck_measure_latency(self, record):
             .get('healthcheck', {}) \
             .get('measure_latency', True)
 
+    def _healthcheck_request_interval(self, record):
+        interval = record._octodns.get('route53', {}) \
+            .get('healthcheck', {}) \
+            .get('request_interval')
+        return interval if (interval in [10, 30]) else 10
```
ross: If these are the only two values allowed it'd be good to throw an exception or at least warning if we get something else.

jrunkel:
I added code to throw an exception if the value is not allowed.

octodns/provider/route53.py
```diff
@@ -550,14 +550,25 @@ def _healthcheck_measure_latency(self, record):
...
 
+    def _healthcheck_request_interval(self, record):
+        interval = record._octodns.get('route53', {}) \
+            .get('healthcheck', {}) \
+            .get('request_interval', 10)
+        if (interval in [10, 30]):
+            return interval
+        else:
+            raise Exception('route53.healthcheck.request_interval '
+                            'parameter must be either 10 or 30.')
```
ross: Going the `raise` route it'd be preferable to create a specific exception to throw rather than the generic `Exception`. For example down below in the test it's catching `Exception` which would handle any problem, not just this one raising. It can be as simple as:

```python
class Route53ProviderException(Exception):
    pass
```

and then `s/Exception/Route53ProviderException/`.

jrunkel:
Like this? Not sure where exactly you like the Exception to be defined.

iaguis:
This would be useful for us. Any chance to get this merged?

ross:
> This would be useful for us. Any chance to get this merged?

Slipped through the notifications/cracks. Queued it up to look at and get shipped if it's ready.

iaguis:
Thank you!

ross:
Wow this one was stale enough that merging in master in https://github.com/github/octodns/pull/326/commits/c5a19b4403e6df393aaef793fd2c438d2c7756d2 was a huge pain. The worse conflict resolution I've run across in years. I think that has it though.

Issue #327, opened 2019-03-07
AttributeError: 'DnsMadeEasyProvider' object has no attribute '_data_for_HTTPRED'


snero:
Hi, 

I am trying to dump/sync to our dns made easy provider and keep getting this error 

`AttributeError: 'DnsMadeEasyProvider' object has no attribute '_data_for_HTTPRED'` 

and idea on what is happening here?

snero:
I figured this one out. It appears that HTTPRED is a particular type of DNS record that is only available dnsmadeeasy. I am thinking of mapping the call to cname instead so that it is compatible with other DNS systems

ross:
Going to go ahead and keep this open. You are correct that the record type isn't supported by octoDNS, but the provider should generally handle that, warn about it, and skip over it rather than throw an error.

ross:
Fixed in https://github.com/github/octodns/pull/329

PR #329, opened 2019-03-22
add supports check to DNSMadeEasy Provider

yzguy:
Addresses #327 

Any record type not in the `self.SUPPORTS` will show a warning and be skipped. HTTPRED happens to be a specific record type to DNSMadeEasy, but is not really a record elsewhere.


ross:
Thanks!

PR #333, opened 2019-04-01
Route53Provider dynamic record support

ross:
- [x] Code up a POC/mostly working setup that's uncommitted and messy (just happens to be how I went about this.)
- [x] Break up ^ into logical/meaningful commits and make sure bits are tested as committed
- [x] Write tests for all the new dynamic bits 
- [x] Go back through and address remaining TODOs or call them out as 🏈s
- [x] Review
- [x] :shipit:

### DIFF
--- a/.git_hooks_pre-commit
+++ b/.git_hooks_pre-commit
@@ -8,4 +8,4 @@ ROOT=`dirname $GIT`
 
 . $ROOT/env/bin/activate
 $ROOT/script/lint
-$ROOT/script/test
+$ROOT/script/coverage
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -800,7 +800,6 @@ class Route53Provider(BaseProvider):
             else:
                 # These are the pool value(s)
                 pool_name = rrset['SetIdentifier'][:-4]
-                # TODO: handle different value types
                 value = rrset['ResourceRecords'][0]['Value']
                 pools[pool_name]['values'].append({
                     'value': value,
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -235,6 +235,35 @@ class _Route53GeoRecord(_Route53Record):
                                                           self.values)
 
 
+_mod_keyer_action_order = {
+    'DELETE': '0',  # Delete things first
+    'CREATE': '1',  # Then Create things
+    'UPSERT': '2',  # Upsert things last
+}
+
+
+def _mod_keyer(mod):
+    rrset = mod['ResourceRecordSet']
+    action_order = _mod_keyer_action_order[mod['Action']]
+
+    # We're sorting by 3 "columns", the action, the rrset type, and finally the
+    # name/id of the rrset
+
+    if rrset.get('GeoLocation', False):
+        return '{}3{}'.format(action_order, rrset['SetIdentifier'])
+    elif rrset.get('AliasTarget', False):
+        # We use an alias
+        if rrset.get('Failover', False) == 'SECONDARY':
+            # We're a secondary we'll ref primaries
+            return '{}2{}'.format(action_order, rrset['Name'])
+        else:
+            # We're a primary we'll ref values
+            return '{}1{}'.format(action_order, rrset['Name'])
+
+    # We're just a plain value, these come first
+    return '{}0{}'.format(action_order, rrset['Name'])
+
+
 class Route53Provider(BaseProvider):
     '''
     AWS Route53 Provider
@@ -827,6 +856,11 @@ class Route53Provider(BaseProvider):
             # Generate the mods for this change
             mod_type = getattr(self, '_mod_{}'.format(c.__class__.__name__))
             mods = mod_type(c, zone_id)
+
+            # Order our mods to make sure targets exist before alises point to
+            # them and we CRUD in the desired order
+            mods.sort(key=_mod_keyer)
+
             mods_rs_count = sum(
                 [len(m['ResourceRecordSet']['ResourceRecords']) for m in mods]
             )
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -14,6 +14,7 @@ import logging
 import re
 
 from ..record import Record, Update
+from ..record.geo import GeoCodes
 from .base import BaseProvider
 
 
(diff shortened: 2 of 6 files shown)


.git_hooks_pre-commit
```diff
@@ -8,4 +8,4 @@ ROOT=`dirname $GIT`
 
 . $ROOT/env/bin/activate
 $ROOT/script/lint
-$ROOT/script/test
+$ROOT/script/coverage
```
ross: Now that coverage enforces 100% this updates the hook to use it.

octodns/record/__init__.py
```diff
@@ -1067,15 +1067,18 @@ class _ChunkedValuesMixin(_ValuesMixin):
     CHUNK_SIZE = 255
     _unescaped_semicolon_re = re.compile(r'\w;')
 
+    def chunked_value(self, value):
+        value = value.replace('"', '\\"')
+        vs = [value[i:i + self.CHUNK_SIZE]
+              for i in range(0, len(value), self.CHUNK_SIZE)]
+        vs = '" "'.join(vs)
+        return '"{}"'.format(vs)
```
ross: This logic needed to be broken out of the values version below so that it could be called on individual values by _Route53Record. More on this below.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
 class _Route53Record(object):
 
     @classmethod
-    def new(self, provider, record, creating):
```
ross: For the most part this shifted down and I broke out some of it's logic into `_new_geo`. `_new_dynamic` should be wholly new code. It bascially creates the various RRSets that make up a dynamic record in Route53.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
-            ret.add(_Route53Record(provider, record, creating))
+
+        # HostedZoneId wants just the last bit, but the place we're getting
+        # this from looks like /hostedzone/Z424CArX3BB224
+        hosted_zone_id = hosted_zone_id.split('/', 2)[-1]
+
+        # Create the default pool
+        fqdn = record.fqdn
+        ret.add(_Route53Record(provider, record, creating,
+                               '_octodns-default-pool.{}'.format(fqdn)))
```
ross: This is the pool of records that come from the base `value(s)` of the record and gets used when all of the other dynamic values fail their health checks. In general that'll hopefully never happen.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+
+        # Pools
+        for pool_name, pool in record.dynamic.pools.items():
+
+            # Create the primary
+            ret.add(_Route53DynamicPool(provider, hosted_zone_id, record,
+                                        pool_name, creating))
```
ross: This is the actual pool that geo targeted records below will point to. It is the primary, aliases to the `N` `-value`s below and considers their health when falling back.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+
+            # Create the fallback
+            fallback = pool.data.get('fallback', False)
+            if fallback:
+                # We have an explicitly configured fallback
+                ret.add(_Route53DynamicPool(provider, hosted_zone_id, record,
+                                            pool_name, creating,
+                                            target_name=fallback))
```
ross: We have an explicit fallback configured. This SECONDARY will alias to that fallback.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+            else:
+                # We fallback on the default
+                ret.add(_Route53DynamicPool(provider, hosted_zone_id, record,
+                                            pool_name, creating,
+                                            target_name='default'))
```
ross: No explicit fallback, use the default (unhealthchecked) pool.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+
+            # Create the values
+            for i, value in enumerate(pool.data['values']):
+                weight = value['weight']
+                value = value['value']
+                ret.add(_Route53DynamicValue(provider, record, pool_name,
+                                             value, weight, i, creating))
```
ross: This are the weighted values for this pool. The PRIMARY will point to them and consider their health when making its decisions.

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+
+        # Rules
+        for i, rule in enumerate(record.dynamic.rules):
+            pool_name = rule.data['pool']
+            geos = rule.data.get('geos', [])
+            if geos:
+                for geo in geos:
+                    ret.add(_Route53DynamicRule(provider, hosted_zone_id,
+                                                record, pool_name, i,
+                                                creating, geo=geo))
```
ross: Create an RRSet for each geo within each rule that links that Geo to the target pool (and through that pool it's values/fallback)

octodns/provider/route53.py
```diff
@@ -29,19 +31,84 @@ def _octal_replace(s):
...
+            else:
+                ret.add(_Route53DynamicRule(provider, hosted_zone_id, record,
+                                            pool_name, i, creating))
```
ross: We have no geo configured so we're take anything that's left and point it to the configured pool. So long as it's healthy it'll pick up all the otherwise untargeted stuff.

octodns/provider/route53.py
```diff
@@ -83,46 +150,235 @@ def __repr__(self):
         return '_Route53Record<{} {} {} {}>'.format(self.fqdn, self._type,
                                                     self.ttl, self.values)
 
+    def _value_convert_value(self, value, record):
+        return value
```
ross: All of these converts were broken out of the `_values_for_*` functions so they could be used directly when converting each `dynamic` value for a record based on its type.

octodns/provider/route53.py
```diff
@@ -211,6 +468,35 @@ def __repr__(self):
...
+    'UPSERT': '2',  # Upsert things last
+}
+
+
+def _mod_keyer(mod):
+    rrset = mod['ResourceRecordSet']
+    action_order = _mod_keyer_action_order[mod['Action']]
+
+    # We're sorting by 3 "columns", the action, the rrset type, and finally the
+    # name/id of the rrset
```
ross: This sorting ensures that Route53 doesn't have problems with the order of the modifications we send it. We need to make sure that anything that's referenced by an Alias comes before that reference. We're doing that by sorting of types rather than explicitly looking through the aliases and targets.

octodns/provider/route53.py
```diff
@@ -465,6 +750,75 @@ def _load_records(self, zone_id):
 
         return self._r53_rrsets[zone_id]
 
+    def _data_for_dynamic(self, name, _type, rrsets):
```
ross: This converts a bunch of "dynamic" rrsets into corresponding Record data, it's used by populate.

octodns/provider/route53.py
```diff
@@ -551,27 +931,44 @@ def _healthcheck_measure_latency(self, record):
...
     def _health_check_equivilent(self, host, path, protocol, port,
-                                 measure_latency, health_check,
-                                 first_value=None):
+                                 measure_latency, health_check, value=None):
         config = health_check['HealthCheckConfig']
+
+        # So interestingly Route53 normalizes IPAddress which will cause us to
+        # fail to find see things as equivalent. To work around this we'll
+        # ip_address's returned object for equivalence
+        # E.g 2001:4860:4860::8842 -> 2001:4860:4860:0:0:0:0:8842
```
ross: This would have been a bug with `AAAA` geo records as well. Guess no one is using them or if they are they're fully specified.

octodns/provider/route53.py
```diff
@@ -606,28 +1003,44 @@ def get_health_check_id(self, record, ident, geo, create):
...
+        self._conn.change_tags_for_resource(ResourceType='healthcheck',
+                                            ResourceId=id,
+                                            AddTags=[{
+                                                'Key': 'Name',
+                                                'Value': name,
+                                            }])
+        # Manually add it to our cache
+        health_check['Tags'] = {
+            'Name': name
+        }
```
ross: This just adds a `Name` tag to the healthchecks that makes them a lot easier to make sense of in the UI...

octodns/provider/route53.py
```diff
@@ -704,14 +1118,91 @@ def _mod_Update(self, change):
...
-    def _mod_Delete(self, change):
+    def _mod_Delete(self, change, zone_id):
         # Existing is the thing that needs to be deleted
-        existing_records = self._gen_records(change.existing, creating=False)
+        existing_records = self._gen_records(change.existing, zone_id,
+                                             creating=False)
         # Now is a good time to clear out all the health checks since we know
         # we're done with them
         self._gc_health_checks(change.existing, [])
         return self._gen_mods('DELETE', existing_records)
```
ross: The next few functions just break up and rework _extra_changes to support dynamic records. The code is pretty different, but the before and after logic is identical for geo and just adds support for dynamic.

octodns/provider/route53.py
```diff
@@ -211,6 +468,35 @@ def __repr__(self):
...
+
+def _mod_keyer(mod):
+    rrset = mod['ResourceRecordSet']
+    action_order = _mod_keyer_action_order[mod['Action']]
+
+    # We're sorting by 3 "columns", the action, the rrset type, and finally the
+    # name/id of the rrset
+
+    if rrset.get('GeoLocation', False):
+        return '{}3{}'.format(action_order, rrset['SetIdentifier'])
```
theojulienne: Any reason we can't return a tuple to sort by, e.g. `(action_order, 3, rrset['SetIdentifier'])`? I think that would have the same sorting properties but avoid needing to munge it into a string.

octodns/provider/route53.py
```diff
@@ -465,6 +750,75 @@ def _load_records(self, zone_id):
...
+                'rules': [],
+            }
+        }
+
+        # For all the rrsets that comprise this dynamic record
+        for rrset in rrsets:
+            name = rrset['Name']
+            if '-pool.' in name:
+                # This is a pool rrset
+                pool_name = name.split('.', 1)[0][9:-5]
```
theojulienne: is `[9:-5]` stripping `_octodns-...-pool`? Might be good to have a comment to note this, took me a sec :smile:

octodns/provider/route53.py
```diff
@@ -465,6 +750,75 @@ def _load_records(self, zone_id):
...
+                    # we'll record
+                    if fallback_name != 'default':
+                        pools[pool_name]['fallback'] = fallback_name
+            elif 'GeoLocation' in rrset:
+                # These are rules
+                _id = rrset['SetIdentifier']
+                # We record rule index as the first part of set-id, the 2nd
+                # part just ensures uniqueness across geos and is ignored
+                i = int(_id.split('-', 1)[0])
+                # Parse the pool name out of _octodns-<pool-name>-pool.
```
theojulienne: Ah I see the comment here - might just be good to drop the comment in at each usage since it's a bit unclear without it.

octodns/provider/route53.py
ross: No, just didn't occur to me when things were evolving through various ordering options. Pushing up a switch shortly.

octodns/provider/route53.py
ross: Pulled this duplicated parsing bit out into a helper function with the doc and doc'd the singular case where it's parsing SetIdentifier below.

ross:
Just ran through a "complete" test of the fallback before and things work as desired/expected.

ross:
> The GH review comments around the overall logic were really useful, would be great to copy those in to inline code comments for posterity/future readers.

Moved everything that wasn't specific to the review process & refactoring into actual code comments. @theojulienne this should be ready for final review.

PR #336, opened 2019-04-04
Allow multiple YAML files to define zones

cfunkhouser:
Introduce `SplitYamlProvider` to manage zones in multiple YAML files, instead of a single, monolithic YAML file. The rationale is that I'm preparing to deploy OctoDNS as the source of truth for two very complex zones across multiple providers. The monolithic YAML file generated by `octodns-dump` was many thousands of lines in length, and somewhat unwieldy to search/modify/understand. Additionally, I'm investigating using `CODEOWNERS` to allow delegation of responsibility over particular hosts to various teams, which is not possible when the entire zone is contained in a single YAML file.

This change modifies the `YamlProvider` slightly in order to allow it to be extended. When using the `SplitYamlProvider`, the expectation is that a directory will contain the entire zone's record definitions, with each YAML file representing all records with the same name. For example:

```
zones/
  example.com./
    $example.com.yaml
    www.yaml
    mail.yaml
```
Any two records with the same name and different types (ie, `A` and `AAAA` for the same host) will be contained in the same file. For example, `www.yaml` above may look like:

```
---
www:
- ttl: 300
  type: A
  value: 12.34.56.78
- ttl: 300
  type: AAAA
  value: fe80::12:34:56:78
```

The file prefixed with the `$` is the "catch-all" file, which contains the top-level domain records (`A`, `AAAA`, `MX`, etc) and any wildcard domains. The idea is that it will contain anything that is tricky to represent as a filename. It also appears first when listing the directory contents. The `$example.com.yaml` file might look like:

```
---
? ''
: - ttl: 30
    type: A
    values:
    - 45.6.7.8
    - 54.3.2.1
  - ttl: 7200
    type: MX
    values:
    - exchange: mx1.example.lol.
      preference: 10
'*':
  ttl: 30
  type: CNAME
  value: wildcard.example.com.
```

In addition to the `SplitYamlProvider`, I've added a `--split` flag to `octodns-dump` to allow exporting zones in multiple files in the first place.


cfunkhouser:
My mistake, I assumed I was running the tests correctly locally. I'll fix this in a moment.

cfunkhouser:
Now that I am _actually_ running the tests correctly, I can see I've got some more work to do. I'll resubmit this when it's ready. Sorry for the noise.

ross:
No need to close and reopen. You can work on it in this existing PR while it's still open/pending.

cfunkhouser:
Fair enough, I've re-opened. The last failing test is not failing locally for me, so I'm going to have to dig in.

cfunkhouser:
Well, that's curious. Starting with a clean zone in the second half of the test causes it to pass on CI. It always passed for me locally (on OS X). Also, the not-split `TestYamlProvider`'s `test_unsorted` does the same thing with no problems.

ross:
Not sure what's going wrong here, but the PR seems to be showing diffs for a bunch of other commits outside of your changes.

cfunkhouser:
Rebase sadness corrected, sorry.

octodns/provider/yaml.py
```diff
@@ -102,7 +107,97 @@ def _apply(self, plan):
...
         with open(filename, 'w') as fh:
             safe_dump(dict(data), fh)
+
+
+# Any record name added to this set will be included in the catch-all file,
+# instead of a file matching the record name.
+_CATCHALL_RECORD_NAMES = ('*', '')
+
+
+def list_all_yaml_files(directory):
```
ross: General practice in octoDNS is to prefix internal helper methods like this with `_`.  Alternatively it'd be fine to make it a `@classmethod` on `SplitYamlProvider`.

tests/config/dynamic.tests.yaml
```diff
@@ -29,6 +29,7 @@ a:
         pool: ams
       - geos:
           - NA-US-CA
+          - NA-US-NC
```
ross: Any specific reason for the NC addition?

tests/test_octodns_provider_splityaml.py
```diff
@@ -0,0 +1,212 @@
+#
```
ross: Would prefer to either include these tests in `tests/test_octodns_provider_yaml.py` or to spilt the provider out into two modules (`octodns/provider/split_yaml.py`) so that the module and tests match up grouping-wise.

octodns/provider/yaml.py
```diff
@@ -102,7 +107,97 @@ def _apply(self, plan):
...
+class SplitYamlProvider(YamlProvider):
+    '''
+    Core provider for records configured in multiple YAML files on disk.
+
+    Behaves mostly similarly to YamlConfig, but interacts with multiple YAML
+    files, instead of a single monolitic one. The files are named RECORD.yaml,
+    except for any record which cannot be represented easily as a file; these
+    are stored in the catchall file, which is a YAML file the zone name,
+    prepended with a '$'. For example, a zone, 'github.com.' would have a
+    catch-all file named '$github.com.yaml'.
```
ross: 👍 I like that this puts things in `<configured-directory>/<zone-name>/...`, was going to ask that it does if it didn't. It wasn't clear that was going to be the case from the info/docs here. Would be good to get a sentence and/or example here calling that out.

octodns/provider/yaml.py
```diff
@@ -102,7 +107,97 @@ def _apply(self, plan):
...
+        # (optional, default 3600)
+        default_ttl: 3600
+        # Whether or not to enforce sorting order on the yaml config
+        # (optional, default True)
+        enforce_order: True
+    '''
+
+    def __init__(self, id, directory, *args, **kwargs):
+        super(SplitYamlProvider, self).__init__(id, directory, *args, **kwargs)
+        self.log = logging.getLogger('SplitYamlProvider[{}]'.format(id))
```
ross: I believe the `self.log =` needs to come before the super call so that the parent class can use it.

octodns/provider/yaml.py
```diff
@@ -102,7 +107,97 @@ def _apply(self, plan):
...
+    def _do_apply(self, desired, data):
         filename = join(self.directory, '{}yaml'.format(desired.name))
         self.log.debug('_apply:   writing filename=%s', filename)
         with open(filename, 'w') as fh:
             safe_dump(dict(data), fh)
+
+
+# Any record name added to this set will be included in the catch-all file,
+# instead of a file matching the record name.
+_CATCHALL_RECORD_NAMES = ('*', '')
```
ross: Would prefer this get moved to be a property on `SplitYamlProvider`, and it can lose the `_` and be referred to as `self.CATCHALL_RECORD_NAMES` then.

tests/config/dynamic.tests.yaml
cfunkhouser: Pure romanticism, since that's where I live. I'll happily yank it back out if you'd prefer.

octodns/provider/yaml.py
cfunkhouser: I opted for the `_`-prefixed approach.

tests/test_octodns_provider_splityaml.py
cfunkhouser: Rolled the tests into `tests/test_octodns_provider_yaml.py`, because I think specifying a provider class name `octodns.provider.yaml.SplitYamlProvider` is a bit cleaner.

octodns/provider/yaml.py
cfunkhouser: That's a little tricky, because if specified _before_ calling `super(..)__init__(..)`, the logger gets blown away with the new one. I've modified the logger creation in `YamlProvider` to be dynamic, and removed it from `SplitYamlProvider` entirely. Let me know what you think.

octodns/provider/yaml.py
cfunkhouser: I have clarified in the PyDoc of `SplitYamlProvider`.

octodns/provider/yaml.py
ross: 👍 that's the right way to go.

Issue #342, opened 2019-04-11
Add support for IPv6 extensions to TinyDNS


adhawkins:
There are some patches to TinyDNS that allow it to serve AAAA records.

The patch is hosted here: https://www.fefe.de/dns/

The record types in the file are '3' (AAAA only) and '6' (AAAA and associated PTR).

Some more documentation is available here:

https://docs.bytemark.co.uk/article/tinydns-format/

ross:
I'd be happy to take a look at a PR. Won't likely be able to pick this up myself.

adhawkins:
Addressed in MR #343.

Issue #340, opened 2019-04-11
Allow TXT records when reading TinyDNS files


adhawkins:
Symbiosis (https://symbiosis.bytemark.co.uk/stretch/docs/symbiosis.html) and its fork Sympl (https://sympl.host/) can generate TXT records in TinyDNS format. However, octodns does not consider a TXT record from TinyDNS as valid.

An example TXT record:

`'gentlyhosting.uk:v=spf1 +a +mx ?all:300`

The format is documented here:

https://cr.yp.to/djbdns/tinydns-data.html

It would be good to allow octodns to be able to read in (and hence sync) these TXT records from TinyDNS files

ross:
Similar to https://github.com/github/octodns/issues/342 PRs welcome

adhawkins:
Pretty sure this is closed as a result of PR #343

PR #343, opened 2019-04-12
Tinydns additions

adhawkins:
Add support for TinyDNS TXT (#340) and AAAA (#342) records.

First pull request, so comments appreciated.


octodns/source/tinydns.py
```diff
@@ -45,6 +46,37 @@ def _data_for_A(self, _type, records):
             'values': values,
         }
 
+    def _data_for_AAAA(self, _type, records):
+        values = []
+        for record in records:
+            values.append(u":".join(textwrap.wrap(record[0], 4)))
```
ross: What's going on with the textwrap stuff here? I'm not grokking what's happening here. It'll definitely need an explanatory comment if it's required, but it's fine to start discussion here first.

adhawkins: TinyDNS files have the ipv6 address written in full, but with the colons removed. This just inserts a colon every 4th character to make the address correct.

Will add a comment when I get chance.

ross: Ugh. OK. Please add a comment to that effect in the code as it's definitely non-obvious/standard.

adhawkins: Will do. Or can you suggest a clearer and perhaps more Pythonic way of doing it?

ross: I can't think of anything cleaner and it sounds like `textwrap.wrap` makes sense here. The suprising/weird part to me is that TinyDNS doesn't have `:`. 

> TinyDNS files have the ipv6 address written in full, but with the colons removed. This just inserts a colon every 4th character to make the address correct.

Will work as the comment.

adhawkins: Comment added. Let me know if you need anything further.

PR #337, opened 2019-04-04
Normalize MX exchange to lower-case, fix #322

PeterDaveHello:

Issue #322, opened 2019-02-12
Normalize records for GCP?


PeterDaveHello:
Hello,

As GCP Cloud DNS store the alphabets in domain names and most of the records (looks like except TXT) in lower case, I wonder is it possible to normalize that before the update of comparison?

Looks like updates like below is not really meaningful, as GCP will still store them in lower case:
```
********************************************************************************
* example.com.
********************************************************************************
* gcp (GoogleCloudProvider)
*   Update
*     <MxRecord MX 60, example.com., [''1 aspmx.l.google.com.'', ''5 alt1.aspmx.l.google.com.'', ''5 alt2.aspmx.l.google.com.'', ''10 aspmx2.googlemail.com.'', ''10 aspmx3.googlemail.com.'']> ->
*     <MxRecord MX 60, example.com., [''1 ASPMX.L.GOOGLE.COM.'', ''5 ALT1.ASPMX.L.GOOGLE.COM.'', ''5 ALT2.ASPMX.L.GOOGLE.COM.'', ''10 ASPMX2.GOOGLEMAIL.COM.'', ''10 ASPMX3.GOOGLEMAIL.COM.'']> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=7
```

Thanks a lot!

ross:
Interesting. The right solution here depends on whether this is specific to GCP or could happen for other providers. In general the `Record` classes should probably normalized their values with `.lower()` (except for text style records.) A quick look and you example above seems to indicate that's not the case. `Record.name` is `.lower()`ed https://github.com/github/octodns/blob/7bf4914a41042cd283cf8cecdec88a9e37f2dbb1/octodns/record/__init__.py#L133 and it looks like `SRV` values are https://github.com/github/octodns/blob/7bf4914a41042cd283cf8cecdec88a9e37f2dbb1/octodns/record/__init__.py#L1156, but others aren't. 

The right fix here is to audit all the values and make sure they `.lower()` if they should. Shouldn't be hard too hard to do.

PeterDaveHello:
I'd like to try sending a pull request :)

PeterDaveHello:
PR sent here: #337 🎉

PeterDaveHello:
Hi @ross, I just found a similar problem with CNAME record, hope you don't mind that I directly send a PR: #346, let me know if it's good enough please, thank you!

PR #346, opened 2019-04-17
Normalize CNAME/ALIAS/PTR value to lower-case

PeterDaveHello:
Related to #337

Currently, CNAME value is not normalized, but might but stored by provider in lower case, which will cause just almost the same issue as #322(one is MX but one is CNAME): found diff and update every sync, and it could be solved by normalization CNAME value.

I also add a small test, just like #337, to make sure the behavior won't appear in the future become a regression bug. Without the normalization, tests will fail as below:

```
..............................................................................................................................................................................................................................F........................................................
======================================================================
FAIL: test_cname_lowering_value (test_octodns_record.TestRecord)
----------------------------------------------------------------------
  File "/home/pete/octodns/tests/test_octodns_record.py", line 42, in test_cname_lowering_value
    self.assertEquals(upper_record.value, lower_record.value)
AssertionError: u'GITHUB.COM' != u'github.com'
- GITHUB.COM
+ github.com

-------------------- >> begin captured logging << --------------------
Record: DEBUG: __init__: zone.name=unit.tests., type=CnameRecord, name=CnameUppwerValue
Record: DEBUG: __init__: zone.name=unit.tests., type=CnameRecord, name=CnameLowerValue
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 279 tests in 4.730s

FAILED (failures=1)

```

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -714,6 +714,8 @@ class _TargetValue(object):
 
     @classmethod
     def process(self, value):
+        if value:
+            return value.lower()
         return value
 
 
--- a/tests/test_octodns_record.py
+++ b/tests/test_octodns_record.py
@@ -8,8 +8,8 @@ from __future__ import absolute_import, division, print_function, \
 from unittest import TestCase
 
 from octodns.record import ARecord, AaaaRecord, AliasRecord, CaaRecord, \
-    CnameRecord, Create, Delete, GeoValue, MxRecord, NaptrRecord, \
-    NaptrValue, NsRecord, Record, SshfpRecord, SpfRecord, SrvRecord, \
+    CnameRecord, Create, Delete, GeoValue, MxRecord, NaptrRecord, NaptrValue, \
+    NsRecord, PtrRecord, Record, SshfpRecord, SpfRecord, SrvRecord, \
     TxtRecord, Update, ValidationError, _Dynamic, _DynamicPool, _DynamicRule
 from octodns.zone import Zone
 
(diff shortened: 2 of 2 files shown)


octodns/record/__init__.py
```diff
@@ -718,7 +718,10 @@ def process(self, value):
 
 
 class CnameValue(_TargetValue):
-    pass
+
+    @classmethod
+    def process(self, value):
+        return value.lower()
```
ross: I believe the `.lower()` will need to be up line ~717 in the parent class so that it'll fix ALIAS records as well.

PeterDaveHello: I'm not sure if you prefer to further revising other parts as simply mv it to `class _TargetValue(object)` will have some other issues:

```
======================================================================
ERROR: test_value_mixin_data (test_octodns_record.TestRecord)
----------------------------------------------------------------------
  File "/home/pete/octodns/tests/test_octodns_record.py", line 152, in test_value_mixin_data
    'value': None
  File "/home/pete/octodns/octodns/record/__init__.py", line 365, in __init__
    self.value = self._value_type.process(data['value'])
  File "/home/pete/octodns/octodns/record/__init__.py", line 717, in process
    return value.lower()
AttributeError: 'NoneType' object has no attribute 'lower'
-------------------- >> begin captured logging << --------------------
Record: DEBUG: __init__: zone.name=unit.tests., type=AliasRecord, name=
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
```

PeterDaveHello: Or I can do a string detection in `class _TargetValue(object)` and decide to lower or not, what do you think?

ross: Yeah. I believe that's a case of where Alias is being used to test _TargetValue. Both CNAME and ALIAS could have a None value, just ALIAS has a test.

It should be fine to do an 

```python
if value:
   return value.lower()
```

There's no need to check for a string in that specific case as that'd go beyond the types of checking that octoDNS currently does (everywhere.)

PeterDaveHello: Fair enough, I also noticed PTR value also uses `_TargetValue`, added it here together(hope I'm not wrong 😓 )

PeterDaveHello: All updated and passed tests now 🎉

PR #349, opened 2019-04-20
Update link to record objects

solvaholic:
#303 renamed `record.py` to `record/__init__.py`.

This updates `docs/records.md` to link to the new path.

### DIFF
--- a/docs/records.md
+++ b/docs/records.md
@@ -174,6 +174,6 @@ In the above example each name had a single record, but there are cases where a 
 
 ### Record data
 
-Each record type has a corresponding set of required data. The easiest way to determine what's required is probably to look at the record object in [`octodns/record.py`](/octodns/record.py). You may also utilize `octodns-validate` which will throw errors about what's missing when run.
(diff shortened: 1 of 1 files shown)


solvaholic:
Am I holding this right?

> The easiest way to determine what's required is probably to look at the record object in [`octodns/record/__init__.py`](/github/octodns/blob/master/octodns/record/__init__.py).

Under [`SshfpValue(object)`](https://github.com/github/octodns/blob/a0bd756c09d686ed390f16bd19d977b64114bc77/octodns/record/__init__.py#L998) I find an `__init__()` definition that seems to spell out what's required for an SSHFP record:

https://github.com/github/octodns/blob/a0bd756c09d686ed390f16bd19d977b64114bc77/octodns/record/__init__.py#L1036-L1039

And under [ `CnameValue(_TargetValue)`](https://github.com/github/octodns/blob/a0bd756c09d686ed390f16bd19d977b64114bc77/octodns/record/__init__.py#L720), for example, I find no such reference:

https://github.com/github/octodns/blob/a0bd756c09d686ed390f16bd19d977b64114bc77/octodns/record/__init__.py#L720-L721

PR #353, opened 2019-04-27
Add OverridingYamlProvider and tests

ross:


dchauran:
Hi Ross - Any reason this has not been integrated? I imagine it might need a bit of py3 work, but this is a feature that my team, and some others could make use of.

ross:
> Hi Ross - Any reason this has not been integrated? I imagine it might need a bit of py3 work, but this is a feature that my team, and some others could make use of.

Just threw it together as a POC to make sure it made sense/was possible. I haven't had an immediate use for it so haven't bothered to get it over the line and ship it. I can throw it on to the TODO list, but it may take me a bit to get there.

Arunothia:
Hi Ross, even our team is looking for this feature. Do you think it would be possible for a few of us to go ahead and complete this? Is there something we should know?

ross:
Just quickly resolved the conflicts and merged master in. Looks like CI is happy with it. That said quickly looking it over I don't think I like the way I did this. I think I'd prefer to just have an optional param to the config/constructor that allows `replace` to be set, defaulting to `False`. 

I'll try to get that worked up and pushed here while I'm waiting on other things this afternoon. I'll need to rely on your all to test it out and :+1: before shipping since I don't have a place to use it meaningfully.

ross:
Pushed up a reworked version. Rather than a completely seperate class that overrides and duplicates some of the code it just adds a new param `populate_should_replace` to the YamlProvider config. That should allow multiple providers to be in the `sources` and override previous values so long as that value is set on them. 

Please test it out/read things over, review the PR, and report back. Once that's done I'll merge.

Arunothia:
Hi Ross, I quickly tested this code by only using YamlProvider for sources and targets. I get **octodns.zone.DuplicateRecordException** when I add the same record in the second source  whereas I would expect it to override in place. Adding a new record seems to work fine. 

The following are the two configs I have that corresponds to your external.yaml and internal.yaml.

---
providers:
  config:
    directory: config
  dest1:
    directory: dest1

zones:
  test.com.:
    sources:
      - config
    targets:
      - dest1

---
providers:
  config:
    directory: config
  dest2:
    directory: dest2

zones:
  test.com.:
    sources:
      - config
      - override
    targets:
      - dest2

---

The following is the test.com zone file in config folder- 

? ''
: 
  - ttl: 3600
    type: A
    value: 18.0.0.108

cn:
- ttl: 3600
  type: CNAME
  value: cn.test.com.

mmm:
- ttl: 3600
  type: CNAME
  value: mmm.test.com.

www:
- ttl: 3600
  type: CNAME
  value: any.test.com.

---

And the following is the same zone file in override folder- 

adding:
- ttl: 3600
  type: CNAME
  value: any.test.com.
  
mmm:
- ttl: 3600
  type: CNAME
  value: arunothia.test.com.

---
I get **octodns.zone.DuplicateRecordException: Duplicate record mmm.test.com., type CNAME**

ross:
Realized the example omitted the `populate_should_replace: true` bit from the `internal` config. Added that in https://github.com/github/octodns/pull/353/commits/f1cc392bc41bddd24f480a1178ccd50ab15019b5.

Arunothia:
Will it be an easy change to merge external and internal yamls into one single config? Maybe we can specify the source list per target or something?

Arunothia:
Our team did local testing as well as a preliminary run through on our mock pipeline, it seems to work as we desire!

ross:
> Our team did local testing as well as a preliminary run through on our mock pipeline, it seems to work as we desire!

:cool:. I've tested things a bit to make sure that this doesn't badly break existing uses so we'll go ahead and move forward with it...

PR #354, opened 2019-04-29
Ignore VSCode settings changes.

cfunkhouser:
I've been ignoring the local changes, but it's only a matter of time before I forget to and ship my settings in a PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
+.vscode/*


ross:
Hi @cfunkhouser as a general rule I don't include editor-specific stuff in the repos since everyone has their own editor/details...

The best way to solve this problem globally (and the way I do so for vim) is to include a global ignores file on my boxes that get applied on top of the repo specific stuff:

```
$ cat ~/.gitconfig
...
[core]
  excludesfile = /Users/ross/.gitignores
```

```
$ cat ~/.gitignores
*.swo
*.swp
.DS_Store
tmp/
```

You could do something similar for the `.vscode/` directory.

Issue #341, opened 2019-04-11
Add support for Mythic Beasts


adhawkins:
Mythic Beasts (https://www.mythic-beasts.com/) offer DNS hosting, and provide an API to update these DNS entries (https://www.mythic-beasts.com/support/api/primary)

Would be nice to have a provider for them to allow DNS syncing using octodns

ross:
https://github.com/ross/octodns-ddns is an example of a 3rd party provider, though different in focus. That seems like it might be a good fit for this provider and would decouple updates and releases from the main octoDNS (eventually all providers will likely move outside the core code.) I won't likely have time to look into this myself, but I'd be happy to take a look at code that does it.

PR #360, opened 2019-05-09
Support for Mythic Beasts DNS API

cwningen:
Resolves: #341 Working initial pull for Mythic API

### DIFF
--- a/octodns/provider/mythicbeasts.py
+++ b/octodns/provider/mythicbeasts.py
@@ -5,26 +5,31 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
+import re
+
 from requests import HTTPError, Session
 from logging import getLogger
 
-from ..record import Create, Record
+from ..record import Record
 from .base import BaseProvider
 
-import re
 
-import sys
+def add_trailing_dot(value):
+    '''
+    Add trailing dots to values
+    '''
+    assert value
+    assert value[-1] != '.'
+    return value + '.'
 
-def add_trailing_dot(s):
-    assert s
-    assert s[-1] != '.'
-    return s + '.'
 
-
-def remove_trailing_dot(s):
-    assert s
-    assert s[-1] == '.'
-    return s[:-1]
+def remove_trailing_dot(value):
+    '''
+    Remove trailing dots from values
+    '''
+    assert value
+    assert value[-1] == '.'
+    return value[:-1]
 
 
 class MythicBeastsProvider(BaseProvider):
@@ -35,7 +40,6 @@ class MythicBeastsProvider(BaseProvider):
       class: octodns.provider.mythicbeasts.MythicBeastsProvider
         zones:
           my-zone: 'password'
-
     '''
 
     SUPPORTS_GEO = False
@@ -45,10 +49,13 @@ class MythicBeastsProvider(BaseProvider):
     BASE = 'https://dnsapi.mythic-beasts.com/'
     TIMEOUT = 15
 
-    def __init__(self, id, passwords, *args, **kwargs):
-        self.log = getLogger('MythicBeastsProvider[{}]'.format(id))
-        self.log.debug('__init__: id=%s, registered zones; %s', id, passwords.keys())
-        super(MythicBeastsProvider, self).__init__(id, *args, **kwargs)
+    def __init__(self, identifier, passwords, *args, **kwargs):
+        self.log = getLogger('MythicBeastsProvider[{}]'.format(identifier))
+        self.log.debug(
+            '__init__: id=%s, registered zones; %s',
+            identifier,
+            passwords.keys())
+        super(MythicBeastsProvider, self).__init__(identifier, *args, **kwargs)
 
         self._passwords = passwords
         sess = Session()
@@ -59,7 +66,10 @@ class MythicBeastsProvider(BaseProvider):
 
         url = self.BASE
         resp = self._sess.request(method, url, data=data, timeout=self.TIMEOUT)
-        self.log.debug('_request:   status=%d data=%s', resp.status_code, resp.text)
+        self.log.debug(
+            '_request:   status=%d data=%s',
+            resp.status_code,
+            resp.text)
         if resp.status_code != 200:
             self.log.info('request failed: %s, response %s', data, resp.text)
             resp.raise_for_status()
@@ -76,26 +86,32 @@ class MythicBeastsProvider(BaseProvider):
             'command': 'LIST',
         })
 
-    def _data_for_single(self, _type, data):
+    @staticmethod
+    def _data_for_single(_type, data):
         return {
             'type': _type,
             'value': data['raw_values'][0]['value'],
             'ttl': data['raw_values'][0]['ttl']
         }
 
-    def _data_for_multiple(self, _type, data):
+    @staticmethod
+    def _data_for_multiple(_type, data):
         return {
             'type': _type,
-            'values': [raw_values['value'] for raw_values in data['raw_values']],
-            'ttl': max([raw_values['ttl'] for raw_values in data['raw_values']]),
+            'values':
+                [raw_values['value'] for raw_values in data['raw_values']],
+            'ttl':
+                max([raw_values['ttl'] for raw_values in data['raw_values']]),
         }
 
-    def _data_for_MX(self, _type, data):
+    @staticmethod
+    def _data_for_MX(_type, data):
         ttl = max([raw_values['ttl'] for raw_values in data['raw_values']])
         values = []
 
-        for raw_value in [raw_values['value'] for raw_values in data['raw_values']]:
-            match = re.match('^([0-9]+)\s+(\S+)$', raw_value, re.IGNORECASE)
+        for raw_value in \
+            [raw_values['value'] for raw_values in data['raw_values']]:
+            match = re.match('^([0-9]+)\\s+(\\S+)$', raw_value, re.IGNORECASE)
 
             if match is not None:
                 exchange = match.group(2)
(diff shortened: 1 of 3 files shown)


cwningen:
Will lint it.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+class MythicBeastsProvider(BaseProvider):
+    '''
+    Mythic Beasts DNS API Provider
+
+    mythicbeasts:
+      class: octodns.provider.mythicbeasts.MythicBeastsProvider
+        zones:
+          my-zone: 'password'
```
ross: The code below makes it look like `passwords` is required and has to be a `dict`, but this has `zones`.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+    SUPPORTS_GEO = False
+    SUPPORTS_DYNAMIC = False
+    SUPPORTS = set(('A', 'AAAA', 'ALIAS', 'CNAME', 'MX', 'NS',
+                    'SRV', 'SSHFP', 'CAA', 'TXT'))
+    BASE = 'https://dnsapi.mythic-beasts.com/'
+
+    def __init__(self, identifier, passwords, *args, **kwargs):
+        self.log = getLogger('MythicBeastsProvider[{}]'.format(identifier))
+
+        assert isinstance(passwords, dict), 'Missing passwords'
```
ross: Probably best to be a little more specific in what's happening here. If someone did:

```yaml
  passwords: foobarbaz
```

this would tell them that they're missing passwords when the problem is really that passwords must be a dictionary. The `passwords` param up above should serve as the "requires" part. This should probably tell them that passwords must be a dict.

```suggestion
        assert isinstance(passwords, dict), 'passwords must be a dictionary'
```

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+
+        resp = self._sess.request(method, path, data=data)
+        self.log.debug(
+            '_request:   status=%d data=%s',
+            resp.status_code,
+            resp.text[:20])
+
+        if resp.status_code == 401:
+            raise Exception('Mythic Beasts unauthorized for domain: {}'
+                            .format(data['domain']))
```
ross: I generally prefer things to throw a provider specific exception in cases like this.

```python
class MythicBeastsUnauthorizedException(Exception):
    pass

...

    raise MythicBeastsUnauthorizedException(...)
```

Its doubtful anyone will ever try to catch it, but it gives a bit more info when it fires and it usually makes it easier to ensure that the right thing went wrong in tests.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+            .format(remove_trailing_dot(zone))
+
+        return self._post({
+            'domain': remove_trailing_dot(zone),
+            'password': self._passwords[zone],
+            'showall': 0,
+            'command': 'LIST',
+        })
+
+    @staticmethod
```
ross: Any reason in particular these are `@staticmethod`? Obviously works fine, but it doesn't line up with the way all of the other providers have thing laid out.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+        }
+
+    @staticmethod
+    def _data_for_MX(_type, data):
+        ttl = max([raw_values['ttl'] for raw_values in data['raw_values']])
+        values = []
+
+        for raw_value in \
+                [raw_values['value'] for raw_values in data['raw_values']]:
+            match = re.match('^([0-9]+)\\s+(\\S+)$', raw_value, re.IGNORECASE)
```
ross: Would prefer that this use a pre-compiled re rather than match which builds one each route. Pedantic and this isn't going to be a hot path or anything, but making it a class-static property, `MX_RE` etc., would compile it once and be more in line with other providers.

It's also best to use `r'...'` for regexes.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+
+            assert match is not None, 'Unable to parse MX data'
```
ross: Not 100% sure, but `assert`ing here may get in the way of the `lenient` flag working as designed for this provider. In general that does a best effort building of records so that a baseline set of data can be pulled out of a provider when things aren't 100% right. When you go to sync the resulting yaml things will complain and requiring fixing then (unless you further set `lenient`.)

The functionality is intended to allow users to pull less than ideal data out of a provider, clean it up locally and then push back (and/or to somewhere else) better data.

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+        before = len(zone.records)
+        exists = False
+        data = dict()
+
+        exists = True
+        for line in resp.content.splitlines():
+            match = re.match(
+                '^(\\S+)\\s+(\\d+)\\s+(\\S+)\\s+(.*)$',
+                line,
+                re.IGNORECASE)
```
ross: It is pretty hard to decipher which group is what here. Probably worth using named groups. `r'(?P<name>\S+...`
& `name = match.group('name')` sort of thing.  https://docs.python.org/2/library/re.html#re.MatchObject.group

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+
+                if _type not in data:
+                    data[_type] = dict()
+
+                if _name not in data[_type]:
+                    data[_type][_name] = {
+                        'raw_values': [{'value': _value, 'ttl': _ttl}],
+                        'name': _name,
+                        'zone': zone.name,
+                    }
```
ross: data up above can use default dict to avoid the `if not in` stuff down here. It's a bit of a trade off between the complicated defaultdict bits up above and making this code down here dirt simple though.

```python
from collections import defaultdict
...
data = defaultdict(lambda: defaultdict(lambda: {
    'raw_values': [],
    'name': None,
    'zone': None,
}))
...
record = data['A']['foo']
record['name'] = _name
record['zone'] = zone.name
record['raw_values'].append({
    'value': _value,
    'ttl': _ttl,
})
```

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def _compile_commands(self, action, change):
+        commands = []
+
+        record = None
+
+        if action == 'ADD':
+            record = change.new
+        else:
+            record = change.existing
```
ross: There's a property on `Change` to do this https://github.com/github/octodns/blob/aeeec85ab61e6ac38ccd41768a8ca247230e3787/octodns/record/__init__.py#L21-L24

octodns/provider/mythicbeasts.py
```diff
@@ -0,0 +1,420 @@
...
+            _type = 'ANAME'
+
+        if hasattr(record, 'values'):
+            values = record.values
+        else:
+            values = [record.value]
+
+        base = '{} {} {} {}'.format(action, hostname, ttl, _type)
+
+        if re.match('[A]{1,4}', _type) is not None:
```
ross: If i understand this I think it'd be clearer as:

```suggestion
        if _type in ('A', 'AAAA'):
```

octodns/provider/mythicbeasts.py
cwningen: Originally it wasn't but I may have been a bit over zealous with pylint at some point as that's the only reason that comes to mind (given it has no need for self and should be static but is also fine to be nonstatic.

octodns/provider/mythicbeasts.py
cwningen: Will do, and update tests

octodns/provider/mythicbeasts.py
cwningen: Okay, I've changed that although I'm pretty sure python compiles and caches re.match etc even if you use it multiple times. I recall doing a test in 2.7 many months ago wondering if there was a difference but couldn't see any between re.compile and loop, vs, loop with re.match in the middle

octodns/provider/mythicbeasts.py
cwningen: Sure will do.

octodns/provider/mythicbeasts.py
cwningen: I'm not sure I follow this at all or the improvement?

ross: It's different more than a specific improvement, that'd be opinion. Personally I like that the loop doesn't have any of the `if` `else` default handling and just focuses on the logic of adding the records. The `defaultdict` stuff up top is a little bit complicated since it's dealing with two-levels, in exchange for the code in the loop just setting things directly.

cwningen: Okay no problem. I think the most recent commit should clear that up.

ross:
Looks like I missed that this PR had been further updated since the last round of discussion. From you all's POV is is ready for a final :eye: and merge?

adhawkins:
I've been using this provider for a while to synchronise records in TinyDNS form to the Mythic Beasts API. I've liased with the developer after finding a few issues, all of which she's fixed.

cwningen:
Not sure if the question is directed also at me, I think from discussions and testing back and forth with @adhawkins that I'm happy for this to be merged if so.

ross:
> Not sure if the question is directed also at me, I think from discussions and testing back and forth with @adhawkins that I'm happy for this to be merged if so.

Yeah. Mostly just wanted to check that you are ready for it to move forward before doing so. Put an item on my todo list to get this in.

PR #364, opened 2019-05-15
Prevent digital ocean provider to crash if records type is not supported

### LINKED ISSUE #363: Error with SOA type for Digital Ocean provider
When I sync Digital Ocean DNS I get this error

```
  File "/Users/alex/.local/share/virtualenvs/dns-m997HFnd/bin/octodns-dump", line 10, in <module>
    sys.exit(main())
  File "/Users/alex/.local/share/virtualenvs/dns-m997HFnd/lib/python2.7/site-packages/octodns/cmds/dump.py", line 35, in main
    *args.source)
  File "/Users/alex/.local/share/virtualenvs/dns-m997HFnd/lib/python2.7/site-packages/octodns/manager.py", line 382, in dump
    source.populate(zone, lenient=lenient)
  File "/Users/alex/.local/share/virtualenvs/dns-m997HFnd/lib/python2.7/site-packages/octodns/provider/digitalocean.py", line 231, in populate
    data_for = getattr(self, '_data_for_{}'.format(_type))
AttributeError: 'DigitalOceanProvider' object has no attribute '_data_for_SOA'
```

After seeing code I see that it doesn't support `SOA` type but instead of just ignore it, code crash, how can we patch it ?

 Adding a check to know if attribute exist [here](https://github.com/github/octodns/blob/master/octodns/provider/digitalocean.py#L231) seems to be a good solution is it okay for you ?

Its-Alex:
This patch will prevent Digital Ocean provider to crash if records type is not supported

Fix #363

### DIFF
--- a/octodns/provider/digitalocean.py
+++ b/octodns/provider/digitalocean.py
@@ -223,6 +223,10 @@ class DigitalOceanProvider(BaseProvider):
         values = defaultdict(lambda: defaultdict(list))
         for record in self.zone_records(zone):
             _type = record['type']
+            if _type not in self.SUPPORTS:
+                self.log.warning('populate: skipping unsupported %s record',
+                                 _type)
+                continue
             values[record['name']][record['type']].append(record)
 
         before = len(zone.records)
--- a/tests/fixtures/digitalocean-page-1.json
+++ b/tests/fixtures/digitalocean-page-1.json
@@ -1,5 +1,16 @@
 {
 	"domain_records": [{
+		"id": null,
+		"type": "SOA",
+		"name": "@",
+		"data": null,
+		"priority": null,
+		"port": null,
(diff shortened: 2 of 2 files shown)


yzguy:
Looks good, this may need to go up to Line 226 to be consistent with some of the other providers.

[DNSMadeEasy](https://github.com/github/octodns/blob/master/octodns/provider/dnsmadeeasy.py#L266-L272)
[DNSSimple](https://github.com/github/octodns/blob/master/octodns/provider/dnsimple.py#L265-L273)

Just need to get the tests to pass and should be good! Thanks for the contribution

Its-Alex:
Okay I moved the line, add fixtures for this case, and fix CI

ross:
Thanks.

PR #379, opened 2019-07-18
Adding support for Akamai's fastdns

basirjamil:


octodns/provider/fastdns.py
```diff
@@ -0,0 +1,540 @@
...
+        """ returns records for a zone, looks for it if not present, or
+            returns empty [] if can't find a match
+        """
+        if zone.name not in self._zone_records:
+            try:
+                name = zone.name[:-1]
+                response = self._dns_client.zone_recordset_get(name)
+                self._zone_records[zone.name] = response.json()["recordsets"]
+
+            except (AkamaiClientException, KeyError):
```
ross: What case is the `AkamaiClientException` trying to catch here? I suspect that actual errors with the service (5xx etc.) or bad data returned (missing `recordsets`) would result in a blank `zone_records` here when it probably just wants to catch the no found case?

octodns/provider/fastdns.py
```diff
@@ -0,0 +1,540 @@
...
+    def _apply(self, plan):
+        desired = plan.desired
+        changes = plan.changes
+        self.log.debug('apply: zone=%s, chnges=%d', desired.name, len(changes))
+
+        zone_name = desired.name[:-1]
+        try:
+            self._dns_client.zone_get(zone_name)
+
+        except AkamaiClientException:
```
ross: Same thing here as above. I believe this would treat all errors as not found and try and create a zone that may already exist when the get fails for other reasons.

octodns/provider/fastdns.py
```diff
@@ -0,0 +1,540 @@
...
+
+    Fast DNS Zone Management API V2, found here:
+    developer.akamai.com/api/web_performance/fast_dns_zone_management/v2.html
+
+    Info on Python Requests library:
+    https://2.python-requests.org/en/master/
+
+    '''
+
+    def __init__(self, _client_secret, _host, _access_token, _client_token):
```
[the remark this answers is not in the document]
basirjamil: it was to differentiate from the same names that are used as parameters by EdgeGridAuth a few lines down:

```
        sess.auth = EdgeGridAuth(
            client_token=_client_token,
            client_secret=_client_secret,
            access_token=_access_token
        )
```

basirjamil:
@ross made some changes to the exception. Looking back, throughout the code / tests, we only care for the case where a zone is not found (404), so I got rid of the other ones and let r.raise_for_status() handle those

octodns/provider/fastdns.py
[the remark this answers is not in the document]
ross: OK. Fwiw it should be fine to have parameter names that mirror the keys in the `EdgeGridAuth`. I'd lean towards removing the `_` as that's kind of exposing a detail of the code of the method to callers, e.g. if someone ever wanted/needed to pass in named parameters to the client object they'd have to add the `_` on. Anyway, not a blocking thing, just opinion.

octodns/provider/fastdns.py
```diff
@@ -0,0 +1,533 @@
...
+        )
+        self._sess = sess
+
+    def _request(self, method, path, params=None, data=None, v1=False):
+
+        url = urljoin(self.base, path)
+        resp = self._sess.request(method, url, params=params, json=data)
+
+        if resp.status_code > 299:
+            raise AkamaiClientException(resp)
```
ross: I believe the way this works now (Having AkamaiClientException's __init__ do the raise_for_status) will result in slightly confusing stack traces. It's probably better to do something like:

```python
if resp.status_code == 404:
    raise AkamaiClientException()
resp.raise_for_status()
```

So that the exception is thrown in the case where it makes sense and the normal requests process happens otherwise. It might make sense to rename `AkamaiClientException` to `AkamaiNotFound` so something like that so it's clear what that exception means.

basirjamil:
@ross made the suggested changes

PR #383, opened 2019-07-25
Add Selectel as OctoDNS provider

TruePack:
Completed https://github.com/github/octodns/pull/116

### DIFF
--- a/octodns/provider/selectel.py
+++ b/octodns/provider/selectel.py
@@ -28,7 +28,7 @@ class SelectelProvider(BaseProvider):
 
     MIN_TTL = 60
 
-    LIMIT = 50
+    PAGINATION_LIMIT = 50
 
     API_URL = 'https://api.selectel.ru/domains/v1'
 
@@ -69,9 +69,9 @@ class SelectelProvider(BaseProvider):
 
     def _request_with_pagination(self, path, total_count):
         result = []
-        for offset in range(0, total_count, self.LIMIT):
+        for offset in range(0, total_count, self.PAGINATION_LIMIT):
             result += self._request('GET', path,
-                                    params={'limit': self.LIMIT,
+                                    params={'limit': self.PAGINATION_LIMIT,
                                             'offset': offset})
         return result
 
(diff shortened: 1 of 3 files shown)


octodns/provider/selectel.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+class SelectelProvider(BaseProvider):
+    SUPPORTS_GEO = False
+
+    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'MX', 'NS', 'TXT', 'SPF', 'SRV'))
+
+    MIN_TTL = 60
+
+    LIMIT = 50
```
ross: A nit, but prefer being a little more explicit here and calling this a page size to make it clear what it applies to.

```suggestion
    PAGINATION_LIMIT = 50
```

octodns/provider/selectel.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def _request(self, method, path, params=None, data=None):
+        self.log.debug('_request: method=%s, path=%s', method, path)
+
+        url = '{}{}'.format(self.API_URL, path)
+        resp = self._sess.request(method, url, params=params, json=data)
+
+        self.log.debug('_request: status=%s', resp.status_code)
+        if resp.status_code == 401:
+            raise SelectelAuthenticationRequired(resp.text)
```
ross: Would this happen when the token is invalid?

octodns/provider/selectel.py
```diff
@@ -0,0 +1,303 @@
...
+                                    params={'limit': self.LIMIT,
+                                            'offset': offset})
+        return result
+
+    def _include_change(self, change):
+        if isinstance(change, Update):
+            existing = change.existing.data
+            new = change.new.data
+            new['ttl'] = max(self.MIN_TTL, new['ttl'])
+            if new == existing:
```
ross: Probably good to include a `self.log.debug` here so that if somebody runs into something that doesn't make sense to them (why is my TTL not working) they can do a `--debug` and get more info about what's happening.

octodns/provider/selectel.py
TruePack: @ross yes

PR #372, opened 2019-07-11
Travis CI: Run tests on Python 3.7 in allow_failures mode

cclauss:

PR #384, opened 2019-07-29
Start on supporting python3 using backwards compat bits for now

ross:
This isn't complete/ready yet, but it gets things pretty far alone and I wanted to push something up to get it off my laptop. It builds on top of https://github.com/github/octodns/pull/373.

python2 is still 100% happy, covered, etc with what's here.

python3 continues to have test failures and other problems. Some of those issues are with libraries we depend on not supporting python3 e.g. azuredns, incf.countryutils, etc. Others are further updates that need to happen to code that lives in octoDNS itself (a lot of times test code.)


octodns/compat.py
```diff
@@ -0,0 +1,10 @@
+#
+# Python 2/3 compat bits
+#
+
+try:  # pragma: no cover
+    from StringIO import StringIO
+except ImportError:  # pragma: no cover
+    from io import StringIO
```
cclauss: Just use __six.StringIO__.

octodns/provider/plan.py
```diff
@@ -5,10 +5,12 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
-from StringIO import StringIO
 from logging import DEBUG, ERROR, INFO, WARN, getLogger
 from sys import stdout
 
+from six import text_type
+from ..compat import StringIO
```
cclauss: __from six import StringIO__

tests/test_octodns_plan.py
```diff
@@ -5,10 +5,11 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
-from StringIO import StringIO
 from logging import getLogger
+from six import text_type
```
cclauss: from six import StringIO

tests/test_octodns_provider_rackspace.py
```diff
@@ -7,8 +7,12 @@
 
 import json
 import re
+from six import text_type
 from unittest import TestCase
-from urlparse import urlparse
+try:
+    from urlparse import urlparse
+except ImportError:
+    from urllib.parse import urlparse
```
cclauss: __from six.move.urllib.parse import urlparse__

tests/test_octodns_yaml.py
```diff
@@ -5,10 +5,10 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
-from StringIO import StringIO
 from unittest import TestCase
 from yaml.constructor import ConstructorError
 
+from octodns.compat import StringIO
```
cclauss: six.StringIO

nefaspartim:
Hey guys,

Just wondering if this work is still ongoing or if I could throw my hat into the ring to help finish it up?

.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: python
-python:
-  - 2.7
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.7
+      dist: xenial  # required for Python >= 3.7 on Travis CI
```
cclauss: Line 6 can be removed because xenial is the default.

cclauss:
I hate these _big bang_ ports because they always stall and then fail.  I have only seen success on larger ports through a series of smaller PRs.  Each one is tested, reviewed, and merged on its own.  Each one must continue to support legacy Python without breakage.  Once the series is all merged, A/B testing can be done to ensure that Py2 and Py3 are both fully supported by a single codebase and then legacy Python support can be dropped.  This is in alignment with the Python porting guidelines.

nefaspartim:
> I hate these _big bang_ ports because they always stall and then fail. I have only seen success on larger ports through a series of smaller PRs. Each one is tested, reviewed, and merged on its own. Each one must continue to support legacy Python without breakage. Once the series is all merged, A/B testing can be done to ensure that Py2 and Py3 are both fully supported by a single codebase and then legacy Python support can be dropped. This is in alignment with the Python porting guidelines.

Yeah, I've noticed that problem with a few other projects I follow, and in a lot of cases it's just using the functionality `six` provides and working off of that (or just using Modernize straight out).

I've got some extra time to run through these and submit smaller PRs (taking the reviews into account of course). Python3 support for octo is huge on my list and anything I can do to give back to the project I'm happy to do.

cclauss:
Can we _please_ start by reviewing and landing #372?  It will be super helpful for tracking the rest of the work.  Let’s all agree on the use of six.  I tend to use Python-future instead of modernize but I have used both and they are effective.

octodns/manager.py
```diff
@@ -275,7 +275,10 @@ def sync(self, eligible_zones=[], eligible_targets=[], dry_run=True,
             self.log.info('sync:   sources=%s -> targets=%s', sources, targets)
 
             try:
-                sources = [self.providers[source] for source in sources]
+                collected = []
+                for source in sources:
+                    collected.append(self.providers[source])
+                sources = collected
```
cclauss: Why?

octodns/provider/base.py
```diff
@@ -58,7 +60,7 @@ def plan(self, desired):
 
         # allow the provider to filter out false positives
         before = len(changes)
-        changes = filter(self._include_change, changes)
+        changes = list(filter(self._include_change, changes))
```
cclauss: Let’s make this a __list comprehension__ as discussed at https://docs.python.org/3/library/functions.html#filter

octodns/provider/ovh.py
```diff
@@ -325,7 +326,8 @@ def _is_valid_dkim(self, value):
         splitted = value.split('\\;')
         found_key = False
         for splitted_value in splitted:
-            sub_split = map(lambda x: x.strip(), splitted_value.split("=", 1))
+            sub_split = list(map(lambda x: x.strip(),
+                                 splitted_value.split("=", 1)))
```
cclauss: Let’s make this a __list comprehension__ as discussed at https://docs.python.org/3/library/functions.html#map

nefaspartim:
@cclauss - What is left on #372 ? I noticed that those changes are also rolled into this ticket (and briefly touched upon in #373). What's the next course of action? Is it something for @ross to review?

ross:
> I hate these _big bang_ ports because they always stall and then fail.

This hasn't stalled, it's actually most of the way there. There's only a handful of tests still failing. 90% of the time has been dealing with annoying iteration and object comparison changes between py2 and py3. Most of what's left is more of that with the exception of the `countryutils` package being abandonware and not supporting python3. 

This hasn't moved b/c I haven't done any optional work on octoDNS since I was last working on this. I've been focusing elsewhere based on priorities. As nice as py3 support would be nothing hard requires it yet and there's still a good amount of time before py2's deadline. 

You've stated your opinion on the approach I prefer several times now and I've heard it and it's not how I wanted to go about it. As is obvious from your tone you still don't like it and that's fine. Please refrain from continuing to say so. You're welcome to continue discussing and/suggesting specific changes provided that's done in a constructive manner.

octodns/compat.py
ross: Makes sense. I think this came out of a porting doc I was originally looking at before I moved towards using six.

tests/test_octodns_plan.py
ross: This one isn't pulling in StringIO

octodns/manager.py
ross: Don't remember off-hand. Don't think I would have changed it unless it was fixing a test failure.

octodns/provider/base.py
ross: I'm not sure what you mean.

ross:
Looks like the azure provider does support python3, there are some details of objects that haven't been fully updated that matter with the way that the octoDNS provider is using them. E.g. RecordSet is not hashable in python3 b/c it defines an `__eq__`, but no `__hash__`. Filed https://github.com/Azure/msrest-for-python/pull/182

ross:
Python 3 now passes all tests & coverage if people want to start 👞 the tires.

octodns/manager.py
ross: Figured out what was going on with this, I think. Looks like the `except` block was throwing an exception and causing the test for this bit to fail. It did capture an exception, but not the expected message. It'd probably make sense to throw a `ManagerException`, something specific to avoid this confusion/case so going to look at that change.

octodns/provider/base.py
ross: The above link didn't really point me in the right direction, but https://portingguide.readthedocs.io/en/latest/iterators.html#higher-order-functions-vs-list-comprehensions did.

cclauss:
I repeat what I said [one month ago](https://github.com/github/octodns/pull/384#issuecomment-530151017) that a single pull request that touches 50+ files is the wrong way to port to Python 3.  83 days until Python 2 end of life.

octodns/provider/azuredns.py
```diff
@@ -373,13 +377,13 @@ def populate(self, zone, target=False, lenient=False):
...
 
-        _records = set()
+        _records = []
         records = self._dns_client.record_sets.list_by_dns_zone
         if self._check_zone(zone_name):
             exists = True
             for azrecord in records(self._resource_group, zone_name):
                 if _parse_azure_type(azrecord.type) in self.SUPPORTS:
-                    _records.add(azrecord)
+                    _records.append(azrecord)
```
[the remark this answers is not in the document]
ross: As best as I could tell this wasn't relying on the `set` to unique-ify things. I couldn't think of a reason/why that there would be duplicates in an Azure zone.

/cc @h-hwang who was the original author who could answer definitively.

octodns/provider/route53.py
```diff
@@ -147,25 +148,42 @@ def mod(self, action, existing_rrsets):
...
+            self.fqdn == other.fqdn and \
+            self._type == other._type
+
+    def __ne__(self, other):
+        return self.__class__.__name__ != other.__class__.__name__ or \
+            self.fqdn != other.fqdn or \
+            self._type != other._type
+
+    def __lt__(self, other):
+        return (((self.__class__.__name__, self.fqdn, self._type)) <
```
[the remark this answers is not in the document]
ross: 👍 makes sense. I was on auto-pilot when copy-n-pasting all these.

octodns/provider/route53.py
```diff
@@ -506,11 +524,50 @@ def __hash__(self):
...
                                  self.geo.code).__hash__()
 
-    def __cmp__(self, other):
-        ret = super(_Route53GeoRecord, self).__cmp__(other)
-        if ret != 0:
-            return ret
-        return cmp(self.geo.code, other.geo.code)
+    def __eq__(self, other):
+        return super(_Route53GeoRecord, self).__eq__(other) and \
+            self.geo.code == other.geo.code
```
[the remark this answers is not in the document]
ross: Much cleaner here.

octodns/equality.py
```diff
@@ -0,0 +1,30 @@
...
+        return self._equality_tuple() < other._equality_tuple()
+
+    def __le__(self, other):
+        return self._equality_tuple() <= other._equality_tuple()
+
+    def __gt__(self, other):
+        return self._equality_tuple() > other._equality_tuple()
+
+    def __ge__(self, other):
+        return self._equality_tuple() >= other._equality_tuple()
```
ross: @theojulienne after looking at the equality stuff for a second I realized that I might as well pull this mixin out and DRY things up a bit. Looked around to see if there were any stdlib things like this and didn't find anything. It seems to work/make sense.

cclauss: functools.cmp_to_key and functools.total_ordering?

Tdowneydev (commented):
**Deleted**

PR #385, opened 2019-07-29
Added fastdns as a supported provider in doc

basirjamil:


nefaspartim:
@basirjamil - Quick question, I noticed in the doc it says FastDNS supports ALIAS (which it does in a janky way by way of AKAMAICDN records), but in the module the support is only built in for these types:

    SUPPORTS_GEO = False
    SUPPORTS_DYNAMIC = False
    SUPPORTS = set(('A', 'AAAA', 'CNAME', 'MX', 'NAPTR', 'NS', 'PTR', 'SPF', 'SRV', 'SSHFP', 'TXT'))

I'll poke at making AKAMAICDN be the ALIAS record type for the module, but should we leave it out until it actually supports the ALIAS type?

yzguy:
@nefaspartim A similar thing is done with the Constellix provider converting ANAME to ALIAS records

https://github.com/github/octodns/blob/master/octodns/provider/constellix.py#L118-L120
https://github.com/github/octodns/blob/master/octodns/provider/constellix.py#L141-L143

It's done in the "client" class so that the Provider doesn't have to know about it. You might be able to do something similar in the FastDNS provider.

ross:
Merged https://github.com/github/octodns/pull/404 rather than this one since it didn't include `ALIAS`. Happy to take a look at and review a PR that adds support to the provider and includes it in the list.

PR #388, opened 2019-08-19
fix Route53Provider to tollerate multiple separators in record values

willblackie:
fix split calls in Route53Provider to not specify ' ' parameter so they can tolerate multiple sequential separators


ross:
Fixes https://github.com/github/octodns/issues/389

PR #391, opened 2019-08-26
Quotation marks in TXT records get stripped

sheax0r:
@cfunkhouser I believe this demonstrates the quotation-marks issue we've been seeing with octodns.

My python is not very good, but if you want to pair on understanding and fixing this at some point, I'd be game :).


sheax0r:
This is not rea

PR #402, opened 2019-09-13
Switch from X-Auth-{Email,Key} to Authorization for Cloudflare

mamercad:
https://api.cloudflare.com/#getting-started-requests

> API Tokens provide a new way to authenticate with the Cloudflare API. They allow for scoped and permissioned access to resources and use the RFC compliant Authorization Bearer Token Header.


.gitignore
```diff
@@ -11,3 +11,4 @@ nosetests.xml
 octodns.egg-info/
 output/
 tmp/
+/.vscode/
```
ross: Please remove this change.  octoDNS avoids including editor specific ignores here. I'd recommend looking at a "global" ignores file. 

```
 ross$ grep exclude ~/.gitconfig
  excludesfile = /Users/ross/.gitignores
```

```
(env)otter:datadog-dashboards ross$ cat ~/.gitignores
*.swo
*.swp
.DS_Store
.env/
tmp/
```

That gives you a way to ignore your editor's files without us having to add all of the possible options into octoDNS.

octodns/provider/cloudflare.py
```diff
@@ -66,16 +64,16 @@ class CloudflareProvider(BaseProvider):
     MIN_TTL = 120
     TIMEOUT = 15
 
-    def __init__(self, id, email, token, cdn=False, *args, **kwargs):
+    def __init__(self, id, token, cdn=False, *args, **kwargs):
```
ross: I think we'll want to go about this in a way that doesn't break existing usage of email & token. That'd likely involve making `email` an optional parameter that defaults to `None` and then if'ing some handling details down below.

mamercad: Yeah, I actually had the same thought the other day... I'll fix it up to do both.

.gitignore
mamercad: Good call, thanks.

parkr:
This PR LGTM. I'm going to give it a whirl with a new API token on Jekyll's DNS.

octodns/provider/cloudflare.py
```diff
@@ -66,17 +66,24 @@ class CloudflareProvider(BaseProvider):
...
+        if email and token:
+            sess.headers.update({
+                'X-Auth-Email': email,
+                'X-Auth-Key': token,
+            })
+        else:
+            # https://api.cloudflare.com/#getting-started-requests
+            # https://tools.ietf.org/html/rfc6750#section-2.1
+            sess.headers.update({
+                'Authorization': 'Bearer %s'.format(token),
```
parkr: I received this error when running with `token: env/CLOUDFLARE_TOKEN` and `CLOUDFLARE_TOKEN` set to `Read Zone.Zone` and `Edit Zone.DNS`.

```text
  File "/usr/local/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/octodns/cmds/sync.py", line 39, in main
    dry_run=not args.doit, force=args.force)
  File "/usr/local/lib/python2.7/site-packages/octodns/manager.py", line 301, in sync
    plans = [p for f in futures for p in f.result()]
  File "/usr/local/lib/python2.7/site-packages/octodns/manager.py", line 56, in result
    return self.func(*self.args, **self.kwargs)
  File "/usr/local/lib/python2.7/site-packages/octodns/manager.py", line 237, in _populate_and_plan
    plan = target.plan(zone)
  File "/usr/local/lib/python2.7/site-packages/octodns/provider/base.py", line 49, in plan
    exists = self.populate(existing, target=True, lenient=True)
  File "/usr/local/lib/python2.7/site-packages/octodns/provider/cloudflare.py", line 261, in populate
    records = self.zone_records(zone)
  File "/usr/local/lib/python2.7/site-packages/octodns/provider/cloudflare.py", line 214, in zone_records
    zone_id = self.zones.get(zone.name, False)
  File "/usr/local/lib/python2.7/site-packages/octodns/provider/cloudflare.py", line 114, in zones
    resp = self._request('GET', '/zones', params={'page': page})
  File "/usr/local/lib/python2.7/site-packages/octodns/provider/cloudflare.py", line 101, in _request
    raise CloudflareError(resp.json())
octodns.provider.cloudflare.CloudflareError: Invalid request headers
```

jrenggli:
Thanks for this fix. Tested it and works for me.
What has to be done to get this merged?

octodns/provider/cloudflare.py
parkr: @mamercad What scopes are necessary for the token?

jrenggli: I had to grant zone resources too. See screenshot below
Screenshot from 2020-02-05 16-31-17

Currently it's not possible to only use a single zone granted in Cloudflare as access to `/zones` in API is restricted.

ross:
> What has to be done to get this merged?

Looks like there's some python2/3 issues since this PR predates that work. After that if CI is passing we'd need to make a pass through the code for final review and ship if it's there.

octodns/provider/cloudflare.py
parkr: @mamercad This wasn't working for me! It turns out I had to do this:

```suggestion
                'Authorization': 'Bearer {}'.format(token),
```

Using Python 2.7.17.

mamercad: Thanks, updated...

parkr:
I'd really love to see this merged, but I'm afraid I'm not sure how to approach the Python 3 CI errors. @mamercad, are you able to grok [the failures](https://github.com/github/octodns/pull/402/checks?check_run_id=428278145)? They appear to come from the new test file you added.

mamercad:
Busy with $DAYJOB lately, I’ll try and have a look this weekend.

ross:
High level it doesn't really make sense to copy the full Cloudflare provider tests just to test the no email case since all of the auth bits are stubbed out anyway. There were a number of Python 3 fixes in that file and looks like this PR happened prior to that. I'll push up a commit shortly that should get things working.

parkr:
Thank you so much, @ross!

mamercad:
Thanks much @ross!

PR #405, opened 2019-09-25
Add a Transip provider

maikelpoot:
Transip is a dutch domain registrar (transip.nl) with an api that supports DNS management.

This PR adds a provider that can ben used both as a target and a source using the transip sdk from pip `$ pip install transip`

Provider passes lint, test and coverage and has been tested against a live domain with the following yamlProvider config: (domain and ip's changed to `example.com` and `1.1.1.1`)

```
'':
  - type: A
    values:
      - 1.1.1.1
  - type: MX
    values:
      - exchange: mail3.example.com.
        preference: 10
      - exchange: mail4.example.com.
        preference: 20
  - type: SSHFP
    values:
    - algorithm: 1
      fingerprint: bf6b6825d2977c511a475bbefb88aad54a92ac73
      fingerprint_type: 1
    - algorithm: 1
      fingerprint: 7491973e5f8b39d5327cd4e08bc81b05f7710b49
      fingerprint_type: 1
'*':
  - type: A
    values:
      - 1.1.1.1
'_testsrv._tcp':
  - type: SRV
    values:
      - port: 8080
        priority: 10
        target: test-aaa.example.com.
        weight: 10
      - port: 8080
        priority: 20
        target: test-bbb.example.com.
        weight: 10
'blaat':
  - type: A
    values:
      - 127.0.0.1
      - 127.0.0.2
'test-a':
  - type: A
    values:
      - 1.1.1.1
'test-aaa':
  - type: AAAA
    values:
      - ::1
'test-cname':
  - type: CNAME
    value: test-a.example.com.
'test-submx':
  - type: MX
    values:
      - exchange: mail1.example.com.
        preference: 10
      - exchange: mail2.example.com.
        preference: 20
'test-text':
  - type: TXT
    values:
    - 'dit is text\; blaat'
    - 'test-text 23'
    - 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs'
```

### DIFF
--- a/octodns/provider/transip.py
+++ b/octodns/provider/transip.py
@@ -23,13 +23,16 @@ class TransipProvider(BaseProvider):
         class: octodns.provider.transip.TransipProvider
         # Your Transip account name (required)
         account: yourname
-        # The api key (required)
+        # Path to a private key file (required if key is not used)
+        key_file: /path/to/file
+        # The api key as string (required if key_file is not used)
         key: |
             \'''
             -----BEGIN PRIVATE KEY-----
             ...
             -----END PRIVATE KEY-----
             \'''
+        # if both `key_file` and `key` are presented `key_file` is used
 
     '''
     SUPPORTS_GEO = False
@@ -41,13 +44,18 @@ class TransipProvider(BaseProvider):
     TIMEOUT = 15
     ROOT_RECORD = '@'
 
-    def __init__(self, id, account, key, *args, **kwargs):
+    def __init__(self, id, account, key=None, key_file=None,  *args, **kwargs):
         self.log = getLogger('TransipProvider[{}]'.format(id))
         self.log.debug('__init__: id=%s, account=%s, token=***', id,
                        account)
         super(TransipProvider, self).__init__(id, *args, **kwargs)
 
-        self._client = DomainService(account, key)
+        if key_file is not None:
+            self._client = DomainService(account, private_key_file=key_file)
+        elif key is not None:
+            self._client = DomainService(account, private_key=key)
+        else:
+            raise Exception('Missing `key` of `key_file` parameter in config')
 
         self.account = account
         self.key = key
--- a/octodns/provider/transip.py
+++ b/octodns/provider/transip.py
@@ -1,0 +1,328 @@
+#
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from suds import WebFault
+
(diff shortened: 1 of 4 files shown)


octodns/provider/transip.py
```diff
@@ -0,0 +1,332 @@
...
+                       target, lenient)
+
+        before = len(zone.records)
+        try:
+            zoneInfo = self._client.get_info(zone.name[:-1])
+        except WebFault as e:
+            if e.fault.faultcode == '102' and target is False:
+                self.log.warning(
+                    'populate: (%s) Zone %s not found in account ',
+                    e.fault.faultcode, zone.name)
```
ross: Would this happen on the first run before a zone has been created? If so `warning` is probably too high of a level as that's an "OK" case.

octodns/provider/transip.py
```diff
@@ -0,0 +1,332 @@
...
+                return exists
+            elif e.fault.faultcode == '102' and target is True:
+                self.log.warning('populate: Transip can\'t create new zones')
+                raise Exception(
+                    ('populate: ({}) Transip used ' +
+                     'as target for non-existing zone: {}').format(
+                        e.fault.faultcode, zone.name))
```
ross: It's minor, but I prefer that providers throw a specific exception rather than a generic `Exception`. That can be as simple as:

```python
class TransipException(Exception):
    pass
```

octodns/provider/transip.py
```diff
@@ -0,0 +1,332 @@
...
+        for value in record.values:
+            value = value.replace('\\;', ';')
+            _entries.append(DnsEntry(name, record.ttl, record._type, value))
+
+        return _entries
+
+    def _parse_to_fqdn(self, value):
+
+        # Enforce switch from suds.sax.text.Text to string
+        value = '' + value
```
ross: Would `str()` work here. Might be a little cleaner than concatenation.

octodns/provider/transip.py
maikelpoot: Indeed, checked other providers and they didn't throw warnings, just an empty zone.
Will remove the warning

octodns/provider/transip.py
maikelpoot: Agreed

octodns/provider/transip.py
maikelpoot: str() works fines as well, will change

PR #408, opened 2019-09-30
Rel 0.9.8 with a bunch of pypi readme/description fixes

ross:
Managed to upload a test/in progress fix of the description by accident and couldn't fix it b/c of the filename reuse prevention. This should hopefully get the fully fixed version along with the fixes themselves. Differences in markdown variants are annoying :-(


ross:
This works as much as I care to make it at the moment. The markdown handling is pretty lacking in general and doesn't seem to support many/any of the advanced stuff. Oh well. Better than it used to be when it just put the raw MD.

PR #411, opened 2019-10-09
Fix 'server error: zone not found' for NS1 provider

josvo:
When writing a zone to the NS1 provider, there is the following error message if a zone does not already exist: "ns1.rest.errors.ResourceException: server error: zone not found"
This pull requests migrates from the legacy nsone SDK to the new ns1-python SDK, see README in https://github.com/ns1/nsone-python which fixes this error.


ross:
Thanks.

PR #415, opened 2019-10-18
Validate Record name & fqdn length

ross:
We've run into records that were too long a couple of times now and this will catch them earlier in the process during validation rather than having providers throw errors when the names don't conform.

/cc https://stackoverflow.com/a/44523868 which has the limits


ross:
Fixes https://github.com/github/octodns/issues/255

Issue #420, opened 2019-11-13
Public and Private DNS in Route53


opardalis:
Am I missing step somewhere? How to distinguish between the public and private hosted zone types in Route53?

We have a single domain for both public and private, the data within is obviosuly very different. We'd like to migrate away from RoadWorker and possible use OctoDNS. When I run octodns-dump it only grabs the private data, not the public. How do you grab both, and ultimately manage and sync both from separate config files?

Thanks

ross:
Not familiar with a setup that uses a single zone for both internal and external records and octoDNS doesn't have any Route53 code specific to handling that sort of set up. We manage dozens of internal zones and dozens of external zones, but nothing that's both.

yzguy:
#260 and #190 are essentially the same thing. It's somewhat trivial to add it just looking at public hosted zones or private hosted zones. The tougher part is how to keep them in config/dump because with a very basic configuration one would overwrite the other.

I believe your earlier response: https://github.com/github/octodns/issues/260#issuecomment-417824820 is probably the way to go

You would have providers configured like 

```
  route53:
    class: octodns.provider.route53.Route53Provider
    access_key_id: env/AWS_ACCESS_KEY_ID
    secret_access_key: env/AWS_SECRET_ACCESS_KEY
```

```
  route53_private:
    class: octodns.provider.route53.Route53Provider
    access_key_id: env/AWS_ACCESS_KEY_ID
    secret_access_key: env/AWS_SECRET_ACCESS_KEY
    private_zones: true
``

Then you can use them like normal.

I have a draft PR up for the small change but need to work on coverage and hear any feedback: https://github.com/github/octodns/pull/421

This would be somewhat of an impactful change for people who use OctoDNS for private hosted zone management already, as they would need to go add the `private_zones: true` to their provider.

opardalis:
Thanks, I use that info above to make it work. Honestly, I just need to get the data out of Route53 via octodns-dump, we're migrating to a new provider and in the future the public/private wont be over the same domain again. The dump saves me time since we have so many records.

Issue #260, opened 2018-08-29
Handle multiple zones with the same name.


f33rx:
Have this situation where I have multiple zone names that are the same, only the last element is returned.

yzguy:
Is this in the case of a zone that is a Public Hosted Zone and Private Hosted Zone as well? I had been wondering about this case.

f33rx:
@yzguy It is! In this case, it's a public and two private zones. (!!) The intent was to persist a zone config to disk and then use the `backup` to load other providers.

ross:
My first thought would be to break things up into separate configs, e.g.

```
config/
  external/
    zone1.com.yaml
  external.yaml
  internal/
    zone1.com.yaml
  internal.yaml
``` 

Things would then be synced with two seperate runs, one with `--config-file config/external.yaml` and another with `--config-file config/internal.yaml`. Each would point the YamlProvider to matching directory. 

Something like that make sense/work or there more nuance to what you're trying to do?

brianeclow:
If I am understanding your suggestion, then for a split config/data setup like mine. The files would be like:

```
config/
   internal.yaml
   external.yaml
   simple.yaml
data/
   internal/
      zone1.com.yaml
   exteranl/
      zone1.com.yaml
   zone2.com.yaml
```

With the `data` keys for `internal` and `external` having the values `./data/internal` and `./data/external` respectively. Then  for `simple` config, the `data` key have the value `./data`

Correct?

f33rx:
`https://github.com/github/octodns/blob/6da23ea7956298c53bb9985a35d0a30fac6cd9a8/octodns/provider/route53.py#L283` Seems to use the name field to find the ID. If I ask route53 for all the things called `domain.com` and I get three things. It seems to choose the last thing it got from the api. I tried to add an optional field for specifying an array of `zone id`'s but failed spectacularly.

ross:
> If I am understanding your suggestion, then for a split config/data setup like mine.

Not quite sure what `simple.yaml` would be for/meaning here. Maybe i'd be better to think of things as two completely different octoDNS setups/configs, one for internal and one for external that know nothing about each other. 

## internal 

### `config/internal.yaml`

```yaml
---
providers:
  config:
    class: octodns.provider.yaml.YamlProvider
    directory: ./data/internal/

  powerdns:
    class: octodns.provider.powerdns.PowerDnsProvider
    host: 1.2.3.4
    api_key: env/POWERDNS_API_KEY_EXT 

zones:

[...]

Issue #190, opened 2018-02-06
Single domain that is both public and private in Route53


brianeclow:
I have been poking at this off and on for a few months, and I wasn't sure if it was an octoDNS issue or a me issue.

We have a single domain set as two hosted zones in Route53, one public and one private. When attempting an `octodns-dump` against the domain, octoDNS is throwing an error about an empty geo value. There are no geo records set for either hosted zone.

config:
```
---
providers:
  config:
    class: octodns.provider.yaml.YamlProvider
    directory: ./data
  route53:
    class: octodns.provider.route53.Route53Provider
    access_key_id: env/AWS_ACCESS_KEY_ID
    secret_access_key: env/AWS_SECRET_ACCESS_KEY

zones:
  sillydomain.com.:
    sources:
      - route53
    targets:
      - config
```

octodns-dump debug output:
```
2018-02-06T09:50:14  [140736519537600] INFO  Manager __init__: config_file=config/ro_config_route53.yaml
2018-02-06T09:50:14  [140736519537600] INFO  Manager __init__:   max_workers=1
2018-02-06T09:50:14  [140736519537600] INFO  Manager __init__:   max_workers=False
2018-02-06T09:50:14  [140736519537600] DEBUG Manager __init__:   configuring providers
2018-02-06T09:50:14  [140736519537600] DEBUG YamlProvider[config] __init__: id=config, directory=./data, default_ttl=3600, enforce_order=1
2018-02-06T09:50:14  [140736519537600] DEBUG YamlProvider[config] __init__: id=config, apply_disabled=False, update_pcent_threshold=0, delete_pcent_threshold=0
2018-02-06T09:50:14  [140736519537600] DEBUG Route53Provider[route53] __init__: id=route53, access_key_id=KEY, secret_access_key=***
2018-02-06T09:50:14  [140736519537600] DEBUG Route53Provider[route53] __init__: id=route53, apply_disabled=False, update_pcent_threshold=0, delete_pcent_threshold=0
2018-02-06T09:50:14  [140736519537600] INFO  Manager dump: zone=sillydomain.com., sources=()
2018-02-06T09:50:14  [140736519537600] DEBUG YamlProvider[dump] __init__: id=dump, directory=data/, default_ttl=3600, enforce_order=1
2018-02-06T09:50:14  [140736519537600] DEBUG YamlProvider[dump] __init__: id=dump, apply_disabled=False, update_pcent_threshold=0, delete_pcent_threshold=0
[...]

PR #421, opened 2019-11-14
Route53Provider: Allow specifying public or private hosted zones

yzguy:
Will address #420, #260 and #190
Need to finish up the code coverage

This would allow you to specify 2 providers, one with `private_zones` set to `true`. That provider would only look at the private hosted zones, whereas by default it will only look at public hosted zones.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -600,6 +600,8 @@ class Route53Provider(BaseProvider):
         # The AWS session token (optional)
         # Only needed if using temporary security credentials
         session_token:
+        # Only use private hosted zones (optional)
+        private_zones: true
 
     Alternatively, you may leave out access_key_id, secret_access_key
     and session_token.
@@ -618,7 +620,7 @@ class Route53Provider(BaseProvider):
 
     def __init__(self, id, access_key_id=None, secret_access_key=None,
                  max_changes=1000, client_max_attempts=None,
-                 session_token=None, *args, **kwargs):
+                 session_token=None, private_zones=False, *args, **kwargs):
         self.max_changes = max_changes
         _msg = 'access_key_id={}, secret_access_key=***, ' \
                'session_token=***'.format(access_key_id)
@@ -643,7 +645,7 @@ class Route53Provider(BaseProvider):
                                 aws_secret_access_key=secret_access_key,
                                 aws_session_token=session_token,
                                 config=config)
-
+        self._private_zones = private_zones
         self._r53_zones = None
         self._r53_rrsets = {}
         self._health_checks = None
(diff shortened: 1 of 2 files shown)


yzguy:
This is close, after getting the tests working and taking it out of Draft PR, I had come up with issues so I made the tests fail again so it wouldn't be merged prematurely.

Right now with the new code, if the private hosted zone doesn't exist, OctoDNS will attempt to create it, but as a public hosted zone. One issue with automatically creating the private hosted zone is that you have to specify a VPC ID that it goes with. I'm thinking the best approach for this is just to raise an Exception and say creation of private hosted zones is unsupported, and that it must be created manually before it can be managed by OctoDNS

The other thing I was thinking of is that this could be a "breaking" change for someone who uses the current version of the Route53Provider with a mix of public and private hosted zones, but none that have the same name.

In that case, this new code would no longer find the private hosted zones and would attempt to make public hosted zones with the same names. A user would probably have to create a new section in their configuration for a `route53_private` version of the provider, and swap all their private hosted zones to use it.

I'm not sure if that's acceptable or not. I'd have to do some thinking to figure out a solution that doesn't create that breaking behavior. The public/private with the same name make it really hard, as the zone name is the unique ID throughout OctoDNS, and other providers don't have the concept of private hosted_zones.

ross:
> One issue with automatically creating the private hosted zone is that you have to specify a VPC ID that it goes with ...

There's been a couple things come up that want/need the ability to attach metadata at the zone level, this seems like it'd be another ++ on that idea (don't remember the others offhand.) Something like:

```yaml
  foo.internal.net.:
    _octodns:
        route53:
          vpc-id: vpc-123456 # implies private
```

That may be more of a change than makes sense to add here. The blow up if it doesn't exist and tell them to create it seems fine (and we can circle back someday and improve things with something like the above.) IIRC the main issue is that there's not really a logical place for the data to be hydrated/attached as the way the manager works with zones currently doesn't involve a Zone object that matches the needed scope (been a long time since I thought about this so could be mis-remembering.) I think it should be possible to address this, just didn't seem like a trivial change at the time.

> The other thing I was thinking of is that this could be a "breaking" change for someone who uses the current version of the Route53Provider with a mix of public and private hosted zones, but none that have the same name. ...

That's a good question/point. Maybe the default (nothing mentioned) has to continue to work the way it currently does and then things explicitly have to say `public` vs `private` to get the new behavior here?

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-route53/. Feel free to rework/reopen this PR there.

JohanLeirnes:
Hey, this looked like a brilliant idea.

Where you able to implement this in some other way to get the split horizon in r53 with octodns? @ross

PR #429, opened 2019-12-04
Quick expirimentation with NS1 dynamic support

ross:


ross:
Replaced by https://github.com/github/octodns/pull/433

PR #430, opened 2019-12-10
AWS Route53: Add Parameter for DelegationSetId

joschi36:
To support DelegationSets for the AWS Provider, I've added a new optional parameter to the Route53 provider.

This fixes https://github.com/github/octodns/issues/320


octodns/provider/route53.py
```diff
@@ -674,10 +675,16 @@ def _get_zone_id(self, name, create=False):
...
             ref = uuid4().hex
+            del_set = self.delegation_set_id
             self.log.debug('_get_zone_id:   no matching zone, creating, '
                            'ref=%s', ref)
-            resp = self._conn.create_hosted_zone(Name=name,
-                                                 CallerReference=ref)
+            if self.delegation_set_id:
+                resp = self._conn.create_hosted_zone(Name=name,
+                                                     CallerReference=ref,
+                                                     DelegationSetId=del_set)
```
ross: Can `DelegationSetId=None` be passed here to avoid the `if`?

octodns/provider/route53.py
```diff
@@ -674,10 +675,16 @@ def _get_zone_id(self, name, create=False):
             return id
         if create:
             ref = uuid4().hex
+            del_set = self.delegation_set_id
             self.log.debug('_get_zone_id:   no matching zone, creating, '
                            'ref=%s', ref)
-            resp = self._conn.create_hosted_zone(Name=name,
-                                                 CallerReference=ref)
+            if self.delegation_set_id:
```
ross:
```suggestion
            if del_set:
```

Since it's pulled out into a local above and used already below.

octodns/provider/route53.py
joschi36: no i've tested it and it failed

joschi36:
@ross can you help me with the test coverage?

octodns/provider/route53.py
ross: :-(, that's actually why I was asking, wouldn't really have to worry about test coverage in that case.

ross:
> @ross can you help me with the test coverage?

The bit in https://github.com/github/octodns/blob/31d904181132858735736e1439dc049ef7f617da/tests/test_octodns_provider_route53.py#L365-L369 will let you create a new provider and pass in params. You can probably create a new `def test_...` function and stick that bit up top in it and get yours set up with a delegation set id. 

The method that looks closest to doing what you're needing so far as I can (quickly) tell is https://github.com/github/octodns/blob/31d904181132858735736e1439dc049ef7f617da/tests/test_octodns_provider_route53.py#L831

You can probably pick enough of the bits out of it to get a call to create mocked https://github.com/github/octodns/blob/31d904181132858735736e1439dc049ef7f617da/tests/test_octodns_provider_route53.py#L852-L877 and in your case look for the `DelegationSetId` to also be present in https://github.com/github/octodns/blob/31d904181132858735736e1439dc049ef7f617da/tests/test_octodns_provider_route53.py#L873-L877

joschi36:
I tried to add testing to the branch. (My first time with python testing)
@ross can you look over it?

ross:
> @ross can you look over it?

Was OOTO for an extended holiday break. I've put this on my todo list and will take a look once I get to it. Hopefully in the next couple days.

ross:
> I've tested with multiple production zones. 👍

Sweet.

PR #433, opened 2019-12-10
Ns1Provider dynamic record support

ross:
Add the initial round of dynamic support to Ns1Provider. There are a few TODO's that I'll probably leave to another PR since this one is already big enough. Very little (almost nothing) of the existing code needed to change for this which is good, but there's a fair bit of logic so the PR isn't small. 

/cc https://github.com/github/octodns/pull/432 and https://github.com/github/octodns/pull/431 which this builds off of.


octodns/provider/ns1.py
```diff
@@ -275,49 +545,342 @@ def populate(self, zone, target=False, lenient=False):
...
+        monitor_id = monitor['id']
+        self.log.debug('_monitor_create:   monitor=%s', monitor_id)
+
+        return monitor_id, self._feed_create(monitor)
+
+    def _monitor_gen(self, record, value):
+        host = record.fqdn[:-1]
+        _type = record._type
+
+        request = 'GET {path} HTTP/1.0\\r\\nHost: {host}\\r\\n' \
```
theojulienne: Optional, but can also be written like this I believe
```suggestion
        request = r'GET {path} HTTP/1.0\r\nHost: {host}\r\n' \
```

octodns/provider/ns1.py
```diff
@@ -275,49 +545,342 @@ def populate(self, zone, target=False, lenient=False):
...
+            if fallback:
+                notes['fallback'] = fallback
+
+            country = set()
+            georegion = set()
+            us_state = set()
+
+            for geo in rule.data.get('geos', []):
+                n = len(geo)
+                if n == 8:
```
theojulienne: Comment examples inline of the strings would be nice here to help remind the reader what we're parsing.

octodns/provider/ns1.py
```diff
@@ -275,49 +545,342 @@ def populate(self, zone, target=False, lenient=False):
...
+        } for v in record.values]
+
+        # Build our list of answers
+        answers = []
+        for pool_name in sorted(pools.keys()):
+            priority = 1
+
+            # Dynamic/health checked
+            current_pool_name = pool_name
+            while current_pool_name:
```
theojulienne: Usual caveat that it would be best to have an infinite loop detection here just to be better safe than sorry.

ross: Added something simple here, but fwiw it's not really possible to create a `Record` with a fallback loop due to https://github.com/github/octodns/blob/208e2b36168cbb7e3972a69de82a5a28d836dc1f/octodns/record/__init__.py#L554-L566.

octodns/provider/ns1.py
ross: Done.

PR #438, opened 2019-12-18
Bug fix: Bypass HTTP exception when adding non existing zones to powerdns

aashifb:
This is in reference to issue https://github.com/github/octodns/issues/437


octodns/provider/powerdns.py
```diff
@@ -178,7 +178,8 @@ def populate(self, zone, target=False, lenient=False):
                 # Nicer error message for auth problems
                 raise Exception('PowerDNS unauthorized host={}'
                                 .format(self.host))
-            elif e.response.status_code == 422:
+            elif e.response.status_code == 422 or \
+                    e.response.status_code == 404:
```
ross: Slightly cleaner to do an `in` here. 
```
     elif e.response.status_code in (404, 422):
```

ross:
Actually looks like the tests are failing. That'll need addressed first.

aashifb:
> Actually looks like the tests are failing. That'll need addressed first.

aashifb:
Thanks @ross. I'll look at the tests.

yzguy:
This behavior was fixed in #540, closing this as unnecessary.

PR #441, opened 2019-12-31
cibuild: Add flake8 tests for Python syntax errors

cclauss:
Also add flake8 tests for undefined names, etc.  __flake8__ is a superset of both pycodestyle and pyflakes so installing flake8 in __requirements-dev.txt__ will continue to install both pycodestyle and pyflakes.

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.


requirements-dev.txt
```diff
@@ -1,8 +1,7 @@
 coverage
+flake8
 mock
 nose
-pycodestyle==2.4.0
-pyflakes==1.6.0
```
ross: Would prefer to leave these in here to pin versions (feel free to update the version to the current/latest.) 

We've run in to problems in the past where CI starts failing unrelated to changes in PRs due to updates the CI process grabbing a newer version than the codebase has been vetted by. (new checks added or existing updated.)

requirements-dev.txt
```diff
@@ -1,8 +1,7 @@
 coverage
+flake8
```
ross: Please pin a version here.

script/lint
```diff
@@ -17,5 +17,6 @@ fi
 
 SOURCES="*.py octodns/*.py octodns/*/*.py tests/*.py"
 
-pycodestyle --ignore=E221,E241,E251,E722,W504 $SOURCES
+pycodestyle --ignore=E117,E221,E241,E251,E722,W504 $SOURCES
```
ross: What's the reason for adding `E117`. That check get updated to be more strict/different?

requirements-dev.txt
cclauss: I tend to pin runtime dependencies but allow testing tools, linters, formatters, etc. to float.  As new tests are added/improved, the testing gets more rigorous and more issues that should get fixed are fixed which improves code quality and resiliency.  It does not matter if the issue is related to a particular PR.  What matters is the code continues to be improved.

Take pyflakes for instance...  Here it is hardcoded so none of the [improved tests](https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst) between 1.6 and 2.1.1 have been run on this codebase.  That is unfortunate because they could have improved code quality and resiliency.

script/lint
cclauss: E117 over-indented seems to disagree with the code formatted that this project uses.

ross: Hrm. Guess that's something that's changed between 2.4.0 and now.

requirements-dev.txt
ross: Understand that, but the decision was made to trade that off and to update the pinned versions from time to time.

ross: Also note that if you go back in time and check out a specific release tag or SHA and install requirements it'll fail to pass linting b/c that code/tag wasn't tested with versions of things that didn't exist at the time. Having them pinned keeps released tested and validated with the version in use at the time. 

I expect dependabot to be able to file PRs to update these versions as the come available in the hopefully not too distant future which should get things to the ideal set up.

script/lint
ross: The new warnings were fixed in master.

script/lint
```diff
@@ -19,3 +19,4 @@ SOURCES="*.py octodns/*.py octodns/*/*.py tests/*.py"
 
 pycodestyle --ignore=E221,E241,E251,E722,W504 $SOURCES
 pyflakes $SOURCES
+flake8 . --count --exclude=./env/* --select=E9,F63,F7,F82 --show-source --statistics
```
ross: Running into problems with the way this does `.` when I have extra virtualenvs in my working directory. It tries to run on all the code in them.

```suggestion
flake8 $SOURCES --count --select=E9,F63,F7,F82 --show-source --statistics
```

Ideally Would like to remove the `--select` eventually and get all the warnings fixed. Can do that in subsequent PR(s).

PR #442, opened 2020-01-08
Fix trailing semicolon in dkim for ovh provider

Kaari:
Pulling dns records from ovh to a yaml file puts a semicolon at the end.
Pushing from yaml to ovh will fail the "dkim-compliant" verification if
there is an empty field (and there is one in case of a trailing
semicolon). With the current logic, pulling dkim record created with ovh
then pushing it back will NOT work.

This small patch ignores all empty fields in a dkim records during dkim
validation.

### DIFF
--- a/octodns/provider/ovh.py
+++ b/octodns/provider/ovh.py
@@ -323,7 +323,7 @@ class OvhProvider(BaseProvider):
                           'n': lambda _: True,
                           'g': lambda _: True}
 
-        splitted = list(filter(None, value.split('\\;')))
+        splitted = [v for v in value.split('\\;') if v]
         found_key = False
         for splitted_value in splitted:
             sub_split = [x.strip() for x in splitted_value.split("=", 1)]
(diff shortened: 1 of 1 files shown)


octodns/provider/ovh.py
```diff
@@ -323,7 +323,7 @@ def _is_valid_dkim(self, value):
                           'n': lambda _: True,
                           'g': lambda _: True}
 
-        splitted = value.split('\\;')
+        splitted = list(filter(None, value.split('\\;')))
```
ross: The python 3 porting guides encourage comprehensions rather than `list(filter(` for these cases. I believe that'd be something like the following here:

```suggestion
        splitted = [v for v in value.split('\\;') if v]
```

Kaari: Clearer indeed. Thanks for the help.

ross:
Thanks for the fix.

PR #465, opened 2020-01-16
Add codespell to find common misspellings

cclauss:


ross:
If we're going to include this I'd prefer this run as part of `./script/lint` so that it's easier to do locally and part of the pre-commit hooks, etc.

PR #466, opened 2020-01-16
Add a way to configure Ns1Provider monitoring regions for records

ross:
This provides a way to manually specify the regions a NS1 Monitor should run in. There may be a more dynamic way to find this in the future and apply the "optimal" value, but for now it sounds like explicitly specifying it at the provider level makes sense. If/when something more is needed we can shift tact.

### DIFF
--- a/tests/test_octodns_provider_ns1.py
+++ b/tests/test_octodns_provider_ns1.py
@@ -1199,47 +1199,47 @@ class TestNs1ProviderDynamic(TestCase):
         self.assertFalse(extra)
         monitors_for_mock.assert_not_called()
 
+    DESIRED = Zone('unit.tests.', [])
+
+    SIMPLE = Record.new(DESIRED, 'sim', {
+        'ttl': 33,
+        'type': 'A',
+        'value': '1.2.3.4',
+    })
+
+    # Dynamic record, inspectable
+    DYNAMIC = Record.new(DESIRED, 'dyn', {
+        'dynamic': {
+            'pools': {
+                'iad': {
+                    'values': [{
+                        'value': '1.2.3.4',
+                    }],
+                },
+            },
+            'rules': [{
+                'pool': 'iad',
+            }],
+        },
+        'octodns': {
+            'healthcheck': {
+                'host': 'send.me',
+                'path': '/_ping',
+                'port': 80,
+                'protocol': 'HTTP',
+            }
+        },
+        'ttl': 32,
+        'type': 'A',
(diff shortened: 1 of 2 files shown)


tests/test_octodns_provider_ns1.py
```diff
@@ -1198,6 +1199,138 @@ def test_extra_changes(self, monitors_for_mock):
         self.assertFalse(extra)
         monitors_for_mock.assert_not_called()
 
+    def test_has_dynamic(self):
+        provider = Ns1Provider('test', 'api-key')
+
+        desired = Zone('unit.tests.', [])
+
+        simple = Record.new(desired, 'sim', {
```
pavantc: Should we move this and the dynamic record below to a function - getSimpleTestRecord() and getDynamicTestRecord() and use as needed? It exists in the current code and duplicated twice in the new tests in this change.

tests/test_octodns_provider_ns1.py
```diff
@@ -1198,6 +1199,138 @@ def test_extra_changes(self, monitors_for_mock):
...
+
+        # Blows up and apply not called even though there's a simple
+        apply_update_mock.reset_mock()
+        with self.assertRaises(Ns1Exception) as ctx:
+            provider._apply(both_plan)
+        self.assertTrue('monitor_regions not set' in text_type(ctx.exception))
+        apply_update_mock.assert_not_called()
+
+        # with monitor_regions set
+        provider.monitor_regions = ['lga']
```
pavantc: While this might be an unlikely situation, since we are handling monitor_regions not defined, should we handle monitor_regions = [] (empty list)?

ross: I believe this will either be allowed or NS1 will throw an error if it's not. This code shouldn't need to enforce a behavior there.

tests/test_octodns_provider_ns1.py
```diff
@@ -1199,47 +1199,47 @@ def test_extra_changes(self, monitors_for_mock):
...
+                'protocol': 'HTTP',
+            }
+        },
+        'ttl': 32,
+        'type': 'A',
+        'value': '1.2.3.4',
+        'meta': {},
+    })
+
+    def test_has_dynamic(self):
```
pavantc: Didn't look in detail, but hoping this is not "update"d in multiple tests in the same run. Assuming we need a pristine copy of these records for each test and that behavior is not changed by moving them as instance variables.

PR #467, opened 2020-01-19
ConstellixProvider: zone creation and records in one run

### LINKED ISSUE #463: Constellix Provider: Fails to create new zone and add requested records in one run
If I add a new zone and records into my config, and the zone does not already exist at Constellix, my expectation would be that when I execute octodns-sync, in that run, a new zone would be created and records added as well.  What actually happens is, the new zone is indeed created, but record creation fails, raising ConstellixClientNotFound().  If I subsequently run octodns-sync immediately thereafter, with no change in my config, the records are then created with no error.  The Constellix API requires providing a zone ID typically when performing a record operation, and I believe the provider is just missing the logic to make itself aware of the zone it just created, before moving on to record creation.  I can tweak the ConstellixClient class' domains method to make this work but I don't think I'm doing it the best way.

I am using octodns version 0.9.9 and python 2.7.17

yzguy:
Fixes #463 

Previously you would have to do 2 runs to sync to Constellix, one for zone creation, and another for record creation.

This fixes that by making sure OctoDNS can find the zone ID after it's created.

### DIFF
--- a/octodns/provider/constellix.py
+++ b/octodns/provider/constellix.py
@@ -88,7 +88,7 @@ class ConstellixClient(object):
         if self._domains is None:
             zones = []
 
-            resp = self._request('GET', '/').json()
+            resp = self._request('GET', '').json()
             zones += resp
 
             self._domains = {'{}.'.format(z['name']): z['id'] for z in zones}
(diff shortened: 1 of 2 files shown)


octodns/provider/constellix.py
```diff
@@ -85,7 +85,7 @@ def _request(self, method, path, params=None, data=None):
 
     @property
     def domains(self):
-        if self._domains is None:
+        if not self._domains:
```
ross: Just to double check. I believe this fall into the block any time this is called and there are no zones returned from the API. That probably won't hurt anything since the end result will be the same, but it could result in multiple calls to the API to fetch the empty zone list. Seems OK, but want to make sure that's expected.

yzguy: Yep you were right, after I ran `--debug` I saw a bunch of API calls that weren't necessary. Fixed it up to be cleaner by checking if the ID was found or not.

yzguy:
This should be cleaned up and working better.

PR #489, opened 2020-02-10
Support DNSimple sandbox

johnlane:
An optional parameter 'sandbox' can be used to select the base URL
for the Sandbox API (see https://developer.dnsimple.com/sandbox ).

### DIFF
--- a/octodns/provider/dnsimple.py
+++ b/octodns/provider/dnsimple.py
@@ -30,13 +30,16 @@ class DnsimpleClientUnauthorized(DnsimpleClientException):
 
 
 class DnsimpleClient(object):
-    BASE = 'https://api.dnsimple.com/v2/'
 
-    def __init__(self, token, account):
+    def __init__(self, token, account, sandbox):
         self.account = account
         sess = Session()
         sess.headers.update({'Authorization': 'Bearer {}'.format(token)})
         self._sess = sess
+        if sandbox:
+            self.BASE = 'https://api.sandbox.dnsimple.com/v2/'
+        else:
+            self.BASE = 'https://api.dnsimple.com/v2/'
 
     def _request(self, method, path, params=None, data=None):
         url = '{}{}{}'.format(self.BASE, self.account, path)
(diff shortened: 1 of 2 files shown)


tests/test_octodns_provider_dnsimple.py
```diff
@@ -38,6 +38,10 @@ class TestDnsimpleProvider(TestCase):
             break
 
     def test_populate(self):
+
+        # Sandbox
+        provider = DnsimpleProvider('test', 'token', 42, 'true')
```
ross: Please throw in a minimal check here, e.g. 

```suggestion
        provider = DnsimpleProvider('test', 'token', 42, 'true')
        self.assertTrue('sandbox' in provider._client.base)
```

octodns/provider/dnsimple.py
```diff
@@ -30,13 +30,16 @@ def __init__(self):
...
-    def __init__(self, token, account):
+    def __init__(self, token, account, sandbox):
         self.account = account
         sess = Session()
         sess.headers.update({'Authorization': 'Bearer {}'.format(token)})
         self._sess = sess
+        if sandbox:
+            self.BASE = 'https://api.sandbox.dnsimple.com/v2/'
+        else:
+            self.BASE = 'https://api.dnsimple.com/v2/'
```
ross: Now that this is an instance variable rather than a constant please `s/BASE/base/g` so that things match up.

tests/test_octodns_provider_dnsimple.py
johnlane: yep, wasn't sure how to do that. thx.

tests/test_octodns_provider_dnsimple.py
```diff
@@ -38,7 +38,13 @@ class TestDnsimpleProvider(TestCase):
...
+        self.assertTrue('sandbox' in provider._client.base)
+
         provider = DnsimpleProvider('test', 'token', 42)
+        self.assertFalse('sandbox' in provider._client.base)
```
ross: Realized this was a good idea and just added it rather than doing another round trip.

ross:
Thanks!

PR #506, opened 2020-03-03
NS1 doesn't support region OC. Handle it explicitly in the provider

pavantc:
NS1 does not handle region 'OC' (Oceania) which is a valid region that could be part of the rules in dynamic DNS configuration.

The strategy used in this fix is to handle 'OC' specified in the rules and convert it to a list of countries using the corresponding country codes (which is definitely supported in NS1).  This is handled in both directions - while pushing data to NS1 and while reading/interpreting response from NS1:

1. When pushing to NS1:
Any occurrence of 'OC' in the dynamic rules are converted to a list of countries before pushing to NS1

2. In the response:
Presence of *all* the countries that are part of 'OC' is detected and translated back to 'OC' region in the provider before presenting the response to the application. Accordingly, the entries under the 'country' tag are removed (since we are replacing those with 'OC').

The other part of the PR is related to testing the added functionality.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -229,9 +229,9 @@ class Ns1Provider(BaseProvider):
 
     # Necessary for handling unsupported continents in _CONTINENT_TO_REGIONS
     _CONTINENT_TO_LIST_OF_COUNTRIES = {
-        'OC': ['FJ', 'NC', 'PG', 'SB', 'VU', 'AU', 'NF', 'NZ', 'FM', 'GU',
+        'OC': {'FJ', 'NC', 'PG', 'SB', 'VU', 'AU', 'NF', 'NZ', 'FM', 'GU',
                'KI', 'MH', 'MP', 'NR', 'PW', 'AS', 'CK', 'NU', 'PF', 'PN',
-               'TK', 'TO', 'TV', 'WF', 'WS'],
+               'TK', 'TO', 'TV', 'WF', 'WS'},
     }
 
     def __init__(self, id, api_key, retry_count=4, monitor_regions=None, *args,
@@ -367,7 +367,7 @@ class Ns1Provider(BaseProvider):
             # Oceania countries list are found, set the region to OC and remove
             # individual oceania country entries
 
-            oc_countries = []
+            oc_countries = set()
             for country in meta.get('country', []):
                 # country_alpha2_to_continent_code fails for Pitcairn ('PN')
                 if country == 'PN':
@@ -376,16 +376,19 @@ class Ns1Provider(BaseProvider):
                     con = country_alpha2_to_continent_code(country)
 
                 if con == 'OC':
-                    oc_countries.append(country)
-                geos.add('{}-{}'.format(con, country))
+                    oc_countries.add(country)
+                else:
+                    # Adding only non-OC countries here to geos
+                    geos.add('{}-{}'.format(con, country))
 
-            if len(oc_countries):
-                all_oc_countries = self._CONTINENT_TO_LIST_OF_COUNTRIES['OC']
-                if sorted(oc_countries) == sorted(all_oc_countries):
-                    # Remove oceania countries from the map, add 'OC' region
-                    for c in oc_countries:
-                        geos.remove('{}-{}'.format('OC', c))
+            if oc_countries:
+                if oc_countries == self._CONTINENT_TO_LIST_OF_COUNTRIES['OC']:
+                    # All OC countries found, so add 'OC' to geos
                     geos.add('OC')
+                else:
+                    # Partial OC countries found, just add them as-is to geos
+                    for c in oc_countries:
+                        geos.add('{}-{}'.format('OC', c))
 
             # States are easy too, just assume NA-US (CA providences aren't
             # supported by octoDNS currently)
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -227,6 +227,13 @@ class Ns1Provider(BaseProvider):
         'NA': ('US-CENTRAL', 'US-EAST', 'US-WEST'),
     }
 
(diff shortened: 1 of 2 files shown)


tests/test_octodns_provider_ns1.py
```diff
@@ -1092,6 +1135,20 @@ def test_data_for_dynamic_A(self):
         data2 = provider._data_for_A('A', ns1_record)
         self.assertEquals(data, data2)
 
+        # Oceania test cases
+        # 1. Full list of countries should return 'OC' in geos
+        oc_country_list = self._CONTINENT_TO_LIST_OF_COUNTRIES['OC']
+        ns1_record['regions']['lhr']['meta']['country'] = oc_country_list
+        data3 = provider._data_for_A('A', ns1_record)
+        assert('OC' in data3['dynamic']['rules'][0]['geos'])
```
pavantc: I probably should add another assert here that none of the 'OC' countries are present under the 'country' tag.

octodns/provider/ns1.py
```diff
@@ -351,12 +358,35 @@ def _data_for_dynamic_A(self, _type, record):
...
+                else:
+                    con = country_alpha2_to_continent_code(country)
+
+                if con == 'OC':
+                    oc_countries.append(country)
                 geos.add('{}-{}'.format(con, country))
 
+            if len(oc_countries):
+                all_oc_countries = self._CONTINENT_TO_LIST_OF_COUNTRIES['OC']
+                if sorted(oc_countries) == sorted(all_oc_countries):
```
ross: If you change `oc_countries` and `_CONTINENT_TO_LIST_OF_COUNTRIES` to be `set`s rather than `list`s you can just compare them directly here. Should be simpler/cleaner than having to sort.

octodns/provider/ns1.py
```diff
@@ -351,12 +358,35 @@ def _data_for_dynamic_A(self, _type, record):
...
-                con = country_alpha2_to_continent_code(country)
+                # country_alpha2_to_continent_code fails for Pitcairn ('PN')
+                if country == 'PN':
+                    con = 'OC'
+                else:
+                    con = country_alpha2_to_continent_code(country)
+
+                if con == 'OC':
+                    oc_countries.append(country)
                 geos.add('{}-{}'.format(con, country))
```
ross: If you do an `else` here and only add non-OC things to `geos` you can avoid having to iterate and remove them down below. The section below can just have an `else` for "not all OC" and in that else loop over what you do have and add them in. It's minor nits, but feels a bit cleaner than adding them and then maybe removing them later.

tests/test_octodns_provider_ns1.py
```diff
@@ -521,6 +521,12 @@ def test_data_for_CNAME(self):
 class TestNs1ProviderDynamic(TestCase):
     zone = Zone('unit.tests.', [])
 
+    _CONTINENT_TO_LIST_OF_COUNTRIES = {
+        'OC': ['FJ', 'NC', 'PG', 'SB', 'VU', 'AU', 'NF', 'NZ', 'FM', 'GU',
+               'KI', 'MH', 'MP', 'NR', 'PW', 'AS', 'CK', 'NU', 'PF', 'PN',
+               'TK', 'TO', 'TV', 'WF', 'WS'],
+    }
```
ross: Better to just refer to this as `Ns1Provider. _CONTINENT_TO_LIST_OF_COUNTRIES` rather than duplicate it here.

tests/test_octodns_provider_ns1.py
```diff
@@ -926,6 +932,43 @@ def test_monitors_gc(self, monitors_for_mock, datafeed_delete_mock,
...
+        ]
+
+        # Set geos to 'OC' in rules[0] (pool - 'lhr')
+        # Check returned dict has list of countries under 'OC'
+        rule0 = self.record.data['dynamic']['rules'][0]
+        saved_geos = rule0['geos']
+        rule0['geos'] = ['OC']
+        ret, _ = provider._params_for_A(self.record)
+        self.assertEquals(sorted(ret['regions']['lhr']['meta']['country']),
+                          sorted(self._CONTINENT_TO_LIST_OF_COUNTRIES['OC']))
```
ross: If the provider stuff is updated to sets you should be able to do something like the following here.

```suggestion
        self.assertEquals(set(ret['regions']['lhr']['meta']['country']),
                          self._CONTINENT_TO_LIST_OF_COUNTRIES['OC'])
```

tests/test_octodns_provider_ns1.py
ross: In tests you'll want to use `self.assert*` rather than bare `assert` so that you get all the unittest features/details.

```suggestion
        self.assertTrue('OC' in data3['dynamic']['rules'][0]['geos'])
```

tests/test_octodns_provider_ns1.py
```diff
@@ -1092,6 +1135,20 @@ def test_data_for_dynamic_A(self):
...
+        ns1_record['regions']['lhr']['meta']['country'] = oc_country_list
+        data3 = provider._data_for_A('A', ns1_record)
+        assert('OC' in data3['dynamic']['rules'][0]['geos'])
+
+        # 2. Partial list of countries should return just those
+        partial_oc_cntry_list = oc_country_list[:5]
+        ns1_record['regions']['lhr']['meta']['country'] = partial_oc_cntry_list
+        data4 = provider._data_for_A('A', ns1_record)
+        for c in partial_oc_cntry_list:
+            assert('OC-{}'.format(c) in data4['dynamic']['rules'][0]['geos'])
```
ross:
```suggestion
            self.assertTrue('OC-{}'.format(c) in data4['dynamic']['rules'][0]['geos'])
```

pavantc:
Addressed all comments @ross. Agree to all of them. (Didn't accept suggestions on the web interface since I needed to make some changes anyway, so limited it to one commit).

PR #508, opened 2020-03-05
Add country based filter to NS1 filter chain

pavantc:
Ns1 needs a country based filter in its filter chain for GeoDNS to work as expected.

/cc @ross
/cc @https://github.com/github/dns/issues/1860
/cc https://github.com/github/dns/pull/1880


octodns/provider/ns1.py
```diff
@@ -192,6 +192,9 @@ class Ns1Provider(BaseProvider):
     }, {
         'config': {},
         'filter': u'geofence_regional'
+    }, {
+        'config': {},
+        'filter': u'geofence_country'
```
ross: I believe this will need to detect the old set and "update" it to the new style in case anyone has already created records with the old filter chains. If this just changes I believe those existing records will cause exceptions when people try to sync anything they created before.

octodns/provider/ns1.py
```diff
@@ -911,9 +945,22 @@ def _params_for_dynamic_A(self, record):
                 }
                 answers.append(answer)
 
+        # Adjust filters as necessary
+        # Make a copy the filters since we might modify it
+        filters = deepcopy(self._DEFAULT_DYNAMIC_FILTERS)
+        if has_country:
+            self.log.debug('Adding country filter')
```
pavantc: This log should be removed.

octodns/provider/ns1.py
```diff
@@ -304,7 +332,9 @@ def _data_for_geo_A(self, _type, record):
 
     def _data_for_dynamic_A(self, _type, record):
         # First make sure we have the expected filters config
-        if self._DYNAMIC_FILTERS != record['filters']:
+        unsupported_fcs = False in [fc in self._SUPPORTED_FILTER_CONFIGS for
+                                    fc in record['filters']]
+        if not record['filters'] or unsupported_fcs:
```
ross: If I'm reading this correctly I believe it won't take order into consideration which we need to. The ordering of the filters is important for octoDNS's dynamic model.

octodns/provider/ns1.py
```diff
@@ -806,13 +837,15 @@ def _params_for_dynamic_A(self, record):
             us_state = set()
 
             for geo in rule.data.get('geos', []):
+
                 n = len(geo)
                 if n == 8:
                     # US state, e.g. NA-US-KY
                     us_state.add(geo[-2:])
```
ross: Do states require the `country` filter too? I don't believe there's a seperate state filter chain item so I'm guessing it's part of the country one, but not sure?

pavantc: Yes, part of the country rules. Will add this as part of the ordering check you commented on above.

octodns/provider/ns1.py
pavantc: Yup, didn't realize the equality check that I replaced was also checking the order. Will address.

octodns/provider/ns1.py
```diff
@@ -298,6 +309,34 @@ def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
...
+    # 1. Filter chain is the basic filter chain
+    # 2. In addition, it could have the 'geofence_country' filter and or
+    # 'geofence_regional' filter only at index 1 in the filter chain
+    # 3. If both country and regional filters are present, they are required
+    # to be as below:
+    #   - 'geofence_regional' at index 1
+    #   - 'geofence_country' at index 2
+    # Any other deviation is returned as an unsupported filter configuration
+    def _valid_filter_config(self, filter_cfg):
+        has_region = 'geofence_regional' in [f['filter'] for f in filter_cfg]
```
ross: Should likely be able to do 

```suggestion
        has_region = self._REGION_FILTER in filter_cfg
```

octodns/provider/ns1.py
```diff
@@ -298,6 +309,34 @@ def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
...
+        has_country = 'geofence_country' in [f['filter'] for f in filter_cfg]
```
ross:
```suggestion
        has_country = self._COUNTRY_FILTER in filter_cfg
```

octodns/provider/ns1.py
```diff
@@ -298,6 +309,34 @@ def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
...
+        return filter_cfg == self._get_updated_filter_chain(has_region,
+                                                            has_country)
+
+    def _get_updated_filter_chain(self, has_region, has_country):
+        filters = deepcopy(self._BASIC_DYNAMIC_FILTERS)
```
ross: Not sure a `deepcopy` is necssary here since you're not modifynig the items themselves, just (maybe) adding/removing items from the list. You can probably just do:

```suggestion
        filters = list(self._BASIC_DYNAMIC_FILTERS)
```

octodns/provider/ns1.py
```diff
@@ -298,6 +309,34 @@ def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
...
+        if has_region:
+            filters.insert(self._REGION_FILTER_INDEX, self._REGION_FILTER)
+        if has_country:
+            if has_region:
+                filters.insert(self._COUNTRY_FILTER_INDEX_WITH_REGION,
+                               self._COUNTRY_FILTER)
+            else:
+                filters.insert(self._COUNTRY_FILTER_INDEX,
+                               self._COUNTRY_FILTER)
```
ross: Might be worth exploring whether it'd be easier to follow if there were 4 lists: base, with_region, with_country, and with_country_and_region which would let you just compare agains the 4 and avoid the inserting and index stuff.

pavantc: > Might be worth exploring whether it'd be easier to follow if there were 4 lists: base, with_region, with_country, and with_country_and_region which would let you just compare agains the 4 and avoid the inserting and index stuff.

Fine by that, changed. Personally wasn't crazy about dealing with specific indices.

octodns/provider/ns1.py
pavantc: Fixed.

octodns/provider/ns1.py
pavantc: Fixed.

octodns/provider/ns1.py
pavantc: Fixed.

octodns/provider/ns1.py
```diff
@@ -238,29 +238,80 @@ class Ns1Provider(BaseProvider):
...
         'config': {},
         'filter': u'geofence_regional'
-    }, {
+    }
+    _COUNTRY_FILTER = {
+        'config': {},
+        'filter': u'geofence_country'
+    }
+
+    _SELECT_FIRST_REGION_FILTER = {
```
ross: In the UI this is called "Select First Group" might be worth a comment about that. It confused me for a second looking over the changes here.

octodns/provider/ns1.py
```diff
@@ -238,29 +238,80 @@ class Ns1Provider(BaseProvider):
...
+
+    _FILTER_CHAIN_WITH_REGION_AND_COUNTRY = [
+        _UP_FILTER,
+        _REGION_FILTER,
+        _COUNTRY_FILTER,
+        _SELECT_FIRST_REGION_FILTER,
+        _PRIORITY_FILTER,
+        _WEIGHTED_SHUFFLE_FILTER,
+        _SELECT_FIRST_N_FILTER
+    ]
```
ross: Nice.

octodns/provider/ns1.py
```diff
@@ -298,6 +349,23 @@ def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
...
+
+    def _get_updated_filter_chain(self, has_region, has_country):
+        if has_region and has_country:
+            filter_chain = self._FILTER_CHAIN_WITH_REGION_AND_COUNTRY
+        elif has_region:
+            filter_chain = self._FILTER_CHAIN_WITH_REGION
+        elif has_country:
+            filter_chain = self._FILTER_CHAIN_WITH_COUNTRY
+        else:
+            filter_chain = self._BASIC_FILTER_CHAIN
```
ross: :cool: like how this turned out.

octodns/provider/ns1.py
pavantc: This is a little confusing.
The ns1 api filter name is select_first_region. From NS1 docs:
```
Select First Region
The select_first_region filter simply returns all answers in the same region as the first answer. There is no metadata input or filter config. In the portal this is called “select first group”.
```
I'll add the above as a comment.

pavantc:
@ross Added changes related to supporting upgrade of filters in extra_changes.

octodns/provider/ns1.py
```diff
@@ -1020,17 +1128,52 @@ def _params_for_SRV(self, record):
...
+        ns1_filters = {}
+        ns1_zone = self._client.zones_retrieve(ns1_zone_name)
+        for ns1_record in ns1_zone['records']:
+            if ns1_record.get('tier', 1) > 1:
+                # Need to get the full record data for geo records
+                full_rec = self._client.records_retrieve(ns1_zone_name,
+                                                         ns1_record['domain'],
+                                                         ns1_record['type'])
+                if 'filters' in full_rec:
+                    filter_key = ns1_record['domain'] + '.'
```
ross: octoDNS's preferred style is to use format over concatentation.

```suggestion
                    filter_key = '{}.'format(ns1_record['domain'])
```

octodns/provider/ns1.py
```diff
@@ -1020,17 +1128,52 @@ def _params_for_SRV(self, record):
...
         for record in desired.records:
             if record in changed or not getattr(record, 'dynamic', False):
                 # Already changed, or no dynamic , no need to check it
                 continue
 
+            # Filter normalization
+            # Check if filters for existing domains need an update
+            # Needs an explicit check since there might be no change in the
+            # config at all. Filters however might still need an update
+            domain = '.'.join([record.name, record.zone.name])
```
ross: Seems cleaner to just do a format here rather than building an array and joining.

```suggestion
            domain = '{}.{}'format(record.name, record.zone.name)
```

octodns/provider/ns1.py
```diff
@@ -1020,17 +1128,52 @@ def _params_for_SRV(self, record):
...
+            if domain in ns1_filters:
+                domain_filters = ns1_filters[domain]
+                if not self._disabled_flag_in_filters(domain_filters, domain):
+                    # 'disabled' entry absent in filter config. Need to update
+                    # filters. Update record
```
ross: Might be nice to add a log in here similar to the ones below, e.g. https://github.com/github/octodns/pull/508/files#diff-0603f6b05b16c9a40a37ca146e6d686eR1182, when we add to `extra`.

PR #509, opened 2020-03-05
NS1 provider: support rate-limiting strategy

rupa:
Adds a "parallelism" argument to the NS1 Provider. If set, we analyze
response headers and attempt to avoid 429 responses.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -7,7 +7,7 @@ from __future__ import absolute_import, division, print_function, \
 
 from logging import getLogger
 from itertools import chain
-from collections import OrderedDict, defaultdict
+from collections import Mapping, OrderedDict, defaultdict
 from ns1 import NS1
 from ns1.rest.errors import RateLimitException, ResourceException
 from pycountry_convert import country_alpha2_to_continent_code
@@ -27,9 +27,11 @@ class Ns1Exception(Exception):
 class Ns1Client(object):
     log = getLogger('NS1Client')
 
-    def __init__(self, api_key, parallelism=None, retry_count=4):
-        self.log.debug('__init__: parallelism=%s, retry_count=%d', parallelism,
-                       retry_count)
+    def __init__(self, api_key, parallelism=None, retry_count=4,
+                 client_config=None):
+        self.log.debug('__init__: parallelism=%s, retry_count=%d, '
+                       'client_config=%s', parallelism, retry_count,
+                       client_config)
         self.retry_count = retry_count
 
         client = NS1(apiKey=api_key)
@@ -60,7 +62,12 @@ class Ns1Client(object):
         # we get the full list of records.
         client.config['follow_pagination'] = True
 
-        self._config = client.config
+        # additional options or overrides
+        if isinstance(client_config, Mapping):
+            for k, v in client_config.items():
+                client.config[k] = v
+
+        self._client = client
 
         self._records = client.records()
         self._zones = client.zones()
@@ -203,14 +210,26 @@ class Ns1Provider(BaseProvider):
     Ns1 provider
 
     ns1:
+        # Required
         class: octodns.provider.ns1.Ns1Provider
         api_key: env/NS1_API_KEY
-        # Optional, to avoid 429s from rate-limiting. Try setting to the
-        # value of max_workers.
-        parallelism: 11
         # Only required if using dynamic records
         monitor_regions:
           - lga
+        # Optional. Default: None. If set, back off in advance to avoid 429s
+        # from rate-limiting. Generally this should be set to the number
+        # of processes or workers hitting the API, e.g. the value of
+        # `max_workers`.
+        parallelism: 11
+        # Optional. Default: 4. Number of times to retry if a 429 response
+        # is received.
+        retry_count: 4
+        # Optional. Default: None. Additional options or overrides passed to
+        # the NS1 SDK config, as key-value pairs.
+        client_config:
+            endpoint: my.nsone.endpoint # Default: api.nsone.net
+            ignore-ssl-errors: true     # Default: false
+            follow_pagination: false    # Default: true
     '''
     SUPPORTS_GEO = True
     SUPPORTS_DYNAMIC = True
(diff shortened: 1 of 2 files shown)


rupa:
Hi, here's my initial idea for handling rate limiting in the NS1 provider, wanted to see what you think ASAP, definitely happy to take advice and make changes. A few comments:

* It looks like we only employ `max_workers` in the plan stage, where we're less likely to hit rate limiting issues. After some thought, it seems reasonable to set parallelism == max_workers anyway: it'll guard against 429s in the plan stage, and overall time of updates should be about the same (we'll wait longer than we have to, but be able to burst a few more requests before waiting again). In my testing a `parallelism` of 1 did a pretty good job of saturating the rate limit without hitting 429s.
* It would be nice to access the `max_workers` setting from Manager, but I didn't want to make any disruptive changes without asking first.
* I put maybe too many docs/words in the Ns1Client class. happy to move/edit them, but hope they help with review at least
* I left the retry logic alone, since `parallelism` is optional, but I think it's a nice hedge even if using a rate limiting strategy

rupa:
Ah woops, I forgot to mention enabling pagination support on the commit message. It's off by default in the SDK to avoid breaking changes, but for octodns it feels like it should be on, and I didn't see a reason to make it configurable.

octodns/provider/ns1.py
```diff
@@ -27,11 +27,41 @@ class Ns1Exception(Exception):
...
+        if parallelism is not None:
+            client.config['rate_limit_strategy'] = 'concurrent'
+            client.config['parallelism'] = parallelism
+
+        # The list of records for a zone is paginated at around ~2.5k records,
+        # this tells the client to handle any of that transparently and ensure
+        # we get the full list of records.
+        client.config['follow_pagination'] = True
+
+        self._config = client.config
```
ross: This just here for the purposes of the tests? Can probably just do `client.client.cofig...` and not have to make a ref into things.

octodns/provider/ns1.py
```diff
@@ -234,15 +267,16 @@ class Ns1Provider(BaseProvider):
                'TK', 'TO', 'TV', 'WF', 'WS'},
     }
 
-    def __init__(self, id, api_key, retry_count=4, monitor_regions=None, *args,
-                 **kwargs):
+    def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
+                 parallelism=None, *args, **kwargs):
```
ross: We could potentially look at passing the `manager` object down into the providers when they're created to give them an option to inspect things like `max_workers`. Anything following the `*args, **kwargs` pattern here would just ignore it if it didn't want it. We can explore that another time, this is probably fine for now and we'd probably want it to allow things to be explicitly set for cases where you might be running multiple sync's in parallel etc.

octodns/provider/ns1.py
```diff
@@ -27,11 +27,41 @@ class Ns1Exception(Exception):
...
+        # or less than our `parallelism`, after which, each process will sleep
+        # for the token replenishment interval times parallelism.
+        # For example, if we can make 10 requests in 60 seconds, a token is
+        # replenished every 6 seconds. If parallelism is 3, we will burst 7
+        # requests, and subsequently each process will sleep for 18 seconds
+        # before making another request.
+        # In general, parallelism should match the number of workers.
+        if parallelism is not None:
+            client.config['rate_limit_strategy'] = 'concurrent'
+            client.config['parallelism'] = parallelism
```
ross: Would it ever make sense to use another strategy (with I'd assume other prams) with octoDNS or would it always be `concurrent`?

octodns/provider/ns1.py
rupa: yeah it's "mostly" there for tests, and I meant to point it out to get your input and played it super-safe for now. I'd prefer setting `self.client = client` but since this is a wrapper for the real client I'd actually rather `self._client = client`. It burns me to write tests that rely on an underscore'd variable, but I think it fits with the way the class behaves.

octodns/provider/ns1.py
rupa: there's only one other one in the SDK right now, and I just don't see it ever being better than this one for octodns. As far as i can tell, assuming that we don't get rate limited in the `plan` stage, a `parallelism` of 1 would be as good as it gets for saturating the connection.
for now, I wanted to keep it simple, and i have a hope that as we make improvements this strategy will still be the one that makes the most sense, and they'll come for free with version bumps.

do you have any recommendations on making these sort of provider caveats more visible? We'd be happy to take some responsibility for maintaining ours, as would others, I'm sure.  I thought maybe if you stuck each provider in a their own dir, there could be a README in each one, and github would serve it really nicely :)

octodns/provider/ns1.py
rupa: absolutely agreed

octodns/provider/ns1.py
ross: It was more that if there are other useful options (or likely to be others in the future) it might make sense to expose a lower-level option, like a `client_config={}` param. If there's not anything useful now we can leave that for if/when.

> do you have any recommendations on making these sort of provider caveats more visible? We'd be happy to take some responsibility for maintaining ours, as would others, I'm sure. I thought maybe if you stuck each provider in a their own dir, there could be a README in each one, and github would serve it really nicely :)

The actual plan, though no clue when I'll get a chance to make it happen, is to break up each of the providers into their own repo under an org https://github.com/octodns. That way they can have their own contributors, releases, ...

octodns/provider/ns1.py
ross: Fwiw within octoDNS `_` vars are considered fair game in tests to facilitate this exact situation. They otherwise maintain their standard "private" connotation.

octodns/provider/ns1.py
rupa: Ah yeah, great point. there are indeed a couple of client config options (not explicitily related to rate-limiting), that I'd like to add support for. I was going to make a lil PR for them once this settled, but having a `client_config={}` param would take care of that in a flexible way.
 
We could also handle the current rate limiting approach this way, but since the settings for it are a bit fiddly, I'd prefer to leave the current `parallelism` setting in place. 

I'll have a patch up shortly, I think I can make it pretty tidy

octodns/provider/ns1.py
rupa: nice, Im a fan of that attitude! I'll swap this out for a `self._client = client` reference

octodns/provider/ns1.py
```diff
@@ -27,11 +27,48 @@ class Ns1Exception(Exception):
...
+
+        # The list of records for a zone is paginated at around ~2.5k records,
+        # this tells the client to handle any of that transparently and ensure
+        # we get the full list of records.
+        client.config['follow_pagination'] = True
+
+        # additional options or overrides
+        if isinstance(client_config, Mapping):
+            for k, v in client_config.items():
+                client.config[k] = v
```
rupa: unfortunately `client.config` is not a "real" dict, and we haven't defined a `.update` for it (yet)

rupa:
🎉  thanks @ross for the help and quick review on this. much appreciated!

ryangsteele:
Thanks for the hard work @ross and @rupa! Out of curiosity, any idea when the next octodns package with this change will be uploaded to PyPi, @ross?

ross:
> Thanks for the hard work @ross and @rupa! Out of curiosity, any idea when the next octodns package with this change will be uploaded to PyPi, @ross?

Don't have a firm target date at the moment. We're in the middle of some fairly major testing and tweaking of the NS1 `dynamic` support and hoping to get that sorted before cutting the next release. 

In the meantime you can use pipy (and others) to install a speicific SHA if that's of use, e.g. 

```
-e git+https://git@github.com/github/octodns.git@2f0a6959428880687ec470e12946b06fd6df0ec7#egg=octodns
```

ryangsteele:
Fair enough, thanks for the prompt response!

PR #511, opened 2020-03-07
Rename fastdns to edgedns w/backwards compat for now

ross:
This renames the fastdns module (and files) to edgedns to match the current product name. There's a backwards compat shim in place for now (with deprecated working) that should allow exiting usage to continue for a time.

/cc Fixes https://github.com/github/octodns/issues/510 @edglynes
/cc @basirjamil


ross:
> Loolks good. Any testing or other required that we can help with?

Nothing specific. If you want to give things a try running on this branch with a config that uses both the old and new `class: ...` that'd probably be sufficent. The tests added here skind of smoke test that, but something more thorough wouldn't be a bad idea.

PR #518, opened 2020-03-26
Use dict to speed up record search

maikzumstrull:
This dict is created earlier, presumably to speed up this exact lookup, but it was using the list form instead, requiring a linear scan for each item.

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -585,7 +585,7 @@ class CloudflareProvider(BaseProvider):
         changed_records = {c.record for c in changes}
 
         for desired_record in desired.records:
-            if desired_record not in existing.records:  # Will be created
+            if desired_record not in existing_records:  # Will be created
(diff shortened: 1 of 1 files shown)


octodns/provider/cloudflare.py
```diff
@@ -585,7 +585,7 @@ def _extra_changes(self, existing, desired, changes):
         changed_records = {c.record for c in changes}
 
         for desired_record in desired.records:
-            if desired_record not in existing.records:  # Will be created
+            if desired_record not in existing_records:  # Will be created
```
ross: Looks like `existing_records[desired_record]` is fetched down below so this is doing two dict lookups (not that that really matters per-wise :grin:.) Something like the following would probably clean that up.

```suggestion
            existing_record = existing_records.get(desired_record, None)
            if existing_record:  # Will be created
```

ross:
Thanks.

PR #530, opened 2020-04-02
dynamic validation tweak, allow reuse of pool if 2nd occur is a default

ross:
Need to allow reuse of a single pool when it's used once with rules and later as a default, e.g. 

```python
                'rules': [{
                    'geos': ['EU-GB'],
                    'pool': 'one',
                }, {
                    'geos': ['EU'],
                    'pool': 'two',
                }, {
                    'pool': 'one',
                }],
```

This is necessary when you want to pull some of a larger region out and send it to a default. All other cases should still be achievable with a single (merged) set of geos afaik.

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -590,6 +590,11 @@ class _DynamicMixin(object):
                     reasons.append('rule {} missing pool'.format(rule_num))
                     continue
 
+                try:
+                    geos = rule['geos']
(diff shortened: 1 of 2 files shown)


markrichardson725:
Screenshot_2020-04-03-17-34-25

markrichardson725:
Look very closely at the background
#$s

PR #532, opened 2020-04-04
Support reuse of default pool in rules in Ns1Provider

pavantc:
The NS1 provider does not handle reuse of pool names in the rules well. Reuse of pool names is in-general disallowed in the OctoDNS config but for one exception. It is allowed when one of the rules uses the default pool. For an example/rationale for this exception, see https://github.com/github/octodns/pull/530#issue-397730162

/cc #530 
/cc @ross

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -468,8 +468,12 @@ class Ns1Provider(BaseProvider):
         pools = defaultdict(lambda: {'fallback': None, 'values': []})
         for answer in record['answers']:
             # region (group name in the UI) is the pool name
+            # Shadow pools are a mechanism internal to this provider
+            # Skip them when constructing octoDNS records from ns1 records
             pool_name = answer['region']
-            pool = pools[answer['region']]
+            if pool_name.endswith('_shadow'):
+                continue
+            pool = pools[pool_name]
 
             meta = answer['meta']
             value = text_type(answer['answer'][0])
@@ -492,6 +496,11 @@ class Ns1Provider(BaseProvider):
         # examples currently where it would
         rules = []
         for pool_name, region in sorted(record['regions'].items()):
+            # Rules that refer to the default dynamic pool would have
+            # used the shadow pool name. Convert it back to the dynamic
+            # default pool name by stripping the '_shadow' in the pool name
+            if pool_name.endswith('_shadow'):
+                pool_name = pool_name[:-7]
             meta = region['meta']
             notes = self._parse_notes(meta.get('note', ''))
 
@@ -935,6 +944,49 @@ class Ns1Provider(BaseProvider):
             notify_list_id = monitor['notify_list']
             self._client.notifylists_delete(notify_list_id)
 
+    def _add_answers_for_pool(self, answers, default_answers, pool_name,
+                              pool_label, pool_answers, pools, priority):
+        current_pool_name = pool_name
+        seen = set()
+        while current_pool_name and current_pool_name not in seen:
+            seen.add(current_pool_name)
+            pool = pools[current_pool_name]
+            for answer in pool_answers[current_pool_name]:
+                answer = {
+                    'answer': answer['answer'],
+                    'meta': {
+                        'priority': priority,
+                        'note': self._encode_notes({
+                            'from': pool_label,
+                        }),
+                        'up': {
+                            'feed': answer['feed_id'],
+                        },
+                        'weight': answer['weight'],
+                    },
+                    'region': pool_label,  # the one we're answering
+                }
+                answers.append(answer)
+
+            current_pool_name = pool.data.get('fallback', None)
+            priority += 1
+
+        # Static/default
+        for answer in default_answers:
+            answer = {
+                'answer': answer['answer'],
+                'meta': {
+                    'priority': priority,
+                    'note': self._encode_notes({
+                        'from': '--default--',
+                    }),
+                    'up': True,
+                    'weight': 1,
+                },
+                'region': pool_label,  # the one we're answering
+            }
+            answers.append(answer)
+
     def _params_for_dynamic_A(self, record):
         pools = record.dynamic.pools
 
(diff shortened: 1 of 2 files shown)


pavantc:
Test cases (other than coverage tests):

1. Test shadow pool creation:
Push a record with a rule set that reuses default pool. Check:
    - [x] If shadow pool is created with the same answers/config as the default pool, but with a different pool name (with `_shadow`)

2. Test record updates (regression test):
Make a modification to the rule to add something more to 'geos' in the ruleset that reuses the default pool. Sync. Check:
    - [x] updates are being detected correctly
Remove update to the geos made above. Sync. Check:
    - [x] updates are being detected correctly

3. Test deletion/creation of shadow pool:
Remove the rule that reuses the default pool. Sync. Check:
    - [x] Shadow pool should be deleted
Reintroduce the ruleset that reuses the default pool. Sync. Check:
    - [x] Shadow pool should have been created

pavantc:
> `shadow` isn't really a meaningful name. It just indicates hidden/behind the scenes which is correct in terms of the yaml, but not really useful if/when someone sees it in the UI. There's also places referring to roughly the same thing as `dynamic default` which isn't confusing since there are default IPs (the ones in the `values` section.)
> 
> I'd recommend using the name `catchall` throughout variable names, comments, and pool (re-)names rather than shadow. I'd also recommend using two `_` in front of it (e.g. `__catchall` when messing with the pool names to effectively ensure we'll never collide with someone's own naming scheme.
> 
> Sort of related to ^, but more broadly I think there's some extra complications happening here due to trying to rename the one with geos rather than the one without. I think this will be easier if you go through and rename the one without geos rather than the one with. Fairly sure that will let you get rid of the extra loop through rules [here](https://github.com/github/octodns/pull/532/files#diff-0603f6b05b16c9a40a37ca146e6d686eR1002-R1005) and the two variables being used to indicate what pools/rules are needed: `needs_shadow` and `dynamic_default_pool_name` which are almost the same thing. As it is I had to read over the changes here 3 or 4 times to start to understand what's happening.

@ross This was done keeping in mind existing deployments. If we made updates unconditional (like in the case of renaming the default), while that might not cause diffs in existing records, wouldn't new records pushed end up have a different naming scheme? Or is that ok?
If the above seems like reason enough to retain the current logic, how about changing the name from `shadow` to `copy`?

ross:
> This was done keeping in mind existing deployments.

I believe this change will impact things that don't require that sort of handling/degree of care. The changes will be to the rules & pools and will take effect the next time things are applied (for any reason.) 

If this version pulls down things created by the old version they'll at worse show as a mis-match from what's expected (which is accuate) and which will resolve when synced. 

If someone pulls the state created by this code down with what's currently on master it'll show as a mis-match to the config and plan a change that will go in and make things look the way master expects. 

Hard to describe, but basically I think we don't have to worry about that as the normal process of plan & apply will move things back and forth cleanly.

pavantc:
> > This was done keeping in mind existing deployments.
> 
> I believe this change will impact things that don't require that sort of handling/degree of care. The changes will be to the rules & pools and will take effect the next time things are applied (for any reason.)
> 
> If this version pulls down things created by the old version they'll at worse show as a mis-match from what's expected (which is accuate) and which will resolve when synced.
> 
> If someone pulls the state created by this code down with what's currently on master it'll show as a mis-match to the config and plan a change that will go in and make things look the way master expects.
> 
> Hard to describe, but basically I think we don't have to worry about that as the normal process of plan & apply will move things back and forth cleanly.

Cool, I will make the changes.

tests/test_octodns_provider_ns1.py
```diff
@@ -1123,6 +1161,21 @@ def test_data_for_dynamic_A(self):
                     'note': 'from:--default--',
                 },
                 'region': 'iad',
+            }, {
```
pavantc, line 1164: I probably should use provider.CATCHALL_PREFIX here.

octodns/provider/ns1.py
```diff
@@ -994,9 +1055,25 @@ def _params_for_dynamic_A(self, record):
...
+                # (We expect only one catchall per record. Any associated
+                # validation is expected to covered under record validation)
+                catchall_pool_name = pool_name
+                regions['{}{}'.format(self.CATCHALL_PREFIX, pool_name)] = {
+                    'meta': meta,
+                }
+            else:
+                regions[pool_name] = {
+                    'meta': meta,
+                }
```
ross, line 1076: Rather than repeating the regions addition, it can come out of the if/else and the pool_name can just be updated.

```suggestion
                pool_name = '{}{}'.format(self.CATCHALL_PREFIX, pool_name)

            regions[pool_name] = {
                'meta': meta,
            }
```

octodns/provider/ns1.py
```diff
@@ -469,16 +470,25 @@ def _data_for_dynamic_A(self, _type, record):
...
+                value_dict = {
                     'value': value,
                     'weight': int(meta.get('weight', 1)),
-                })
+                }
+                # If we have the original pool name and the catchall pool name
+                # in the answers, they point at the same pool. Add values only
+                # once
+                if value_dict not in pool['values']:
+                    pool['values'].append(value_dict)
```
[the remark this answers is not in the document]
pavantc: Not sure about this one. `values` is part of the pools structure that is being returned and used everywhere else (like in record/__init__.py). There are probably assumptions that `[dynamic]['pools']['values']` should be a list. Would need more testing.

tests/test_octodns_provider_ns1.py
```diff
@@ -1123,6 +1094,7 @@ def test_data_for_dynamic_A(self):
         # Test out a small, but realistic setup that covers all the options
         # We have country and region in the test config
         filters = provider._get_updated_filter_chain(True, True)
+        catchall_pool_name = '{}{}'.format(provider.CATCHALL_PREFIX, 'iad')
```
pavantc, line 1097: Made this change since, the last two times, tests failed due to the changing prefix in the code while the pool name was hardcoded here.

octodns/provider/ns1.py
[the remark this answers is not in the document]
ross: May be no hard assumptions that it's specifically a `list`, it probably just needs to be iterable, which a `set` would satisfy. Not a huge deal, just an option to potentially get rid of a bit of code.

PR #540, opened 2020-05-01
Support PowerDNS 4.3.x

acm1:
We recently upgraded from PowerDNS 4.1.1 to 4.3.0 and ran into a couple issues that this PR addresses:

1. The API now returns 404 when a zone doesn't exist, rather than 422 (captured in https://github.com/github/octodns/issues/437).
2. `INCEPTION-INCREMENT` is no longer a valid value for the `soa_edit_api` metadata field on new zones. `DEFAULT` captures that behavior instead (see https://doc.powerdns.com/authoritative/domainmetadata.html#soa-edit-api and https://doc.powerdns.com/authoritative/dnsupdate.html#soa-edit-dnsupdate-settings)

This PR addresses those issues.

1. The provider will handle status code 404 in addition to 422 to indicate a nonexistent zone.
2. `soa_edit_api` still defaults to `INCEPTION-INCREMENT` but it can be overridden by setting `soa_edit_api` in the provider config.


ross:
> 2\. `soa_edit_api` still defaults to `INCEPTION-INCREMENT` but it can be overridden by setting `soa_edit_api` in the provider config.

Would we want to try and detect the PowerDNS version somehow and auto-pick a the default? Probably still good to allow it to be set explicitly in case someone wants/needs another behavior.

acm1:
> Would we want to try and detect the PowerDNS version somehow and auto-pick a the default?

Good idea. I'll take a look at implementing that this week.

ross:
> Good idea. I'll take a look at implementing that this week.

Cool. I was intending to look into it, but my days have repeatedly gotten away from me so I haven't gotten the chance (and probably won't soon.)

maikelpoot:
There are 2 ways of detecting the version:

1. response headers contains `Server: PowerDNS/4.1.10`
2. api endpoint for server details contains `"version": "4.1.10",`

as option 1 could be altered due to proxies/loadbalancers etc, i would go for option 2.
A get request to `/api/v1/servers/localhost` would return:
```json
{
  "config_url": "/api/v1/servers/localhost/config{/config_setting}",
  "daemon_type": "authoritative",
  "id": "localhost",
  "type": "Server",
  "url": "/api/v1/servers/localhost",
  "version": "4.1.10",
  "zones_url": "/api/v1/servers/localhost/zones{/zone}"
}```
(tested with v4.1.10, 4.2.0 and 4.3.0)

Will take a look if i can conjure up some code magic to PR with acm1's branch :)

**Comment by ross** [comment]
> (tested with v4.1.10, 4.2.0 and 4.3.0)

Tested with 4.0.5 as well.

File: octodns/provider/powerdns.py [hunk 1]
```diff
@@ -165,23 +169,78 @@ def _data_for_SRV(self, rrset):
...
+        if major == 4 and minor >= 3:
+            self.log.debug("configure_for_version: Version >= 4.3")
+            self.soa_edit_api = "DEFAULT"
+            self.check_status_not_found = True
+
     def populate(self, zone, target=False, lenient=False):
         self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
                        target, lenient)
 
+        self.detect_version()
```
ross, line 220: Currently it's detecting the version here in populate, but as far as I can tell only needs the info when things are applying. It's minor, but it's appears to be an unneeded step/round trip and just something else to potentially fail. 

Going to look at making a quick refactor of this logic so that it only pulls the info when it's needed and also explicitly makes sure that it's pull before its needed. Will push that up in a bit.

ross: Half-wrong here. This check is needed when zones are not found during populate. It's not needed when all zones exists though so it's still good to only do it when required.

ross:
@acm1, @maikelpoot

Just pushed up the refactor that reworks things to use "caching" properties. Nice this about this style is that callers don't have to think about it or explicitly "warm" the data by calling something like `detect_version`. They just use the properties `soa_edit_api` and `check_status_not_found` normally and everything happens behind the scenes on-demand. 

Wanted to just do the refactors rather than drop PR comments b/c I felt bad that this PR has sat around for a while already. Please take a look over this and once you all 👍 I'll get it :ship:ed.

maikelpoot:
Thanks for the refactor, much cleaner, kinda forgot that that was an option in python.

👍 from me :)

PR #544, opened 2020-05-13
Update NS1 _REGION_FILTER to include remove_no_georegion in config

ross:
NS1 api appears to be returning `remove_no_georegion` in the `config` for `geofence_regional` now which breaks the filter expectations (but does match what it does for `geofence_country`.

The box is checked in the UI so this probably makes sense. Not sure if the `remove_no_georegion` stuff is a new addition or just newly added to the API, but this change will be required to prevent octoDNS from rejecting the filter chains of NS1 dynamic records.

Return value now is:
```
            'config': {
                'remove_no_georegion': True
            },
            'filter': u'geofence_regional'
```

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -253,7 +253,9 @@ class Ns1Provider(BaseProvider):
 
     def _REGION_FILTER(self, with_disabled):
         return self._update_filter({
-            'config': {},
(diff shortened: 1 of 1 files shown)


ross:
This was an expected change in a recent API update so should be gtg.

Issue #534, opened 2020-04-12
Use Lexicon for DNS manipulation


AnalogJ:
Hey,

You might find https://github.com/AnalogJ/lexicon interesting. Its a python library that provides a standardized API to manipulate DNS records on DNS providers. 
At last count it supports ~60 DNS providers out of the box, and it's used under the hood by certbot.

ross:
Hi @AnalogJ I'd welcome a PR adding support for a `LexiconProvider` or alternatively you could include the octoDNS provider in Lexicon itself and allow people to get access to it by installing both libraries and configuring the class with `lexicon.` 

octoDNS supports "advanced" record types and does A LOT of work beyond just interacting with the provider APIs to make disparate providers behave in a similar manner so it's not really something that octoDNS could just be switched to wholesale. 

It would likely be super useful to have a `LexiconProvider` that easily allowed users to configure any of its supported providers, e.g.:

```
---
providers:
    ...

    godaddy:
        class: octodns.provider.lexicon.LexiconProvider
        provider: godaddy
        ...

```

/cc https://github.com/github/octodns/issues/28 which this is very similar to.

doddo:
Hello! I took it upon myself to create such a provider. Whilst working I have gained a more throughout understanding of how Lexicon works  and will update the PR shortly.

doddo:
@AnalogJ I read here:
https://github.com/AnalogJ/lexicon/blob/master/SPECIFICATION.md

quoting the part about update:

> Normal Behaviour Update a record. Record to be updated can be specified by providing id OR name, type and content

But for example in the powerdns [_update](https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/powerdns.py#L214) method it looks like this ie identifier mandatory:

     def _update_record(self, identifier, rtype=None, name=None, content=None):
        self._delete_record(identifier)
        return self._create_record(rtype, name, content)

Furthermore the [base class](https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/base.py#L96) for all lexicon providers states the following:

        """
        Update a record. Identifier must be specified.
        """

This vexes me somewhat. It does seem like some providers makes the assumption that identifier is mandatory for certain operations, but the specification indicates it as an optional field (when listing records)

Now, my approach to this was just to parse whatever information I can from the various providers _list_records method and pass this on to the update / delete ones. The providers I have looked at seems to handle this and to me therefore this seems to be the best approach.

[...]

Issue #476, opened 2020-01-23
feature request: gandi livedns support


resmo:
See https://doc.livedns.gandi.net/

yzguy:
Reference: https://github.com/tiramiseb/go-gandi

jleroy:
See #619.

resmo:
closed by #619

Issue #386, opened 2019-08-05
Feature request for Aliyun(WanWang) as DNS provider


shixinyu:
Aliyun(WanWang) as the biggest Domain Register Provider in China, please consider support them.

ross:
Hi @shixinyu I'd welcome a PR that adds support for them, but it's unlikely that I'll have time to implement it myself. At this point the majority of the providers have been added by others and it's generally best for them to be built by someone who works with the provider and plans to use it going forward. https://github.com/github/octodns/blob/master/octodns/provider/dnsimple.py is one of the simpler providers and would likely serve as a good example.

ross:
There is now a template repo and process to ease this sort of work, https://github.com/octodns/octodns-template. https://github.com/octodns/octodns-powerdns is a good example for simple providers with good apis.

Issue #365, opened 2019-05-24
Request for feature: GoDaddy support


nick-george:
As the title suggests, I'd like to request support for GoDaddy to be added to what appears to be an awesome tool. 

Many thanks, 
Nick

ross:
Hi @nick-george thanks for the suggestion. I won't likely have bandwidth to look into this, but others may. PRs are always welcome. If you or anyone else wants to look into adding GoDaddy support https://github.com/github/octodns/blob/c313ae38aefdfc62b688cefa4d066fb0c07f9d64/octodns/provider/dnsimple.py#L82 is pretty straightforward example to work off of.

nick-george:
Thanks @ross. I'd love to have the time to work on this, but unfortunately I don't. Would you like me to close the ticket or leave it here open? 

Cheers,
Nick

ross:
It can stay open in case someone wants to pick it up and so that anyone else who wonders about GoDaddy can find it.

nick-george:
Thanks.

ross:
There is now a template repo and process to ease this sort of work, https://github.com/octodns/octodns-template. https://github.com/octodns/octodns-powerdns is a good example for simple providers with good apis.

PR #546, opened 2020-05-17
Add Lexicon provider

doddo:
Add Lexicon provider per #534 which could potentially solve gandi
support #476, GoDaddy support #365 and maybe Aliyun(WanWang) #386 to name a few
by using this "wrapper" for Lexicon (which has support for these)

All record types possible are not added yet, and although test coverage is 100%, I have yet to assert enough data is correct between the various transformations, but I figure I'd open this PR up for review, and will complete the rest meanwhile, as they'd not introduce any significant changes.

Another caveat is that I have only tried this one with the gandi Lexicon provider.


doddo:
I put this one into third party library instead. It can be found here: https://github.com/doddo/octodns-lexicon

PR #551, opened 2020-05-29
Cloudflare: Add Support for Rate Limit

lancehudson:
I added basic support for when a 429 HTTP error is returned.
I defaulted to 4 retries with a backoff of 300s.

I copied most of the implementation from the old NS1 implementation
Cloudflare does not return headers to warn you of an upcoming rate limit.

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -104,7 +104,7 @@ class CloudflareProvider(BaseProvider):
         self._zones = None
         self._zone_records = {}
 
-    def _try(self, *args, **kwargs):
+    def _try_request(self, *args, **kwargs):
         tries = self.retry_count
         while True:  # We'll raise to break after our tries expire
             try:
@@ -141,7 +141,8 @@ class CloudflareProvider(BaseProvider):
             page = 1
             zones = []
             while page:
-                resp = self._try('GET', '/zones', params={'page': page})
+                resp = self._try_request('GET', '/zones',
+                                         params={'page': page})
                 zones += resp['result']
                 info = resp['result_info']
                 if info['count'] > 0 and info['count'] == info['per_page']:
@@ -250,7 +251,7 @@ class CloudflareProvider(BaseProvider):
             path = '/zones/{}/dns_records'.format(zone_id)
             page = 1
             while page:
-                resp = self._try('GET', path, params={'page': page})
+                resp = self._try_request('GET', path, params={'page': page})
                 records += resp['result']
                 info = resp['result_info']
                 if info['count'] > 0 and info['count'] == info['per_page']:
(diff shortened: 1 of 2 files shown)


octodns/provider/cloudflare.py
```diff
@@ -85,11 +97,27 @@ def __init__(self, id, email=None, token=None, cdn=False, *args, **kwargs):
...
             })
         self.cdn = cdn
+        self.retry_count = retry_count
+        self.retry_period = retry_period
         self._sess = sess
 
         self._zones = None
         self._zone_records = {}
 
+    def _try(self, *args, **kwargs):
```
ross, line 107: Minor, but what do you think about calling this `_try_request`? It seems a little more descriptive of what it's doing since `_try` is super generic.

octodns/provider/cloudflare.py
```diff
@@ -85,11 +97,27 @@ def __init__(self, id, email=None, token=None, cdn=False, *args, **kwargs):
...
+            try:
+                return self._request(*args, **kwargs)
+            except CloudflareRateLimitError:
+                if tries <= 1:
+                    raise
+                tries -= 1
+                self.log.warn('rate limit encountered, pausing '
+                              'for %ds and trying again, %d remaining',
+                              self.retry_period, tries)
+                sleep(self.retry_period)
```
ross, line 119: Does CF return any further info in the 429 response, suggestions on back off time or anything? If so i'd be nice to utilize it, if not then a generic sleep seems fine. I see their docs say the limits are 1200/5m so guess maybe it's just a generic period and you're following what the docs say should be enough.

octodns/provider/cloudflare.py
lancehudson: Sure, incoming update

octodns/provider/cloudflare.py
lancehudson: No, their entire response is

```
{
	"success": false,
	"errors": [{
		"code": 10100,
		"message": "More than 1200 requests per 300 seconds reached. Please wait and consider throttling your request speed"
	}],
	"messages": [],
	"result": null
}
```

Nothing in the headers

octodns/provider/cloudflare.py
```diff
@@ -18,7 +19,7 @@ class CloudflareError(Exception):
     def __init__(self, data):
         try:
             message = data['errors'][0]['message']
-        except (IndexError, KeyError):
+        except (IndexError, KeyError, TypeError):
```
ross, line 22: What resulted in needing the `TypeError` here?

lancehudson: I added it to catch responses that were not valid json. Example returning `boo` instead of `{}`. When I copied the tests from NS1, it seemed like a good case to handle. I can see how it is only semi related and I can remove it this PR if that makes sense.

ross: Seems fine, just wanted to make sure it was intentional.

ross:
Thanks.

PR #562, opened 2020-06-03
Dhanks geo branch

danhanks:


ross:
Please update the title of this PR and body to talk about adding support for CA providences so that it'll show up if/when people search the history for it. Otherwise lgtm and will merge.

PR #565, opened 2020-06-11
Cloudflare: Increase page size

lancehudson:
Increase Cloudflare page size to reduce request count

`GET zones` has a MAX of 50 and a default of 20
https://api.cloudflare.com/#zone-list-zones

`GET zones/:zone_identifier/dns_records` has a MAX of 100 and a default of 20
https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records


lancehudson:
My biggest worry with this is, what is the chances of Cloudflare lowering the limit?

ross:
> My biggest worry with this is, what are the chances of Cloudflare lowering the limit?

Hopefully low.

Main suggestion would be to make it a parameter to the provider that defaults to 50. That way it can be overridden in the yaml provider config. That'll let people tune it if they feel the need and mitigate problems without code changes if CF ever did lower the value.

lancehudson:
Should I then make it so it then defaults to not including the parameter unless overridden in the provider?

ross:
> Should I then make it so it then defaults to not including the parameter unless overridden in the provider?

I was thinking of something like `timeout` in https://github.com/github/octodns/blob/master/octodns/provider/powerdns.py#L22-L23. It'd have a default value of 50 as you propose in the provider's `__init__` which gets stored in a property `per_page` on `self`. Then when calls are made to get zones and records `per_page: self.per_page` is used.

PR #571, opened 2020-06-17
Change continent code for Timor Leste

danhanks:


ross:
Fixes https://github.com/github/octodns/issues/570

PR #581, opened 2020-07-09
Adding Octodns provider class for easyDNS

actazen:
This provider class for easydns.com adds support for basic dns records
through the easyDNS v3 API.  Support for dynamic and geo based dns
records is planned for a future update.

Sample configuration for the easyDNS provider are:

```
easydns:
     class: octodns.provider.easydns.EasyDNSProvider
     token: <token>
     apikey: <key>
```

The token and key values are found on the easyDNS customer portal at:

https://cp.easydns.com/manage/security/api/production_info.php

Also, below are some optional configuration parameters which can be
added to override the class defaults.  By default the provider class
connects with the LIVE easyDNS API, if you wish to perform testing
with the easyDNS Sandbox API you can enable it by adding the following
configuration parameter:

     sandbox: True

Note, the API token and key are different for the sandbox than they
are for the production API, you can obtain sandbox credentials at:

https://cp.easydns.com/manage/security/api/sandbox_info.php

Lastly, if you have created Domain Portfolios through the easyDNS CP
you can configure which portfolio new domains will be added to by
supplying the portfolio option with the name of your portfolio.

     portfolio: <portfolio name>


octodns/provider/easydns.py
```diff
@@ -0,0 +1,453 @@
...
+    # Default Currency CAD
+    defaultCurrency = 'CAD'
+    # Domain Portfolio
+    domainPortfolio = 'myport'
+
+    def __init__(self, token, apikey, currency, portfolio, sandbox):
+        self.log = logging.getLogger('EasyDNSProvider[{}]'.format(id))
+        self.token = token
+        self.apikey = apikey
+        self.defaultCurrency = currency
```
ross, line 54: octoDNS uses snake_case, rather than camelCase. Please make a pass through and convert things for consistency with the existing codebase.

octodns/provider/easydns.py
```diff
@@ -0,0 +1,453 @@
...
+        self._request('PUT', path, data=domainData).json()
+
+        # EasyDNS creates default records for MX, A and CNAME for new domains,
+        # we need to delete those default record so we can sync with the source
+        # records, first we'll sleep for a second before gathering new records
+        # We also create default NS records, but they won't be deleted
+        sleep(1)
+        records = self.records(name, True)
+        for record in records:
+            if record['type'] in ('A', 'MX', 'CNAME'):
```
ross, line 103: Would it make sense to check the `host` part of the records too here just to be double sure it's deleting the things it expects to be?

octodns/provider/easydns.py
```diff
@@ -0,0 +1,453 @@
...
+            if record['host'] == '@':
+                record['host'] = ''
+
+            # change any apex value to zone name
+            if record['rdata'] == '@':
+                record['rdata'] = '{}.'.format(zone_name)
+
+            # change "URL" & "STEALTH" to a "CNAME"
+            if record['type'] == "URL" or record['type'] == "STEALTH":
+                record['type'] = 'CNAME'
```
ross, line 130: If `URL` is for redirects to arbitrary URLs I'm not sure a `CNAME` will work here since it'd lose any path and scheme information from the URL. Other providers just ignore `URL` style records for this reason.

What are `STEALTH` records? Are they similar to `ALIAS` in other providers, i.e. the DNS resolvers do the lookup for the target and then return its value(s)?

octodns/provider/easydns.py
```diff
@@ -0,0 +1,453 @@
...
+class EasyDNSProvider(BaseProvider):
+    '''
+    EasyDNS provider using API v3
+
+    easydns:
+        class: octodns.provider.easydns.EasyDNSProvider
+        # Your EasyDNS API token (required)
+        token: foo
+        # Your EasyDNS API Key (required)
+        apikey: bar
```
ross, line 159: apikey would likely be more consistent with other providers as `api_key`, but can stay this way if easy dns refers to it as a single word/combined like that.

octodns/provider/easydns.py
actazen: I appreciate the feedback.  I've removed the camelCase (bad habit's sometimes sneak in) and will commit an update shortly.

octodns/provider/easydns.py
actazen: Good idea just in case the API author decides to make some changes without informing me.  This has been added to the next commit.

octodns/provider/easydns.py
actazen: While EasyDNS's URL and STEALTH records do result in being sent to a specific URL, their underlying type is actually a CNAME (if you were to dig the DNS record it would return as a CNAME) which points to an EasyDNS host which then performs a redirect to the client's specified URL, so nothing is lost.  A STEALTH record is the same as a URL but the resulting URL is hidden from the browser (when possible).  

If you believe this would cause confusion with how other providers handle such records, I can ask the API author to change the record type during parsing so they come through as CNAME records and remove the related code from the provider class.

octodns/provider/easydns.py
actazen: No problem, happy to keep it consistent.

octodns/provider/easydns.py
ross: Mostly concerned with what the `value` would be. This translation would result in a `CNAME` in the octoDNS config, but there's nothing that I see that would pull out the hostname part of that and record that in the config. 

Also there'd be no way afaict to edit these translated records. If you change the `CNAME` `value` in the config it'd try and push a `CNAME` `value` up to the API in a place that previously held a `STEALTH` or `URL` record. 

I believe octoDNS will need to ignore these types of records as it does for `URL` in other providers.

actazen: No problem, I've updated the code to remove the URL and STEALTH record conversions.

octodns/provider/easydns.py
```diff
@@ -0,0 +1,450 @@
...
+                record['host'] = ''
+
+            # change any apex value to zone name
+            if record['rdata'] == '@':
+                record['rdata'] = '{}.'.format(zone_name)
+
+            # EasyDNS supports DYNamic A records so we'll convert these
+            # to their underlying DNS record type before processing
+            if record['type'] == "DYN":
+                record['type'] = 'A'
```
ross, line 130: Sorry didn't notice this one under the URL and STEALTH types before. What's the behavior of `DYN` records? Will it have similar problems to STEALTH and URL in terms of losing information and being impossible to manage?

actazen: Sorry I overlooked that too, it's been removed, just pushed the update.

PR #585, opened 2020-07-16
Adding environment variable record injection

phelpsw:
Per the discussion on https://github.com/github/octodns/issues/583 here is a work in progress of environment variable injection for discussion.

### DIFF
--- a/octodns/source/envvar.py
+++ b/octodns/source/envvar.py
@@ -1,0 +1,105 @@
+
+import logging
+import os
+
+from ..record import Record
+from .base import BaseSource
+
+
+class EnvVarSourceException(Exception):
+    pass
+
+
+class EnvironmentVariableNotFoundException(EnvVarSourceException):
+    def __init__(self, data):
+        super(EnvironmentVariableNotFoundException, self).__init__(
+            'Unknown environment variable {}'.format(data))
+
+
+class EnvVarSource(BaseSource):
+    '''
+    This source allows for environment variables to be embedded at octodns
+    execution time into zones.  Intended to capture artifacts of deployment to
+    facilitate operational objectives.
+
+    The TXT record generated will only have a single value.
+
+    The record name cannot conflict with any other co-existing sources.  If
+    this occurs, an exception will be thrown.
+
+    Possible use cases include:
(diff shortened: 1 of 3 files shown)


octodns/source/envvar.py
```diff
@@ -0,0 +1,105 @@
...
+                       self.envvar, self.value)
+
+    def populate(self, zone, target=False, lenient=False):
+        self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
+                       target, lenient)
+
+        # if target:
+        # TODO: Environment Variable Source cannot act as a target,
+        # throw exception?
+        # return
```
ross, line 91: Been a couple years since I looked at it closely/wrote it, but don't think you have to worry about target=True in sources. It can just be ignored.

octodns/source/envvar.py
```diff
@@ -0,0 +1,105 @@
...
+
+        before = len(zone.records)
+
+        self._read_variable()
```
ross, line 95: I think it'd be a little cleaner to have the read return the value rather than set it to a property when this is called and then access it down below.

octodns/source/envvar.py
```diff
@@ -0,0 +1,105 @@
...
+          - ultra
+          - ns1
+    '''
+    SUPPORTS_GEO = False
+    SUPPORTS_DYNAMIC = False
+    SUPPORTS = set(('TXT'))
+
+    DEFAULT_TTL = 60
+
+    def __init__(self, id, variable, record, ttl=DEFAULT_TTL):
```
ross, line 65: Minor, but I think i'd prefer `name` here rather than record since record is a class/object type in a lot of octoDNS's code it'll make it clearer what's expected here.

octodns/source/envvar.py
phelpsw: Great

octodns/source/envvar.py
phelpsw: Agreed

octodns/source/envvar.py
phelpsw: Makes sense

Issue #586, opened 2020-07-16
AzureDNS: Problem converting a CNAME record to an A record 


Arunothia:
Similar to #507, facing trouble creating an A record in place of a CNAME record in AzureDNS.

Example Scenario: 

a:
  ttl: 3600
  type: CNAME
  value: sample.com.

change the above to - 

a:
  ttl: 3600
  type: A
  value: 1.1.1.1

This leads to the following error - 

* Create <ARecord A 3600, a.zone.com, ['1.1.1.1']> (config)
* Delete <CnameRecord CNAME 3600, a.zone.com, sample.com.>

msrestazure.azure_exceptions.CloudError: 409 Client Error: Conflict for url: <>

The error appears to be occurring because the create of A record is attempted before the delete of the CNAME record.

ppieprzycki:
I would add that the same issue is with converting  NS record into CNAME. Probably even more cases exist for this issue.

ross:
There is a stalled PR for this issue https://github.com/octodns/octodns/pull/587.  The Azure provider isn't currently well owned/maintained and it's not something I use/have access set up to so it'll probably have to fall to someone else to dig into this and address the problem in the near term.

Arunothia:
> I would add that the same issue is with converting NS record into CNAME. Probably even more cases exist for this issue.
FYI: @ross 

I have updated the PR to force all deletes to happen first before other operations. 
Tested for the following scenarios -

1. Update CNAME to A
2. Update A to CNAME
3. Update NS to CNAME
4. Update CNAME to NS

Arunothia:
PR for forcing delete operation to happen before other operations in AzureDNS has gone through.

PR #587, opened 2020-07-16
Forcing delete to happen before create

Arunothia:
Trying to resolve the issue #586 by reversing the apply operation order.


Arunothia:
Thank you Ross, makes sense. I will enforce the ordering.

ppieprzycki:
Any plans to proceed with this PR. As well have an issue when I need change CNAME record into A. (in other direction this is not visible)

Arunothia:
> Any plans to proceed with this PR. As well have an issue when I need change CNAME record into A. (in other direction this is not visible)

@ppieprzycki I see, thanks for this input. I got delayed in getting back to this as we no longer are dependent on this feature. I will try to get back to this in the next couple of days, will be good to have this completed.

octodns/provider/azuredns.py
```diff
@@ -488,6 +488,19 @@ def _apply(self, plan):
...
+        Force the operation order to be Delete() before all other operations.
+        Helps avoid problems in updating
+            - a CNAME record into an A record.
+            - an A record into a CNAME record.
+        '''
+
+        for change in changes:
+            class_name = change.__class__.__name__
+            if class_name == 'Delete':
+                getattr(self, '_apply_{}'.format(class_name))(change)
```
ross, line 501: No need to dynamically look up the apply method since there's an `if` above looking for a specific one.

```suggestion
                self._apply_Delete(change)
```

Arunothia: That makes sense, thanks, made the update!

ross (approved):
Long term I think we should look at adding ordering support to the `Plan` class so that `changes` are ordered in a consistent manner, e.g. deletes before adds, so that providers are more likely to do the right thing without having to take special care like this.

ross:
Filed https://github.com/octodns/octodns/issues/656.

PR #589, opened 2020-07-21
octodns-compare: Prefix filtering and status code on on mismatch

phelpsw:
Adding a non-zero error code when changes exist.  Also adding an --ignore-prefix option to filter out records from this comparison which are expected to mismatch.


phelpsw:
Hi @ross , can we get this merged?

PR #590, opened 2020-07-23
Root NS record support

acm1:
This is built on https://github.com/github/octodns/pull/434. I'm opening this PR to test mergeability, but I've also opened https://github.com/joschi36/octodns/pull/6 so that @joschi36 can merge it there if desired.


gildegoma:
Hi @acm1 @ross @joschi36 !
This is great! We were also working on reactivating this change request "ASAP" 😄.
 
In the meantime, we (@swisstxt) have rebased onto the upstream master and integrated Joschi's changes with some Ross' comments on #434. See https://github.com/swisstxt/octodns-2/tree/feature/root-nameservers.

@acm1 if you agree, we could _restart_ from this new branch (single commit so far) and add your changes. Up to you if you prefer to
- pull it on your side, and open (yet another) pull request, or
- open a PR against https://github.com/swisstxt/octodns-2 and let us deal to get all this merged into the upstream project 😉

gildegoma:
Sorry, I updated the above comment.

@ross please let us know if you're okay to review+merge #434 or #590, or if you need/appreciate a cleaner+rebased feature branch.

istr:
@acm1 you could set `SUPPORTS_ROOT_NS = True` for `digitalocean.py` provider.
I tried it and it works out-of-the-box with this PR.

ross:
> @ross please let us know if you're okay to review+merge #434 or #590, or if you need/appreciate a cleaner+rebased feature branch.

Swamped at the moment and haven't been able to string together enough time to look at them. 

> @acm1 you could set `SUPPORTS_ROOT_NS = True` for `digitalocean.py` provider.
> I tried it and it works out-of-the-box with this PR.

I'd like to see the ability to manage ROOT NS records and Dyn was always one of the problematic providers (and it's 💀 soon) so 🤞 we can find a path here. `SUPPORTS_ROOT_NS` is probably part of that.

istr:
Another finding with the ns1 provider that I am currently unable to find a good fix for is this:
NS handling works, _but not right after creation of new zones_.
This is what happens

1. the plan to create a new zone is added
1. the plan to _insert_ a new NS entry is added
1. the provider ns1 creates the zone
1. the provider ns1 automatically creates a new NS entry with the default servers
1. now the plan fails, because we need an _update_ of the NS entries instead of a _create_

There are two possible ways to remedy the situation, both would need to re-scan / re-check the provider's state:
1. always make zone creations the first items in a plan; then re-check the state (re-scan the inventory); then build the plan for new/changed entries
1. after each zone creation that is followed by NS manipulation re-check immediately before the planned NS items and change a _create_ plan to a _update_ plan when necessary

Another option that would need more research would be to create new zones at the ns1 provider _always_ as secondary zones (this suppresses the automatic creation of the NS entries, as per documentation). But I am unsure if this could be done for zones that are meant to be primary.

I could dig deeper into this, but I am not yet deeply acquainted with the workflow within the provider logic (plan building, hooks, actions etc.), so I could not tell which of these options is the better one (less code / easier to implement / better to read).

ramnes:
Just a quick word here to say that we also needed to modify Route 53 root NS for a couple of zones, and this branch worked just as expected.

josvo:
@ross Any chance to get this merged? I think @joschi36 and @acm1 did a great job here! Would be great to have that in an official release.

ross:
> @ross Any chance to get this merged? I think @joschi36 and @acm1 did a great job here! Would be great to have that in an official release.

I don't believe we got this to a point where it was safe/robust enough to ship. As you can see from the discussion in this PR and previous exploration when I was originally looking into it back in the early days of octoDNS is that root NS management is a mess and every provider has different hiccups and behaviors.

istr:
> Swamped at the moment and haven't been able to string together enough time to look at them.

With the split provider approach maybe this could be re-activated and much of the work needed could be offloaded to separate provider maintainers (I see a fair chance that those are each sponsored by their respective provider company; even if the support boils down to some free accounts for real integration testing...)

The diff in `octodns/manager.py` and `octodns/provider/base.py` are relatively small and do no real harm when there is an implicit `False` default set on the additional property.
With only a handful of all providers having `dynamic` support there is no good argument left, not to provide this option to those who want it, need it, and are willing and able to implement it in their favorite providers.

ross:
> With the split provider approach maybe this could be re-activated and much of the work needed could be offloaded to separate provider maintainers (I see a fair chance that those are each sponsored by their respective provider company; even if the support boils down to some free accounts for real integration testing...)

Yeah. That'd be the hope/dream, but I'll be a little surprised if it comes true.

Would definitely appreciate free accounts/credits for testing. NS1 is kind enough to provide me with them and it does make it much easier to work on & test their provider.  No one provider is particularly expensive to test/mess with, but when you add all of them up it's not something I'm OK with paying out of pocket :grin:

> The diff in `octodns/manager.py` and `octodns/provider/base.py` are relatively small and do no real harm when there is an implicit `False` default set on the additional property.
> With only a handful of all providers having `dynamic` support there is no good argument left, not to provide this option to those who want it, need it, and are willing and able to implement it in their favorite providers.

Yeah. I think this should now be more doable with the use of [`_process_desired_zone`](https://github.com/octodns/octodns/blob/master/octodns/provider/base.py#L33). I've never liked that apex `NS` records were a special case and ignored differently than everything else https://github.com/octodns/octodns/blob/master/octodns/zone.py#L27-L30. I'd like to see that go away, but there'd be some details to sort out to get that to happen and to have the provider handle including/ignoring them appropriately. 

It might be something I dig into once all the extract stuff has completed.

acm1:
https://github.com/octodns/octodns/pull/876 has been merged (😃 ) superseding this.

Issue #591, opened 2020-07-28
Assigning multiple zones to the same zone file


jleroy:
I have a lot of domains who uses the same zone file. Is the a way to link a bunch of domains to a zone without adding an entry under `zones` for each domain?

I'm thinking of something like this:
```
zones:
  example.com.:
  sources:
    - config
  targets:
    - dnsimple
  zones:
    - example.net.
    - example.org.
```

yzguy:
You could create symlinks on the filesystem.

```
example.net.yaml -> example.com.
example.org.yaml -> example.com.
```

brianeclow:
I wrote a sorting script that pulls in the octoDNS sort and yaml formatting. Which when you run against this type of configuration, starting with

```
# Arbitary hash in config
zoneconfig: &zoneconfig
  sources:
    - config
  targets:
    - dnsimple

zones:
 example.com.:
   <<: *zoneconfig
 example.net.:
   <<: *zoneconfig
```

results in

```
# Arbitary hash in config
zoneconfig:
  sources: &id001
    - config
  targets: &id002
    - dnsimple

zones:
 example.com.:
  sources: *id001
  targets: *id002
 example.net.:
  sources: *id001
  targets: *id002
```

jleroy:
I've submitted a pull request adding support for zones aliases (#592).

PR #592, opened 2020-08-02
Add support for zones aliases

jleroy:
This commit adds support for zones aliases. This allows to define one or multiple zone as aliases of an existing zone without using workarounds like simlinks and miltiple `zones:` entries in the configuration file.

An alias zone share all of its content with it parent zone, only the name of the zone is different:

```
zones:
  example.com.:
    aliases:
      - example.net.
      - example.org.
    sources:
      - in
    targets:
      - out
```

Known issues:

- [x] ~~No documentation~~
- [x] ~~Only the `octodns-sync` and `octodns-validate` commands supports aliases zones at this time~~

I added a loop in the manager init function which convert all alias zone to "real" ones during config validation, however I'm not sure this is the right approach. Comments welcome.


jleroy:
See also #591.

ross:
👍 conceptually, though I think I'd prefer to flip things around and have a zone defined which references it's target, something like:

```yaml
zones:
  example.com.:
    sources:
      - in
    targets:
      - out

  example.net.:
    alias: example.com.

  example.org.:
    alias: example.com.
```

That should work well in terms of being able to search/scan the list of zones or iterate over them normally in code. It'll also be a little cleaner in diffs when things are committed to repos and moved to/from aliases.

Lastly it should allow things to expand on the idea in the future and do things like:

```yaml
  example.org.:
    alias: example.com.
    targets:
      - other
```

Where it has the same base config, but changes or adds to the targets etc. 

I haven't had a chance to look at the PR in detail, but my worry is that all providers will have to support the alias concept to pull things off as they're currently sketched out. That's something I think whatever we'd do here would need to avoid as I'd rather not add to what's required to implement a provider as that will get very involved for the more advanced geo/dynamic providers. This is part of why this concept hasn't been implemented up to now. I feel like it should be possible to have the Manager (possibly in cooperation with the Zone) hide these details from the providers.

jleroy:
Thank you @ross for your detailed comment. I must admit that I have hesitated between your approach and the one I've used in the initial PR.

I've updated the PR to make each alias zone reference its target zone, and also added a `is_alias` property to the `Zone` object which is true if the zone is an alias.

All tests pass, but I still need to look at how to "fix" the others `octodns-*` commands.

jleroy:
The two lasts commits add zones aliases support to `octodns-report` and `octodns-compare` commands. `octodns-dump` was already OK.

jleroy:
Any comments on this PR? Can I start writing documentation for it?

ross:
Sorry, been swamped with non-octoDNS stuff lately. I have this on my list to re-look at, but still haven't made it down to do so.

ross:
### Hiccups

Looks like this is making an assumption that any aliased zone is populated by a single `yaml` file. In reality there are a lot of ways that zone data is sourced, probably 2/3 of our internal zones use `yaml` as the base, but pull data from other custom sources (lists of hosts, services, network devices, ...,) often multiple places. It's also possible to source the data from multiple yaml files which others do.

There's two issues there:
1. Alias won't work unless it's used on a zone with a single yaml file.
1. Populate is run on the aliased zone

The second one of those is a bit tougher to see so maybe an example will help:

```yaml
providers:

providers:
  # Primary source for manually configured data
  config:
    directory: ./config

  services-api:
    class: github.source.sites_api.ServicesApiSource
    token: env/SERVICES_API_TOKEN

zones:

  original.com.:
    sources:
      - config
      - services-api

  mirror.com.:
    alias: original.com.  
```

`ServicesApiSource` details are beyond the scope of a review comment, but the TL;DR is that the source takes the zone name it's asked to populate and makes an api call to get the list of services that are on that domain.

So if we run sync we'll load original's config from the yaml file and then add the records from services-api for original.com. That's all as expected. The problem happens when we get to mirror.com. The manager copied over the `file` the yaml provider is going to use, but it won't know anything about the `services-api` source and thus the alias won't actually have a copy of the data. That's just 1) though. 

The 2nd would come in if we continued down this path and copied over the info that each sources needed to the aliased zone. If we then populated config, and then called into `services-api` we'd be asking it for `mirror.com.` and not `original.com.` and it wouldn't return us the right data (if anything.)

### Solutions

The good news is that the solution here will likely simplify things quite a bit. We probably won't have to touch the `Zone` or `YamlProvider` classes much, if at all.

Instead the manager can run though and populate all the normal zones and keep track of the names of the aliased zones. 

This first pass will populate all the normal zones fully/as always. After that the manager has populated all the normal zones a second pass can be made to do all the aliased ones. That pass just needs to copy the zone record objects from the source into the target and then plan.

Roughly that'd look like:

Having a list of aliased zone names, checking for aliases and adding them to that list [here](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L289) and then skipping the rest of that for

Letting everything else happen normally and then right after we've waited for all the normal populate & plans to finish [here](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L354) a second round of work can be queued up. 

The existing [_populate_and_plan](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L242-L243) can likely be reused and instead of passing in `is_alias` you can pass the the source zone (`plan.desired`) object as an optional param. If that param is `not None` It can [skip the populate from sources](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L250-L258) and just do the `set` copy. It can then plan normally. 

So that'd just be a loop for the aliased zones adding them to a futures array like [this](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L347-L350) passing the `aliased` zone and then another wait for things to finish appending to the list of plans, similar to [this](https://github.com/github/octodns/blob/04a2d7a585630222f5c8894e8c8e898cc1a3e9b2/octodns/manager.py#L354).

jleroy:
Thank you @ross for your detailled feedback.
Regarding your 1., I can't manage to reproduce the issue using the following configuration:

```
manager:
  max_workers: 2

providers:
  config:
    directory: ./config
  
  dnsimple:
    class: octodns.provider.dnsimple.DnsimpleProvider
    account: secrect
    token: secret
    sandbox: true
  
  dnsmadeeasy:
    class: octodns.provider.dnsmadeeasy.DnsMadeEasyProvider
    api_key: secret
    secret_key: secret
    sandbox: true

zones:
  example.com.:
    sources:
      - config
      - dnsimple
    targets:
      - dnsmadeeasy
  
  example.net.:
    alias: example.com.
```

Everything seems to be working fine, or maybe I am missing something?

```
$ octodns-sync --config=config/production.yaml
2020-10-29T22:27:01  [4696542656] INFO  Manager __init__: config_file=config/production.yaml
2020-10-29T22:27:01  [4696542656] INFO  Manager __init__:   max_workers=2
2020-10-29T22:27:01  [4696542656] INFO  Manager __init__:   include_meta=False
2020-10-29T22:27:01  [4696542656] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=True, force=False
2020-10-29T22:27:01  [4696542656] INFO  Manager sync:   zone=example.com.
2020-10-29T22:27:01  [4696542656] INFO  Manager sync:   sources=['config', 'dnsimple'] -> targets=['dnsmadeeasy']
2020-10-29T22:27:01  [4696542656] INFO  Manager sync:   zone=example.net.
2020-10-29T22:27:01  [4696542656] INFO  Manager sync:   sources=['config', 'dnsimple'] -> targets=['dnsmadeeasy']
2020-10-29T22:27:01  [123145392812032] INFO  YamlProvider[config] populate:   found 2 records, exists=False
2020-10-29T22:27:01  [123145409601536] INFO  YamlProvider[config] populate:   found 2 records, exists=False
2020-10-29T22:27:02  [123145392812032] INFO  DnsimpleProvider[dnsimple] populate:   found 17 records, exists=True
2020-10-29T22:27:02  [123145392812032] INFO  DnsMadeEasyProvider[dnsmadeeasy] plan: desired=example.com.
2020-10-29T22:27:02  [123145409601536] INFO  DnsimpleProvider[dnsimple] populate:   found 17 records, exists=True
2020-10-29T22:27:02  [123145409601536] INFO  DnsMadeEasyProvider[dnsmadeeasy] plan: desired=example.net.
2020-10-29T22:27:03  [123145392812032] INFO  DnsMadeEasyProvider[dnsmadeeasy] populate:   found 0 records, exists=False
2020-10-29T22:27:03  [123145392812032] INFO  DnsMadeEasyProvider[dnsmadeeasy] plan:   Creates=16, Updates=0, Deletes=0, Existing Records=0
2020-10-29T22:27:03  [123145409601536] INFO  DnsMadeEasyProvider[dnsmadeeasy] populate:   found 0 records, exists=False
2020-10-29T22:27:03  [123145409601536] INFO  DnsMadeEasyProvider[dnsmadeeasy] plan:   Creates=16, Updates=0, Deletes=0, Existing Records=0
2020-10-29T22:27:03  [4696542656] INFO  Manager
* example.com.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create Zone<example.com.>
*   Create <ARecord A 300, example.com., ['1.2.3.4', '1.2.3.5']> (dnsimple)
*   Create <CaaRecord CAA 3600, example.com., ['0 issue "ca.unit.tests"']> (dnsimple)
*   Create <SrvRecord SRV 600, _srv._tcp.example.com., [''10 20 30 foo-1.unit.tests.'', ''12 20 30 foo-2.unit.tests.'']> (dnsimple)
*   Create <AaaaRecord AAAA 600, aaaa.example.com., ['2601:644:500:e210:62f8:1dff:feb8:947a']> (dnsimple)
*   Create <CnameRecord CNAME 300, cname.example.com., unit.tests.> (dnsimple)
*   Create <CnameRecord CNAME 3600, excluded.example.com., unit.tests.> (dnsimple)
*   Create <MxRecord MX 300, mx.example.com., [''10 smtp-4.unit.tests.'', ''20 smtp-2.unit.tests.'', ''30 smtp-3.unit.tests.'', ''40 smtp-1.unit.tests.'']> (dnsimple)
*   Create <PtrRecord PTR 300, ptr.example.com., foo.bar.com.> (dnsimple)
*   Create <SpfRecord SPF 600, spf.example.com., ['v=spf1 ip4:192.168.0.1/16-all']> (dnsimple)
*   Create <CnameRecord CNAME 1800, test2.example.com., target.example.com.> (dnsimple)
*   Create <CnameRecord CNAME 1800, test3.example.com., target.example.com.> (config)
*   Create <TxtRecord TXT 600, txt.example.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (dnsimple)
*   Create <TxtRecord TXT 600, txt3.example.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Create <ARecord A 300, www.example.com., ['2.2.3.6']> (dnsimple)
*   Create <ARecord A 300, www.sub.example.com., ['2.2.3.6']> (dnsimple)
*   Create <AliasRecord ALIAS 3600, zzalias.example.com., example.com.> (dnsimple)
*   Summary: Creates=16, Updates=0, Deletes=0, Existing Records=0
* example.net.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create Zone<example.net.>
*   Create <ARecord A 300, example.net., ['1.2.3.4', '1.2.3.5']> (dnsimple)
*   Create <CaaRecord CAA 3600, example.net., ['0 issue "ca.unit.tests"']> (dnsimple)
*   Create <SrvRecord SRV 600, _srv._tcp.example.net., [''10 20 30 foo-1.unit.tests.'', ''12 20 30 foo-2.unit.tests.'']> (dnsimple)
*   Create <AaaaRecord AAAA 600, aaaa.example.net., ['2601:644:500:e210:62f8:1dff:feb8:947a']> (dnsimple)
*   Create <CnameRecord CNAME 300, cname.example.net., unit.tests.> (dnsimple)
*   Create <CnameRecord CNAME 3600, excluded.example.net., unit.tests.> (dnsimple)
*   Create <MxRecord MX 300, mx.example.net., [''10 smtp-4.unit.tests.'', ''20 smtp-2.unit.tests.'', ''30 smtp-3.unit.tests.'', ''40 smtp-1.unit.tests.'']> (dnsimple)
*   Create <PtrRecord PTR 300, ptr.example.net., foo.bar.com.> (dnsimple)
*   Create <SpfRecord SPF 600, spf.example.net., ['v=spf1 ip4:192.168.0.1/16-all']> (dnsimple)
*   Create <CnameRecord CNAME 1800, test2.example.net., target.example.com.> (dnsimple)
*   Create <CnameRecord CNAME 1800, test3.example.net., target.example.com.> (config)
*   Create <TxtRecord TXT 600, txt.example.net., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (dnsimple)
*   Create <TxtRecord TXT 600, txt3.example.net., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Create <ARecord A 300, www.example.net., ['2.2.3.6']> (dnsimple)
*   Create <ARecord A 300, www.sub.example.net., ['2.2.3.6']> (dnsimple)
*   Create <AliasRecord ALIAS 3600, zzalias.example.net., example.com.> (dnsimple)
*   Summary: Creates=16, Updates=0, Deletes=0, Existing Records=0
```

ross:
> Everything seems to be working fine, or maybe I am missing something?

When `dnsimple` is synced for the aliased copy it is pulling the records for `example.net.` and not `example.com.` so it's not actually doing what you'd want/expect. `dnsimple` will gladly populate things in that case, but i'll be pulling the records from the wrong thing zone: example.net.



Maybe try setting up an example with 2 `YamlProvider`s. `config1`, and `config2`. Create example.com.yaml in both directories and put different records in each. Don't create example.net.yaml (don't need it, it's supposed to be a copy of example.com.) I believe the aliased zone will have one of the two sources assigned to the `file` key and when things go to populate it'll load the data from that zone twice (once for each YamlProvider) so example.com will be correct, but example.net will only have 1/2 the records.

jleroy:
Yeah you're right, all records are here for the main domain, but two times, so the synchronization fail with a duplicate record error: 

```
* example-domain.com.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create <AaaaRecord AAAA 600, bbbb.example-domain.com., ['2601:644:500:e210:62f8:1dff:feb8:947a']> (config2)
*   Create <CnameRecord CNAME 1800, test3.example-domain.com., target.inikup.eu.> (config)
*   Create <TxtRecord TXT 600, txt3.example-domain.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create <AaaaRecord AAAA 600, bbbb.example-domain.com., ['2601:644:500:e210:62f8:1dff:feb8:947a']> (config2)
*   Create <CnameRecord CNAME 1800, test3.example-domain.com., target.inikup.eu.> (config)
*   Create <TxtRecord TXT 600, txt3.example-domain.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0
* example-domain.net.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create Zone<example-domain.net.>
*   Create <AaaaRecord AAAA 600, bbbb.example-domain.com., ['2601:644:500:e210:62f8:1dff:feb8:947a']> (config2)
*   Create <CnameRecord CNAME 1800, test3.example-domain.com., target.inikup.eu.> (config)
*   Create <TxtRecord TXT 600, txt3.example-domain.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0


2020-10-30T23:07:14  [4528168384] INFO  DnsMadeEasyProvider[dnsmadeeasy] apply: making changes
2020-10-30T23:07:17  [4528168384] INFO  DnsMadeEasyProvider[dnsmadeeasy] apply: making changes
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/bin/octodns-sync", line 33, in <module>
    sys.exit(load_entry_point('octodns==0.9.10', 'console_scripts', 'octodns-sync')())
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/cmds/sync.py", line 38, in main
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/manager.py", line 399, in sync
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/base.py", line 95, in apply
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 394, in _apply
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 365, in _apply_Create
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 138, in record_create
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 77, in _request
octodns.provider.dnsmadeeasy.DnsMadeEasyClientBadRequest:
  - Record with this type (AAAA), name (bbbb), and value (2601:644:500:e210:62f8:1dff:feb8:947a) already exists.
```

jleroy:
Nevermind, the duplicate records in plan was caused by a bug in my code.
But the duplicate record error is still here:

```
* example-002.com.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create <CnameRecord CNAME 3600, only-in-dot-com.example-002.com., target.tld.> (dnsimple)
*   Create <CnameRecord CNAME 1800, test3.example-002.com., target.inikup.eu.> (config)
*   Create <TxtRecord TXT 600, txt3.example-002.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0
* example-002.net.
* dnsmadeeasy (DnsMadeEasyProvider)
*   Create Zone<example-002.net.>
*   Create <CnameRecord CNAME 3600, only-in-dot-com.example-002.com., target.tld.> (dnsimple)
*   Create <CnameRecord CNAME 1800, test3.example-002.com., target.inikup.eu.> (config)
*   Create <TxtRecord TXT 600, txt3.example-002.com., ['Bah bah black sheep', 'have you any wool.', 'v=DKIM1\;k=rsa\;s=email\;h=sha256\;p=A/kinda+of/long/string+with+numb3rs']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0


2020-10-30T23:49:10  [4647837120] INFO  DnsMadeEasyProvider[dnsmadeeasy] apply: making changes
2020-10-30T23:49:14  [4647837120] INFO  DnsMadeEasyProvider[dnsmadeeasy] apply: making changes
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/bin/octodns-sync", line 33, in <module>
    sys.exit(load_entry_point('octodns==0.9.10', 'console_scripts', 'octodns-sync')())
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/cmds/sync.py", line 38, in main
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/manager.py", line 396, in sync
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/base.py", line 95, in apply
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 394, in _apply
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 365, in _apply_Create
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 138, in record_create
  File "/Users/jleroy/.pyvenv/OctoDNS_pip/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/dnsmadeeasy.py", line 77, in _request
octodns.provider.dnsmadeeasy.DnsMadeEasyClientBadRequest:
  - A CNAME record already exists with this name.
  - CNAME records are required to be unique values. There is a record in this domain that already exists with this name.
  - Record with this type (CNAME), name (only-in-dot-com), and value (target.tld.) already exists.
```

jleroy:
So I think everything works as expected now, except that we don't remove existing records on the aliased zone as we don't populate it... This is a tricky one, I don't really know how to handle it properly.

ross (commented):
> So I think everything works as expected now, except that we don't remove existing records on the aliased zone as we don't populate it... This is a tricky one, I don't really know how to handle it properly.

I'll have to pull this branch down to experiment to make sure, but I believe what's here should be functionally correct. 

The existing records are pulled down for each of the targets when `target.plan(desired)` is called. 

https://github.com/github/octodns/blob/3acea0d89d88a7f921429ab77e3efe9cd3fde392/octodns/provider/base.py#L47-L51

It should pull the `existing` state of that target with the populate call there. It will then compare that it to the desired state that's passed in to figure out what needs to change. That should include removing records.

Few bits minor bits inline until I get a chance to try this out.

octodns/manager.py
```diff
@@ -222,21 +222,32 @@ def configured_sub_zones(self, zone_name):
...
+                    self.log.warn(': provider %s does not accept lenient '
+                                  'param', source.__class__.__name__)
+                    source.populate(zone)
+
+        else:
+            for _, records in desired._records.items():
+                for record in records:
+                    d = record.data
+                    d['type'] = record._type
+                    r = Record.new(zone, record.name, d, source=record.source)
```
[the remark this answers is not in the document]
jleroy: Done.

octodns/manager.py
```diff
@@ -222,21 +222,32 @@ def configured_sub_zones(self, zone_name):
...
-            try:
-                source.populate(zone, lenient=lenient)
-            except TypeError as e:
-                if "keyword argument 'lenient'" not in text_type(e):
-                    raise
-                self.log.warn(': provider %s does not accept lenient param',
-                              source.__class__.__name__)
-                source.populate(zone)
+
+        if not desired:
```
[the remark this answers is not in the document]
jleroy: Done.

octodns/manager.py
```diff
@@ -331,6 +355,25 @@ def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
...
+        for zone_name, zone_source in aliased_zones.items():
+            plan = [p for t, p in plans if p.desired.name == zone_source]
+            if not plan:
+                continue
+
+            source_config = self.config['zones'][zone_source]
+            futures.append(self._executor.submit(
+                self._populate_and_plan,
+                zone_name,
+                [self.providers[s] for s in source_config['sources']],
```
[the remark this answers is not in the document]
jleroy: Done.

jleroy:
> It should pull the existing state of that target with the populate call there. It will then compare that it to the desired state that's passed in to figure out what needs to change. That should include removing records.

Nope, at least not im my tests. I've created a record called `only-in-dot-net-target` in my alias zone at DNS Made Easy and its not delete.

octodns/manager.py
```diff
@@ -331,6 +352,25 @@ def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
         # plan pairs.
         plans = [p for f in futures for p in f.result()]
 
+        # Populate aliases zones.
+        futures = []
+        for zone_name, zone_source in aliased_zones.items():
+            plan = [p for t, p in plans if p.desired.name == zone_source]
+            if not plan:
```
jleroy, line 359: I have not found how to generate an empty plan in order to test this condition.

tests/test_octodns_manager.py
```diff
@@ -309,6 +336,19 @@ def populate(self, lenient=False):
...
+    def test_alias_zones(self):
+        with TemporaryDirectory() as tmpdir:
+            environ['YAML_TMP_DIR'] = tmpdir.dirname
+
+            Manager(get_config_filename('simple-alias-zone.yaml')) \
+                .validate_configs()
+
+            with self.assertRaises(ManagerException) as ctx:
+                Manager(get_config_filename('unknown-source-zone.yaml')) \
+                    .validate_configs()
```
jleroy, line 348: I don't understand why this test does not cover this lines: https://github.com/github/octodns/pull/592/files#diff-1ed4adeb738f3f2202e89f6a229378fd0bcaf8467a61f694c34b73e24169280eR465

octodns/manager.py
ross: > Nope, at least not im my tests. I've created a record called `only-in-dot-net-target` in my alias zone at DNS Made Easy and it's not deleted.

This is actually the issue that's preventing there from being a run/plan on the aliased zone when there are no changes in the original. 

See https://gist.github.com/ross/a154334eb3a121a9240ba9c52f1c637c for the miniaml test case i did to set things up and test changes to the config. All that works as expected.

When I go into the zone in Route53 and manually add a record to aliased.net. Though it isn't detected as you saw. Reason for this is that orignal.net didn't have any changes so there's no plan for it. We come into this loop and get back no plans and come into this if and skip planning for aliased.net and don't see that it needs the extra record deleted (or any other changes that it might need.) 

So this if is why it's not even trying, but the problem is a little deeper as you point out since we won't have a copy of the desired state to pass in. Going to poke at this a bit.

ross: https://github.com/github/octodns/commit/661cbb2cd255fe4259e7b5130f3a0e5c8859c964 should have the fix for the `desired` issue. Basically it modifies `_populated_and_plan` to always returne the (desired) zone so that it can be stored and used for aliased zones if needed. 

You should be able to cherry pick that in with something like:

```console
$ git fetch origin
$ git cherry-pick 661cbb2cd255fe4259e7b5130f3a0e5c8859c964
```

I would have PR'd against yours so you could just merge it, but looks like that's not supported across forks.

octodns/manager.py
```diff
@@ -265,9 +273,22 @@ def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
...
 
+        aliased_zones  = {}
         futures = []
         for zone_name, config in zones:
             self.log.info('sync:   zone=%s', zone_name)
+            if 'alias' in config:
+                source_zone = config['alias']
+                # Check that the source zone is defined.
+                if source_zone not in self.config['zones']:
+                    self.log.exception('Invalid alias zone')
```
ross, line 284: There isn't an exception here top log, you're getting ready to create and throw one, so this probably needs to be

```suggestion
                    self.log.error('Invalid alias zone {}, target {} does '
                                   'not exist'.format(zone_name, source_zone))
```

or something similar.

tests/test_octodns_manager.py
ross: I believe it does run through the block as you expect. The lack of coverage is for the other case, where the zone does exist. What the coverage is saying is that things ALWAYS go into that block and never skip it and make it down to the `continue`.

octodns/manager.py
jleroy: I've applied you patch and everything woks perfectly now. Thank you!

octodns/manager.py
jleroy: Fixed.

tests/test_octodns_manager.py
```diff
@@ -286,6 +302,29 @@ def test_validate_configs(self):
...
 
+        # Alias zone using an invalid source zone.
+        with self.assertRaises(ManagerException) as ctx:
+            Manager(get_config_filename('unknown-source-zone.yaml')) \
+                .validate_configs()
+        self.assertTrue('Invalid alias zone' in
+                        text_type(ctx.exception))
+
+        # Valid config file using an alias zone.
+        Manager(get_config_filename('simple-alias-zone.yaml')) \
```
jleroy, line 313: I'm loosing my mind on this one. I simply not understand of this test using `simple-alias-zone.yaml` could not trigger `continue`. 
Or maybe I'm too tired. 🤷‍♂️

jleroy: Maybe this is caused by this coverage.py bug: https://github.com/nedbat/coveragepy/issues/198#issuecomment-399705984 ?

jleroy:
Documentation added, fell free to correct my frenglish.

octodns/manager.py
```diff
@@ -421,16 +466,25 @@ def validate_configs(self):
...
             zone = Zone(zone_name, self.configured_sub_zones(zone_name))
 
+            source_zone = config.get('alias')
+            if source_zone:
+                if source_zone not in self.config['zones']:
+                    self.log.exception('Invalid alias zone')
+                    raise ManagerException('Invalid alias zone {}: '
+                                           'source zone {} does not exist'
+                                           .format(zone_name, source_zone))
+                continue
```
ross, line 476: Yeah, this definitely appears to be https://github.com/nedbat/coveragepy/issues/198#issuecomment-399705984. Surprised I haven't run into this before. There's a pragma for coverage, but we've disabled the use of it b/c things slipped through in the past that were actually bugs/problems where it was used. 

The best short-term work-around I can think of is something like this:

```suggestion
                # this is just here to satisfy coverage, see
                # https://github.com/nedbat/coveragepy/issues/198
                source_zone = source_zone
```

I may look into a clean way to do the pragma and allow it "just this once" in the future, but for now that should do the trick without any real side effects :-/

tests/test_octodns_manager.py
jleroy: So yeah that's it: coverage is OK if I set the `continue` statement in a `try`/`except` block.
I've added a `pragma: no cover` comment in order to achieve 100% code coverage.

octodns/manager.py
jleroy: Woops sorry, I missed your comment. I will fix that.

jleroy: Hmmm... All checks have passed so I don't know if I should apply your change or not?

ross: Please apply it. It's a shortcoming of the check for the pragma. It looks for `pragma: nocover` and doesn't consider the case where it's `pragma: no cover` with a space. That's by accident and it'll get fixed the next time I'm in there.

jleroy: Done.

ross (commented):
I think this is there. Since it touches central/critical bits of things I'll :eye: it with a 🔍  Monday.

Thanks!

jleroy:
Found a little bug. Fixed.

octodns/record/__init__.py
```diff
@@ -219,6 +220,18 @@ def changes(self, other, target):
...
 
+    def copy(self, zone=None):
+        _type = getattr(self, '_type')
+        if not self._raw_data.get('type'):
+            self._raw_data['type'] = _type
+
+        return Record.new(
+            zone if zone else self.zone,
+            self.name,
+            self._raw_data,
```
ross, line 231: I believe it'll be safer to do `self.data` here rather than pipe the raw data through so that we get a copy of the record as it is now and not in the state it was when created.

```suggestion
            self.data,
```

octodns/manager.py
```diff
@@ -222,21 +222,29 @@ def configured_sub_zones(self, zone_name):
...
-            try:
-                source.populate(zone, lenient=lenient)
-            except TypeError as e:
-                if "keyword argument 'lenient'" not in text_type(e):
-                    raise
-                self.log.warn(': provider %s does not accept lenient param',
-                              source.__class__.__name__)
-                source.populate(zone)
+
+        if desired:
```
ross, line 233:
```suggestion
        if desired:
            # This is an alias zone, rather than populate it we'll copy the records 
            # over from `desired`
```

octodns/record/__init__.py
```diff
@@ -219,6 +220,18 @@ def changes(self, other, target):
...
+            self.source
```
ross, line 232: I believe we'll want to pass `lenient=True` here so that we copy records that aren't 100% compliant. We don't know how this record was created, it could have had `lenient=True`, so it may not pass validation. If we try to copy that it'd fail. It should be fine to always copy with lenient as whatever is in `self` should be duplicated as-is wether it complies or not.

```suggestion
            self.source,
            lenient=True
```

ross (commented):
Added a few comments as suggestions and there's a few bits that I believe need to be addressed in the Record.copy method and then we should be good to go.

octodns/record/__init__.py
jleroy: The issue is that `self.data` only contains the record TTL, not the record type. Sometimes the record object is not a specific record like `ARecord` or `TXTRecord`, but just a `Record` object which has no `_type` attribute.

ross: Would it work to do something like?

```python
data = self.data
data['type'] = self._type
# then use `data` as the param.
```

jleroy: All tests passed so everything seems to be OK.

octodns/manager.py
jleroy: Done.

octodns/record/__init__.py
jleroy: Done.

ross (approved):
Definitely a journey of a PR, but happy with where we got. Thanks again.

jleroy:
Thank you for your help and patience @ross.

PR #599, opened 2020-08-27
Implement octodns-sync --source

ramnes:
It can be useful to only synchronize zones that use a certain source. For example, in a situation where some zones use a dynamic source and others don't, you probably want to synchronize those with a dynamic source regularly, and only synchronize the others when a change is made.

Although we only synchronize the zones that use a given source, we still want to synchronize all sources to avoid deleting records that would live in another source of the zone.


ramnes:
Awesome. :)

Can we expect a release soon?

ross:
> Can we expect a release soon?

It's on my todo list, but no ETA at the moment as it's kind of overflowing :-/

PR #610, opened 2020-09-11
Allow saved plan

thattommyhall:
I am in an environment where I can't be certain people don't change things manually, this allows me to only apply the plan I've seen and approved, similar to terraform

### DIFF
--- a/octodns/cmds/sync.py
+++ b/octodns/cmds/sync.py
@@ -32,11 +32,16 @@ def main():
     parser.add_argument('--target', default=[], action='append',
                         help='Limit sync to the specified target(s)')
 
+    parser.add_argument('--save-plan', action='store_true', default=False, help="Save the plan")
+
+    parser.add_argument('--load-plan', action='store_true', default=False, help="Load the saved plan")
+
     args = parser.parse_args()
 
     manager = Manager(args.config_file)
     manager.sync(eligible_zones=args.zone, eligible_sources=args.source,
                  eligible_targets=args.target, dry_run=not args.doit,
+                 save_plan=args.save_plan, load_plan=args.load_plan,
                  force=args.force)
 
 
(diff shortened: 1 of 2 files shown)


thattommyhall:
Let me know if this is something you might want and I can sort out tests.

ross:
> Let me know if this is something you might want and I can sort out tests.

Overall 👍 on the idea and it was originally my intention to add something like this, things were built with it in mind. Just never personally ran into a situation that required it so hadn't gotten around to it. 

Main concern would be around decoupling the plan from the apply is that after some time the state of the world could have changed in ways that would cause the plan not to make sense any longer or to fail to apply at all. E.g. say you planned an 15m ago and have a change that will add a new `A` record `foo.bar.com`. In that hour someone came in and manually added an `A` record with that name. Now you go to apply things and you change will fail since that record already exists. At least in that case it'd fail and not break things, but I'm fairly sure I could come up with problematic scenarios if I thought about it for a while. 

My original thinking is that we'd need to do the normal populated & plan process and then compare the newly made plan to the old one. If they're the same they can be applied, if they differ then we'd need to balk and refuse to proceed. That'd roughly be the way that terraform works. 

The above leads to me lean towards a plan checksum rather than saving and loading of a plan. When you plan part of that plan prints out a checksum value for itself. Instead of passing in the plan output they'd instead pass in the checksum value `--doit --plan-checksum abcdef123456...`. The normal flow would would happen: octodns would pull down the current state and build a plan and only go forward with applying it if the checksums match. Mean the current state and the changes being made hadn't drifted since the checksum was generated.

My rough thinking on how to go about that would be a `def to_data` on the `Plan` object that called down to `def to_data` methods on the change objects to build up a data structure of the plan. Then running that through `json` with sorted keys etc to keep it consistent should reliable reproduce the same sha.

thattommyhall:
Cheers for the reply, I'll see about moving towards what you suggest. 

In terms of reputability of the hash, it's better to bencode or use cbor I think, I've some previous work doing it with bencode (https://github.com/binaryaffairs/a-la-mode)

PR #616, opened 2020-10-14
Add support for dynamic records to Constellix provider

mintopia:
This PR adds support for dynamic records to the Constellix provider. This maps to the Constellix Pool resource.

I would like to eventually add support for Geo Records, so this is a step towards that.


octodns/provider/constellix.py
```diff
@@ -161,9 +162,52 @@ def record_delete(self, zone_name, record_type, record_id):
             record_type = 'ANAME'
 
         zone_id = self.domains.get(zone_name, False)
-        path = '/{}/records/{}/{}'.format(zone_id, record_type, record_id)
+        path = '/domains/{}/records/{}/{}'.format(zone_id, record_type,
+                                                  record_id)
         self._request('DELETE', path)
 
+    def pools(self, pool_type):
+        if self._pools is None:
```
ross, line 170: It looks like the parameter `pool_type` will make a difference in the API call made, but this will cache things the first time through and then just short-circuit and return the cached value from then on. Seems like if this was called with an `A` and then later with `AAAA` it'd return the `A`s the second time.

octodns/provider/constellix.py
```diff
@@ -194,12 +238,41 @@ def __init__(self, id, api_key, secret_key, ratelimit_delay=0.0,
...
+            })
+            values.append(value['value'])
+        return {
+            'ttl': record['ttl'],
+            'type': _type,
+            'dynamic': {
+                'pools': pools,
+                'rules': [{
+                    'pool': pool_name
+                }]
```
ross, line 271: 👍 This is a good way to incrementally work towards dynamic records. It should plan & apply fine for anything with a single pool and only a default rule.

ross (commented):
:wave: @mintopia. This looks like a good start on dynamic support. I made a quick pass through it and left a comment/question inline.

octodns/provider/constellix.py
mintopia: I've updated this logic to now check the pool type as well as the name. I've also added unit tests to verify this behaviour.

mintopia:
Is there anything else you need from me before this can be accepted/merged?

ross:
> Is there anything else you need from me before this can be accepted/merged?

Hi @mintopia if we merge a partial implementation of `dynamic` support like this anyone using it for records beyond it's level of support would run into cases where the provider would constantly see differences and apply would continually update the record with no changes and they'd be stuck in a loop. This can probably be tested by adding a couple rules to a record and trying to apply it.

Unfortunately there isn't currently a mechanism for partial feature/dynamic support.

ross:
Sorry there's not a better story around partial `dynamic` support. This something you're still planning to pursue? I'm not familiar enough with Constellix to know what it's geo dns setup looks like so I don't really know which of the existing dynamic providers would serve as the best reference. It's dobutful that Route53 would as it's the most unique. Dyn's is the simplest as the octoDNS model is fairly close to what it does. NS1 didn't start out too complicated, but over time it got more so as we ran into hiccups details in getting it's behavior to match octoDNS's expectations. 

I'd be happy to help to the degree I can in terms of 👀 code and making suggestions. Thanks for what you've done thus far.

gartline:
Hi @ross 
Full Dynamic records support has been added by my last commit.
Dynamic Records now support geo rules.
Please review it.
Geo Records going to be added soon.

octodns/provider/constellix.py
```diff
@@ -99,14 +102,26 @@ def domain(self, name):
...
     def domain_create(self, name):
-        resp = self._request('POST', '/', data={'names': [name]})
+        resp = self._request('POST', '/domains', data={'names': [name]})
         # Add newly created zone to domain cache
         self._domains['{}.'.format(name)] = resp.json()[0]['id']
 
+    def domain_enable_geoip(self, domain_name):
+        domain = self.domain(domain_name)
+        log = logging.getLogger('ConstellixProvider[{}]'.format('test'))
+        log.debug('===== domain %s', domain)
```
ross, line 116: logging to debug is 👍, but please make this a more meaningful message.

octodns/provider/constellix.py
```diff
@@ -99,14 +102,26 @@ def domain(self, name):
         zone_id = self.domains.get(name, False)
         if not zone_id:
             raise ConstellixClientNotFound()
-        path = '/{}'.format(zone_id)
+        path = '/domains/{}'.format(zone_id)
```
ross, line 105: Heads up that we're moving towards f-strings now that we're >3.6 so feel free to make that change here if you wish.

/cc https://github.com/octodns/octodns/pull/768

ross (commented):
See https://github.com/octodns/octodns/pull/761, specifically the Route53 `_process_desired_zone` bit for how detailed "supports" stuff is going to work in the future. If Constellix has any mis-matches with octoDNS's `dynamic` semantics that's the best way to handle it going forward.

Overall seems reasonable. I don't have the ability to verify the semantics of `dynamic` work correctly w/Constellix, but code-wise this is looking good. Couple minor things inline.

octodns/provider/constellix.py
```diff
@@ -156,9 +169,108 @@ def record_delete(self, zone_name, record_type, record_id):
...
 
         zone_id = self.domains.get(zone_name, False)
-        path = '/{}/records/{}/{}'.format(zone_id, record_type, record_id)
+        path = f'/domains/{zone_id}/records/{record_type}/{record_id}'
         self._request('DELETE', path)
 
+    def pools(self, pool_type):
+        if self._pools is None:
+            self._pools = {}
+            path = f'/pools/{pool_type}'
```
ross, line 178: Are there multiple values for `pool_type`? If so this only seems to cache the first one that comes through and the others will just get back whatever `pool_type` was passed the first time. (noticed this when I looked back at the original discussion, https://github.com/octodns/octodns/pull/616/files#r505936387.

If there's only one type then it probably makes sense to remove the parameter and hard code, or if it stays around a comment about how there's really only one value would prevent future questions. 

Sorry I didn't catch this on the previous round.

ross (commented):
One last question inline, sorry I missed it in the previous round. Once that's answered/resolved we can :shipit:

octodns/provider/constellix.py
gartline: @ross , you are right, missed this issue. Thanks for pointing me into it.
Have updated pools cash by adding separate cash values for every supported pool_type: A, AAAA, CNAME

ross: Fwiw it's best not to force push once you've opened the PR (as ready for review.) It looses the context and makes it tough to see what's changed since the last round of discussion.

ross (approved):

octodns/provider/constellix.py
gartline: Got it, sorry.

gartline:
Ho @ross

ross:
> Could you please confirm that Geo Records going to be deprecated?

They are. No reason to add support for them.

PR #619, opened 2020-10-24
Add support for Gandi LiveDNS

jleroy:
This pull request implement a Gandi provider using Gandi LiveDNS API.
Comments welcome.

### DIFF
--- a/octodns/provider/gandi.py
+++ b/octodns/provider/gandi.py
@@ -41,6 +41,12 @@ class GandiClientNotFound(GandiClientException):
         super(GandiClientNotFound, self).__init__(r.text)
 
 
+class GandiClientUnknownDomainName(GandiClientException):
+
+    def __init__(self, msg):
+        super(GandiClientUnknownDomainName, self).__init__(msg)
+
+
 class GandiClient(object):
 
     def __init__(self, token):
@@ -62,6 +68,16 @@ class GandiClient(object):
             raise GandiClientNotFound(r)
         r.raise_for_status()
         return r
+
+    def zone(self, zone_name):
+        return self._request('GET', '/livedns/domains/{}'
+                                    .format(zone_name)).json()
+
+    def zone_create(self, zone_name):
+        return self._request('POST', '/livedns/domains', data={
+            'fqdn': zone_name,
+            'zone': {}
+        }).json()
 
     def zone_records(self, zone_name):
         records = self._request('GET', '/livedns/domains/{}/records'
@@ -318,8 +334,24 @@ class GandiProvider(BaseProvider):
     def _apply(self, plan):
         desired = plan.desired
         changes = plan.changes
+        zone = desired.name[:-1]
         self.log.debug('_apply: zone=%s, len(changes)=%d', desired.name,
                        len(changes))
+
+        try:
+            self._client.zone(zone)
+        except GandiClientNotFound:
+            self.log.info('_apply: no existing zone, trying to create it')
+            try:
+                self._client.zone_create(zone)
+                self.log.info('_apply: zone has been successfully created')
+            except GandiClientNotFound:
+                raise GandiClientUnknownDomainName('This domain is not '
+                                                   'registred at Gandi. '
+                                                   'Please register or '
+                                                   'transfer it here '
+                                                   'to be able to manage its '
+                                                   'DNS zone.')
 
         # Force records deletion to be done before creation in order to avoid
         # "CNAME record must be the only record" error when an existing CNAME
--- a/octodns/provider/gandi.py
+++ b/octodns/provider/gandi.py
@@ -354,12 +354,16 @@ class GandiProvider(BaseProvider):
                 self._client.zone_create(zone)
                 self.log.info('_apply: zone has been successfully created')
             except GandiClientNotFound:
-                raise GandiClientUnknownDomainName('This domain is not '
-                                                   'registred at Gandi. '
-                                                   'Please register or '
-                                                   'transfer it here '
-                                                   'to be able to manage its '
-                                                   'DNS zone.')
+                # We suppress existing exception before raising
+                # GandiClientUnknownDomainName.
+                e = GandiClientUnknownDomainName('This domain is not '
+                                                 'registred at Gandi. '
+                                                 'Please register or '
+                                                 'transfer it here '
+                                                 'to be able to manage its '
+                                                 'DNS zone.')
+                e.__cause__ = None
+                raise e
 
         # Force records deletion to be done before creation in order to avoid
         # "CNAME record must be the only record" error when an existing CNAME
--- a/octodns/provider/gandi.py
+++ b/octodns/provider/gandi.py
@@ -1,0 +1,334 @@
+#
+#
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from collections import defaultdict
+from requests import Session
+import logging
+
+from ..record import Record
+from .base import BaseProvider
+
+
+class GandiClientException(Exception):
+    pass
+
+
+class GandiClientBadRequest(GandiClientException):
+
+    def __init__(self, r):
+        super(GandiClientBadRequest, self).__init__(r.text)
+
+
+class GandiClientUnauthorized(GandiClientException):
+
+    def __init__(self, r):
+        super(GandiClientUnauthorized, self).__init__(r.text)
+
+
+class GandiClientForbidden(GandiClientException):
+
+    def __init__(self, r):
+        super(GandiClientForbidden, self).__init__(r.text)
+
+
+class GandiClientNotFound(GandiClientException):
+
+    def __init__(self, r):
+        super(GandiClientNotFound, self).__init__(r.text)
+
+
+class GandiClient(object):
+
+    def __init__(self, token):
+        session = Session()
+        session.headers.update({'Authorization': 'Apikey {}'.format(token)})
+        self._session = session
+        self.endpoint = 'https://api.gandi.net/v5'
+
+    def _request(self, method, path, params={}, data=None):
+        url = '{}{}'.format(self.endpoint, path)
+        r = self._session.request(method, url, params=params, json=data)
+        if r.status_code == 400:
+            raise GandiClientBadRequest(r)
+        if r.status_code == 401:
+            raise GandiClientUnauthorized(r)
+        elif r.status_code == 403:
+            raise GandiClientForbidden(r)
+        elif r.status_code == 404:
+            raise GandiClientNotFound(r)
+        r.raise_for_status()
+        return r
+
+    def zone_records(self, zone_name):
+        records = self._request('GET', '/livedns/domains/{}/records'
+                                .format(zone_name)).json()
+
+        for record in records:
+            if record['rrset_name'] == '@':
+                record['rrset_name'] = ''
+
+        return records
+
+    def record_create(self, zone_name, data):
+        self._request('POST', '/livedns/domains/{}/records'.format(zone_name),
+                      data=data)
+
+    def record_delete(self, zone_name, record_name, record_type):
+        self._request('DELETE', '/livedns/domains/{}/records/{}/{}'
+                      .format(zone_name, record_name, record_type))
+
+
+class GandiProvider(BaseProvider):
+    '''
+    Gandi provider using API v5.
+
+    gandi:
+        class: octodns.provider.gandi.GandiProvider
+        # Your API key (required)
+        token: XXXXXXXXXXXX
+    '''
+
+    SUPPORTS_GEO = False
+    SUPPORTS_DYNAMIC = False
+    SUPPORTS = set((['A', 'AAAA', 'ALIAS', 'CAA', 'CNAME', 'DNAME',
+                     'MX', 'NS', 'PTR', 'SPF', 'SRV', 'SSHFP', 'TXT']))
+
+    def __init__(self, id, token, *args, **kwargs):
+        self.log = logging.getLogger('GandiProvider[{}]'.format(id))
+        self.log.debug('__init__: id=%s, token=***', id)
+        super(GandiProvider, self).__init__(id, *args, **kwargs)
+        self._client = GandiClient(token)
+
+        self._zone_records = {}
+
+    def _data_for_multiple(self, _type, records):
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'values': [v.replace(';', '\\;') for v in
+                       records[0]['rrset_values']] if _type == 'TXT' else
+            records[0]['rrset_values']
+        }
+
+    _data_for_A = _data_for_multiple
+    _data_for_AAAA = _data_for_multiple
+    _data_for_TXT = _data_for_multiple
+    _data_for_SPF = _data_for_multiple
+    _data_for_NS = _data_for_multiple
+
+    def _data_for_CAA(self, _type, records):
+        values = []
+        for record in records[0]['rrset_values']:
+            flags, tag, value = record.split(' ')
+            values.append({
+                'flags': flags,
+                'tag': tag,
+                # Remove quotes around value.
+                'value': value[1:-1],
+            })
+
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'values': values
+        }
+
+    def _data_for_single(self, _type, records):
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'value': records[0]['rrset_values'][0]
+        }
+
+    _data_for_ALIAS = _data_for_single
+    _data_for_CNAME = _data_for_single
+    _data_for_DNAME = _data_for_single
+    _data_for_PTR = _data_for_single
+
+    def _data_for_MX(self, _type, records):
+        values = []
+        for record in records[0]['rrset_values']:
+            priority, server = record.split(' ')
+            values.append({
+                'preference': priority,
+                'exchange': server
+            })
+
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'values': values
+        }
+
+    def _data_for_SRV(self, _type, records):
+        values = []
+        for record in records[0]['rrset_values']:
+            priority, weight, port, target = record.split(' ', 3)
+            values.append({
+                'priority': priority,
+                'weight': weight,
+                'port': port,
+                'target': target
+            })
+
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'values': values
+        }
+
+    def _data_for_SSHFP(self, _type, records):
+        values = []
+        for record in records[0]['rrset_values']:
+            algorithm, fingerprint_type, fingerprint = record.split(' ', 2)
+            values.append({
+                'algorithm': algorithm,
+                'fingerprint': fingerprint,
+                'fingerprint_type': fingerprint_type
+            })
+
+        return {
+            'ttl': records[0]['rrset_ttl'],
+            'type': _type,
+            'values': values
+        }
+
+    def zone_records(self, zone):
+        if zone.name not in self._zone_records:
+            try:
+                self._zone_records[zone.name] = \
+                    self._client.zone_records(zone.name[:-1])
+            except GandiClientNotFound:
+                return []
+
+        return self._zone_records[zone.name]
+
+    def populate(self, zone, target=False, lenient=False):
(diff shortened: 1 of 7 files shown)


octodns/provider/gandi.py
```diff
@@ -0,0 +1,334 @@
...
+            raise GandiClientNotFound(r)
+        r.raise_for_status()
+        return r
+
+    def zone_records(self, zone_name):
+        records = self._request('GET', '/livedns/domains/{}/records'
+                                .format(zone_name)).json()
+
+        for record in records:
+            if record['rrset_name'] == '@':
```
jleroy, line 71: I don't know how to test this line to achieve 100% coverage.

octodns/provider/gandi.py
```diff
@@ -0,0 +1,334 @@
...
+        return self._zone_records[zone.name]
+
+    def populate(self, zone, target=False, lenient=False):
+        self.log.debug('populate: name=%s, target=%s, lenient=%s', zone.name,
+                       target, lenient)
+
+        values = defaultdict(lambda: defaultdict(list))
+        for record in self.zone_records(zone):
+            _type = record['rrset_type']
+            if _type not in self.SUPPORTS:
```
jleroy, line 217: The same.

tests/fixtures/gandi-no-changes.json
```diff
@@ -0,0 +1,127 @@
+[
+      {
+        "rrset_type": "A",
+        "rrset_ttl": 300,
+        "rrset_name": "",
+        "rrset_href": "https://api.gandi.net/v5/livedns/domains/reductioncarbone.fr/records/%40/A",
```
yzguy, line 6: You'll want to swap this to unit.tests instead of reductioncarbone.fr

yzguy:
Hello,

Awesome for the provider contribution! Always cool to see more supported providers.

For both your comments for testing, you can reference a provider like [DigitalOcean](https://github.com/github/octodns/blob/master/tests/test_octodns_provider_digitalocean.py) and it's tests/fixtures

For the `@` one, you would want to add/modify a record to have the `rrset_name` of `@`. I see in yours you have just `""`, is this expected back from the Gandi API? If so, then you don't have to swap between `@` and `""`. This was done in other providers because they use an `@` to denote a record at the apex of the zone, and that will be the value in the API, and is the value expected by the API when creating a record at the apex of the zone.

Your second comment with the `self.SUPPORTS`, you would just have a record that is of a type that is not supported by the provider, but supported by Gandi, in this case something like a `CDS` record, or a `DS` record. This would show up from an API from Gandi, but because it's not in the supported record types of the provider it would not be considered.

octodns/provider/gandi.py
```diff
@@ -0,0 +1,334 @@
...
+
+    gandi:
+        class: octodns.provider.gandi.GandiProvider
+        # Your API key (required)
+        token: XXXXXXXXXXXX
+    '''
+
+    SUPPORTS_GEO = False
+    SUPPORTS_DYNAMIC = False
+    SUPPORTS = set((['A', 'AAAA', 'ALIAS', 'CAA', 'CNAME', 'DNAME',
```
yzguy, line 97: I don't believe DNAME is used by any other provider, nor does it have validation code to be handled. I believe it can be removed from this supported list to avoid issues.

tests/fixtures/gandi-no-changes.json
jleroy: Fixed.

octodns/provider/gandi.py
jleroy: I just submitted a PR to add support for DNAME records (#620).

jleroy:
> For the `@` one, you would want to add/modify a record to have the `rrset_name` of `@` in the fixtures. I see in yours you have just `""`, is this expected back from the Gandi API? 

You're right, Gandi's API return "@" for APEX records. I've updated the fixtures.


> Your second comment with the `self.SUPPORTS`, you would just have a record that is of a type that is not supported by the provider, but supported by Gandi, in this case something like a `CDS` record, or a `DS` record. This would show up from an API from Gandi, but because it's not in the supported record types of the provider it would not be considered. 

Thank you for your help. We're now at 100% code coverage. :)

yzguy:
Giving this provider a try, couple things I noticed

When trying to create an ALIAS record for `blah.yzguy.xyz`, I get this error
```
octodns.provider.gandi.GandiClientBadRequest: {"status": "error", "errors": [{"location": "body", "name": "items", "description": "only '@' is allowed for ALIAS records (param: {'rrset_type': u'ALIAS', 'rrset_ttl': 3600, 'rrset_name': u'blah', 'rrset_values': [u'google.com.']})"}]}
```

Another thing I'm not sure how it'll work is the domain creation if it doesn't exist. I am not super familiar with Gandi (created zone to test this provider), so I'm not sure if you can just create a domain in their DNS service without purchasing it from them or transferring it to them first.

You probably need to go add in some code to always take any relative values from Gandi's API and make them absolute, so when they are compared to the absolute values of other providers/config they are seen as equal. Otherwise, OctoDNS will delete the relative values and replace them with absolute. This is done in a couple others providers like [DNSMadeEasy](https://github.com/github/octodns/blob/master/octodns/provider/dnsmadeeasy.py#L120-L126), [Constellix](https://github.com/github/octodns/blob/master/octodns/provider/constellix.py#L130-L139)
```
* yzguy.xyz.
* gandi (GandiProvider)
*   Update
*     <NsRecord NS 1800, blah.yzguy.xyz., ['ns1', 'ns2']> ->
*     <NsRecord NS 1800, blah.yzguy.xyz., ['ns1.yzguy.xyz.', 'ns2.yzguy.xyz.']> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=5
```

jleroy:
> Seems an ALIAS record is only allowed on the Apex, which is a less than ideal limitation of Gandi's LiveDNS server it seems. Not sure how to handle that. Other providers allow ALIAS records anywhere in the zone.

An ALIAS record is basically a CNAME for APEX, so it makes sense to me (?).
Maybe I will try to reach the Gandi team to see if there's a technical reason for this limitation.


> Potentially if you have to have the zone in Gandi (buy or transfer) then you can catch the NotFound with an exception in `record_create` and just raise something like a `GandiClientZoneNotFound` exception, and a 'Zone Not Found, purchase or transfer zone to Gandi' type message

Done. 
However, I have absolutely no idea how to test that the `GandiClientUnknownDomainName` exception is raised without breaking all others tests. If you have an idea...


> 1. You probably need to go add in some code to always take any relative values from Gandi's API and make them absolute, so when they are compared to the absolute values of other providers/config they are seen as equal. Otherwise, OctoDNS will delete the relative values and replace them with absolute. 

Done.

jleroy:
So Gandi LiveDNS is based on PowerDNS, which only allow ALIAS records at zone apex.

I'm not sure about the best way to handle this limitation: raise an error or silently convert the record to a CNAME (on the fly)?
The second solution seems to be the most convenient from the users point of view.

ross:
> So Gandi LiveDNS is based on PowerDNS, which only allow ALIAS records at zone apex.

Interesting. I believe (trying to remember back several years now) that my intention for octoDNS's ALIAS record type was to only allow it at the apex, but it doesn't look like that's backed up with validations if that was the case though. I guess some providers may allow it to go anywhere... the PowerDNS provider does support ALIAS and thus would break if it was used elsewhere. 

Are there other providers that support them anywhere? If not we should go ahead and add that validation. If some do allow them anywhere I'm not really sure what the best thing to do is.

jleroy:
At least DNSimple and NS1 allows ALIAS records below the zone apex, although NS1 strongly [discourage](https://help.ns1.com/hc/en-us/articles/360020248973-About-CNAME-ALIAS-and-linked-records) its usage.

If we want to handle on the fly record conversion, we probably need to add an `ALIASES_LIMITED_TO_APEX` property to the providers and handle the conversion in the manager, which is a bit dirty. 
Honestly the more I dive into it, the more I think we should just raise an error as there is absolutely no advantage to uses an `ALIAS` record instead of a `CNAME` below the zone apex.

octodns/provider/gandi.py
jleroy: Merged.

Toronto00 (approved):

ross:
> Honestly the more I dive into it, the more I think we should just raise an error as there is absolutely no advantage to uses an `ALIAS` record instead of a `CNAME` below the zone apex.

I'd be :+1: on that. Should be fairly easy to add to the validations. I'll look at getting that going in a PR shortly.

ross:
> I'd be 👍 on that. Should be fairly easy to add to the validations. I'll look at getting that going in a PR shortly.

/cc https://github.com/github/octodns/pull/621

jleroy:
> #621

Thanks!

Toronto00 (approved):

Toronto00:
> Giving this provider a try, couple things I noticed below. Otherwise, everything worked as expected, I did a dump from Gandi into another provider, did it to a YAML file, took that YAML file and pushed it to a provider, synced a provider directly to Gandi with no issues
> 
> 
> 
> 1. When trying to create an ALIAS record for `blah.yzguy.xyz`, I get this error
> 
> 
> 
> ```
> 
> octodns.provider.gandi.GandiClientBadRequest: {"status": "error", "errors": [{"location": "body", "name": "items", "description": "only '@' is allowed for ALIAS records (param: {'rrset_type': u'ALIAS', 'rrset_ttl': 3600, 'rrset_name': u'blah', 'rrset_values': [u'google.com.']})"}]}
> 
> ```
> 
> 
> 
> Seems an ALIAS record is only allowed on the Apex, which is a less than ideal limitation of Gandi's LiveDNS server it seems. Not sure how to handle that. Other providers allow ALIAS records anywhere in the zone.
> 
> 
> 
> 2. OctoDNS will detect a zone does not exist, then create it before adding records to it. This was my experience trying to do so
> 
> 
> 
> ```
> 
> ********************************************************************************
> 
> * yzguy.io.
> 
> ********************************************************************************
> 
> * gandi (GandiProvider)
> 
> *   Create Zone<yzguy.io.>
> 
> ...
> 
> *   Summary: Creates=7, Updates=0, Deletes=0, Existing Records=0
> 
> ********************************************************************************
> 
> 
> 
> 2020-10-27T01:28:23  [4384497088] INFO  GandiProvider[gandi] apply: making changes
> 
> Traceback (most recent call last):
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/bin/octodns-sync", line 11, in <module>
> 
>     load_entry_point('octodns==0.9.10', 'console_scripts', 'octodns-sync')()
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/cmds/sync.py", line 38, in main
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/manager.py", line 361, in sync
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/base.py", line 95, in apply
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 331, in _apply
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 305, in _apply_create
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 77, in record_create
> 
>   File "/Users/yzguy/.pyenv/versions/3.8.5/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 62, in _request
> 
> octodns.provider.gandi.GandiClientNotFound: {"code": 404, "message": "The resource could not be found.", "object": "HTTPNotFound", "cause": "Not Found"}
> 
> ```
> 
> 
> 
> Potentially if you have to have the zone in Gandi (buy or transfer) then you can catch the NotFound with an exception in `record_create` and just raise something like a `GandiClientZoneNotFound` exception, and a 'Zone Not Found, purchase or transfer zone to Gandi' type message
> 
> 
> 
> 3. You probably need to go add in some code to always take any relative values from Gandi's API and make them absolute, so when they are compared to the absolute values of other providers/config they are seen as equal. Otherwise, OctoDNS will delete the relative values and replace them with absolute. This is done in a couple others providers like [DNSMadeEasy](https://github.com/github/octodns/blob/master/octodns/provider/dnsmadeeasy.py#L120-L126), [Constellix](https://github.com/github/octodns/blob/master/octodns/provider/constellix.py#L130-L139)
> 
> ```
> 
> ********************************************************************************
> 
> * yzguy.xyz.
> 
> ********************************************************************************
> 
> * gandi (GandiProvider)
> 
> *   Update
> 
> *     <NsRecord NS 1800, blah.yzguy.xyz., ['ns1', 'ns2']> ->
> 
> *     <NsRecord NS 1800, blah.yzguy.xyz., ['ns1.yzguy.xyz.', 'ns2.yzguy.xyz.']> (config)
> 
> *   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=5
> 
> ********************************************************************************
> 
> ```
> 
> 
> 
> I need to look into the DNAME stuff, I have no experience with those records.

ross:
Heads up that I'll let @yzguy continue the discussion review on this one since he's already involved. Once it's ✅ it I can merge.

yzguy:
Looking at the zone creation thing yields

```
* yzguy.io.
* gandi (GandiProvider)
*   Create Zone<yzguy.io.>
..
************************************************************************

2020-11-03T08:48:02  [139844368529216] INFO  GandiProvider[gandi] apply: making changes
2020-11-03T08:48:02  [139844368529216] INFO  GandiProvider[gandi] _apply: no existing zone, trying to create it
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 350, in _apply
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 73, in zone
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 68, in _request
octodns.provider.gandi.GandiClientNotFound: {"code": 404, "message": "The resource could not be found.", "object": "HTTPNotFound", "cause": "Not Found"}

During handling of the above exception, another exception occurred:

  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 354, in _apply
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 77, in zone_create
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 68, in _request
octodns.provider.gandi.GandiClientNotFound: {"code": 404, "message": "The resource could not be found.", "object": "HTTPNotFound", "cause": "Not Found"}

During handling of the above exception, another exception occurred:

  File "/home/yzguy/Projects/octodns/env/bin/octodns-sync", line 11, in <module>
    load_entry_point('octodns==0.9.10', 'console_scripts', 'octodns-sync')()
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/cmds/sync.py", line 38, in main
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/manager.py", line 418, in sync
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/base.py", line 95, in apply
  File "/home/yzguy/Projects/octodns/env/lib/python3.8/site-packages/octodns-0.9.10-py3.8.egg/octodns/provider/gandi.py", line 357, in _apply
octodns.provider.gandi.GandiClientUnknownDomainName: This domain is not registred at Gandi. Please register or transfer it here to be able to manage its DNS zone.
```
I'm thinking at these lines

```python
        try:
            self._client.zone(zone)
        except GandiClientNotFound:
            raise GandiClientUnknownDomainName('Domain was not found. Please register or transfer it to Gandi to manage it.')
```

If the zone is not found, then immediately raise the `GandiClientUnknownDomainName` exception, and we can tweak the message to something like "Domain was not found. Please register or transfer it to Gandi to manage it.' 

This would fail quickly, and get rid of 1 of the `GandiClientNotFound` exceptions, leaving just the one for when it tries to find the zone, and

yzguy:
The relative values equal to absolute normalization works correctly.

jleroy:
> If the zone is not found, then immediately raise the `GandiClientUnknownDomainName` exception, and we can tweak the message to something like "Domain was not found. Please register or transfer it to Gandi to manage it.'
> This would fail quickly, and get rid of 1 of the `GandiClientNotFound` exceptions, leaving just the one for when it tries to find the zone, and

We can't do that because a domain can be registered / transferred to Gandi with externals DNS servers. In that case no DNS zone is created on Gandi DNS servers. So we need to try to create the zone to known if the domain belongs to Gandi.

Another solution could be to call the domain API (`https://api.gandi.net/v5/domain/domains/{domain}`) to check that the domain belongs to the current account. 
Or we could just suppress the first exception by resetting `e.__context__` in order to avoid to add an API call: 

```python
try:
    self._client.zone(zone)
except GandiClientNotFound as e:
    e.__context__ = None
    self.log.info('_apply: no existing zone, trying to create it')
    try:
        self._client.zone_create(zone)
        self.log.info('_apply: zone has been successfully created')
    except GandiClientNotFound:
        raise GandiClientUnknownDomainName('This domain is not registred at Gandi. Please register or transfer it here to be able to manage its DNS zone.')
```

Unfortunately we can't use `raise ... from None` as it require Python 3.3+.

yzguy:
It should be fine, I just checked by swapping my test zone `yzguy.xyz` that is registered in Gandi to use external nameservers, put in NS1s just for good measure.

Before doing so, a call to `/livedns/domains/yzguy.xyz` gave me a valid response, after making the change to external nameservers I get a 404 response.

The LiveDNS API seems to be completely for DNS record management and nothing to do with domain registration. In that case, if OctoDNS can't find the domain in LiveDNS then it should error out and say it can't be managed.

Perhaps changing the message to something akin to "Cannot manage zone, zone not under Gandi DNS management". This would be a better reflection of saying DNS specifically is not managed by Gandi, therefore OctoDNS has no business trying to manage it through Gandi.

yzguy:
Ah I see, so even if you have the domain in Gandi, but aren't using LiveDNS it will show up in `/livedns/domains/{domain}`.
Just confirmed this on my zone.

There is a `/livedns/domains/{domain}/nameservers` path that could be called, then just look if `gandi.net` or even `gandi` is in any of them?

jleroy:
There's no need to look at the nameservers: if `/domain/domains/{domain}` return an HTTP status code `200`, then the zone can be created. Checking name servers is complicated as there is a lot possibles values (`a/b/c.dns.gandi.net`, `ns-([0-9]+)-(a|b|c).gandi.net`...) and in some cases a zone can exist even if the domain use external nameservers.

But IMHO there's very little benefit in doing that as we just replace the call to `/livedns/domains` (which is called by `zone_create()`) by another API call, which will be called everytime we run `_apply()`.

If the issue is the multiples exceptions, I still think that the best solution is to suppress previous exceptions before raising the `GandiClientUnknownDomainName` exception.

yzguy:
After messing around and basically coming to the same conclusions as you, I think you're right. Sorry for the round about.

> If the issue is the multiples exceptions, I still think that the best solution is to suppress previous exceptions before raising GandiClientUnknownDomainName.

This seems like the way to go.

jleroy:
No problem @yzguy. Done.

yzguy:
Looks good to me! Great job! @jleroy

yzguy (approved):

jleroy:
Thank you @yzguy and @ross. :)
Is there any plan to publish a new release on PyPI?

ross:
I plan to cut a release soon-ish, but I have some fixes and improvements to setup.py (few PRs and issues from a couple months back) that I want to get to first. I'm hoping to get those knocked out this week.

PR #620, opened 2020-10-26
Add support for DNAME records

jleroy:
This PR adds support for DNAME records, which is needed for #619.
I did not find a previous contribution to add a record type, so it is possible that I forgot some things.


tests/config/dynamic.tests.yaml
```diff
@@ -109,6 +109,40 @@ cname:
       - pool: iad
   type: CNAME
   value: target.unit.tests.
+dname:
+  dynamic:
```
ross, line 113: Any idea what provider(s) support complex dynamic DNAME records? It may be safer/easier to only support basic records to stat.

ross (commented):
Overall looks good/like what I'd expect. I somehow hadn't really run across DNAME records before so had to dig into what they do/are. One comment/question inline around `dynamic` support that could use discussion before this moves forward.

tests/config/dynamic.tests.yaml
jleroy: You're right, no provider support `DNAME` dynamic records right now.
I've removed this part.

BTW NS1 support for dynamic records seems to be broken when the record type is anything other than `A` / `AAAA`.

ross: Yeah. `dynamic` CNAME is pretty much alpha atm. It was tested some back when it was written, but very lightly and it needs some more attention/time 😦

ross (approved):
This looks like it probably covers everything. If not we'll find out soon enough when you try and use it on a provider :grin:

jleroy:
Thank you!

PR #621, opened 2020-11-02
ALIAS records only allowed at the root of zones

ross:
Support for this varies, and in general CNAMEs are a better tool to use for non-root cases. 

/cc https://github.com/github/octodns/pull/619#issuecomment-720532654 for discussion

### DIFF
--- a/script/coverage
+++ b/script/coverage
@@ -27,7 +27,7 @@ export DYN_USERNAME=
 export GOOGLE_APPLICATION_CREDENTIALS=
 
 # Don't allow disabling coverage
-grep -r -I --line-number "# pragma: nocover" octodns && {
+grep -r -I --line-number "# pragma: +no.*cover" octodns && {
   echo "Code coverage should not be disabled"
   exit 1
 }
--- a/docs/records.md
+++ b/docs/records.md
@@ -89,6 +89,8 @@ Each record type has a corresponding set of required data. The easiest way to de
 
 octoDNS is fairly strict in terms of standards compliance and is opinionated in terms of best practices. Examples of former include SRV record naming requirements and the latter that ALIAS records are constrained to the root of zones. The strictness and support of providers varies so you may encounter existing records that fail validation when you try to dump them or you may even have use cases for which you need to create or preserve records that don't validate. octoDNS's solution to this is the `lenient` flag.
 
+It's best to think of the `lenient` flag as "I know what I'm doing and accept any problems I run across." The main reason being is that some providers may allow the non-compliant setup and others may not. The behavior of the non-compliant records may even vary from one provider to another. Caveat emptor.
+
 #### octodns-dump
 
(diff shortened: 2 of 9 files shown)


script/coverage
```diff
@@ -27,7 +27,7 @@ export DYN_USERNAME=
 export GOOGLE_APPLICATION_CREDENTIALS=
 
 # Don't allow disabling coverage
-grep -r -I --line-number "# pragma: nocover" octodns && {
+grep -r -I --line-number "# pragma: +no.*cover" octodns && {
```
ross, line 30: Just piggybacking this one on this PR.

omgjlk (approved):

brianeclow:
There is a case in which using an ALIAS for a non-AWS provider is preferred over a CNAME. When you need to have any additional record type co-exist on a sub-domain, as in the case of a sub-domain being used for mail from a third party tool or a remailer provider.

example:
```
email:
- type: ALIAS
  value: jigglypuff.provider.net.
- type: TXT
  value: v=spf1 include:remailer.com include:_spf.bigmail.com -all
```

ross:
😞 I was worried there were valid case that I hadn't run across since we don't have (universal) support for them across our providers we haven't gone this route in such situations, instead just going with `A` records. 

Still a little torn here. Since provider support isn't consistent it'd be fairly surprising if you tried to do non-root `ALIAS` and then switched from a provider that does support it to one that doesn't. 

We can potentially go down this route and reject non-root `ALIAS`s as something that's "against" octoDNS's spect. They'd still be doable if you wanted to "try harder" and put an 

```yaml
octodns
  lenient: true
```

On the records. Going to test out and make sure that works now. If so how do you feel about that @brianeclow?

ross:
> On the records. Going to test out and make sure that works now. If so how do you feel about that @brianeclow?

Verified that lenient on the `ALIAS` will skip validation and allow the record to be pushed as expected/is the case for other slightly non-compliant records.

brianeclow:
We currently have the sighted use case in one place due to the constraints of the intersection between a particular vendor and a hard security requirement. However, due to the singular nature of CNAMEs and the proliferation of remailers, I can see this spreading to other locations in our DNS deployments.

To your waxing poetic about moving from one provider to another, I have been keeping my ALIAS records contained to one provider and known behaviors. And documenting the lack of parity ALIAS records across providers.

To the question, yes, that sounds like a viable solution. I can deploy and document the use case.

ross:
> To the question, yes, that sounds like a viable solution. I can deploy and document the use case.

:cool: Will look at trying to find a good place to document this before merging the PR. Thanks!

ross:
Bunch of docs around `lenient` added in https://github.com/github/octodns/pull/621/commits/b017f90c669d17113ec53db055fbc3ad2ba02aee. Going to give them another proofread and :ship: if that doesn't turn anything up.

Issue #624, opened 2020-11-12
axfr source does not support CAA records.


ConscriptMR:
we use Octodns in our environment and we use the axfr source. we need to add support to CAA records but the current source doesn't support it.

yzguy:
Can do, easy peasy. Not sure why I didn't do it originally. Keep an eye out, working on it now.

PR #625, opened 2020-11-12
Add CAA record support to AxfrSource/ZoneFileSource

yzguy:
Addresses #624

When originally created the AxfrSource/ZoneFileSource did not support CAA records. This change adds support for this record type

/cc @ConscriptMR


ross (approved):

ConscriptMR:
Thank you very much!

ConscriptMR:
@yzguy Hey Adam, should this be available using `pip install octodns==0.9.11` ? because it doesn't work for me for some reason

ross:
Looks like this was merged after that release was cut. There won't be another release for a bit still as not a whole lot has changed.

Best option is to install with a sha, e.g. a `requirements.txt` with the following in it will pull the sha that's currently master. 

```
-e git+https://git@github.com/github/octodns.git@07b8c19acc337efc90391f5024a346ce0956a5be#egg=octodns
```

That won't change on you and will always pick that specific point in time.

PR #627, opened 2020-11-16
Fix name length validation

### LINKED ISSUE #626: Invalid record name length validation
Hey there,

I think the implementation of #255 in #415 is wrong: it's not the whole record name that should be shorter than 63 character, but each label.

From [Route 53 documentation](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html):

> Domain names (including the names of domains, hosted zones, and records) consist of a series of labels separated by dots. Each label can be up to 63 bytes long. The total length of a domain name cannot exceed 255 bytes, including the dots.

From [Wikipedia DNS article](https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_space):

> The domain name space consists of a tree data structure. Each node or leaf in the tree has a label [...] A label may contain zero to 63 characters.

This is resulting in octoDNS blocking valid records:

```
octodns.record.ValidationError: Invalid record _acme-challenge.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxx.domain.com.
  - invalid name, "_acme-challenge.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.xxxxxxxx" is too long at 65 chars, max is 63
```

A PR will follow in a few minutes.

cc @ross

ramnes:
Closes #626

### DIFF
--- a/tests/test_octodns_record.py
+++ b/tests/test_octodns_record.py
@@ -1315,7 +1315,7 @@ class TestRecordValidation(TestCase):
         self.assertTrue(reason.endswith('.unit.tests." is too long at 254'
                                         ' chars, max is 253'))
 
-        # label length, DNS defins max as 63
+        # label length, DNS defines max as 63
         with self.assertRaises(ValidationError) as ctx:
             # The . will put this over the edge
             name = 'x' * 64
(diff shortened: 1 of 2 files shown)


tests/test_octodns_record.py
```diff
@@ -1275,10 +1275,18 @@ def test_base(self):
...
                                         ' chars, max is 63'))
 
+        # should not raise with dots
+        name = 'xxxxxxxx.' * 10
+        Record.new(self.zone, name, {
+            'ttl': 300,
+            'type': 'A',
+            'value': '1.2.3.4',
+        })
+
```
ross, line 1289: Please add in a case where the there's some valid pieces and some that are too long, e.g.

```
`foo.` + ('x' * 65) + .bar`
```

so that we're explicitly testing for a "bad" case with dots and not just relying on the non-dot too long case up above to trigger the validation, in case the implementation changes someday.

ramnes: Good idea, done.

ross (approved):
Thanks!

PR #631, opened 2020-11-22
Improve ALIAS, CNAME, DNAME & PTR record FQDN validation

### LINKED ISSUE #628: PTR, Alias, CNAME, and DNAME target value validation
Hello,

I'm preparing a pull request to improve the validation (cc #3) of CNAME, DNAME, ALIAS target record, make sure that they should be a valid FQDN in the record, the PR is almost ready, just need to pass the tests right now.

During the test process, I found that there are two test cases of Azure DNS in [`test_octodns_provider_azuredns.py`](https://github.com/github/octodns/blob/master/tests/test_octodns_provider_azuredns.py) seem to have empty record, and the target value of these two tests seem to be `.` if you print it out:

1. [`cname2`](https://github.com/github/octodns/blob/79cb88ef2566fc7863d3f71ad078bea081c1c04e/tests/test_octodns_provider_azuredns.py#L392-L394)
1. [`ptr2`](https://github.com/github/octodns/blob/79cb88ef2566fc7863d3f71ad078bea081c1c04e/tests/test_octodns_provider_azuredns.py#L416-L417)

So, before this _not sent yet_ pull request, OctoDNS seems to treat those records with value `.` valid, and the above tests should pass, looks a little bit weird.

Here are some error message from the cloud DNS services that I tested and doesn't support `.` as the record of the types mention above:

- Cloudflare: `DNS Validation Error (Code: 1004) Content for CNAME/PTR record is invalid`
- DigitalOcean: `Invalid hostname`
- Gandi: `An error has occurred. Please try again.` / `The server has either erred or is incapable of performing the requested operation.`
- Azure: `Each label must contain at least one character. You may not input consecutive period '.' characters.`

Even though that AWS Route53 and Google Cloud DNS seems to accept `.`, the nslookup or dig process won't return something really useful or meaningful(I replaced my own domain record with `root.example.com` as masked result):

```
$ nslookup root.example.com 8.8.8.8
Server:		8.8.8.8
Address:	8.8.8.8#53

Non-authoritative answer:
root.example.com	canonical name = .
```

The content of RFC is a lot, and I haven't read it all or enough to tell should it be valid or not, just guess this behavior may need to be changed, what do you think? Thanks!

PeterDaveHello:
Use fqdn package to help verify if the record value is really valid.

The original behavior will treat value like `_.` or `.` be a valid
record, which is strange, and the real world may not have those use
cases at all.

The RFC documents are pretty long, as I didn't read them all or enough
to tell should it be valid or not by the spec, so I opened issue #612 to
discuss this case and got a positive response from the main maintainer
to have the change.

Close #628

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -10,6 +10,7 @@ from logging import getLogger
 import re
 
 from six import string_types, text_type
+from fqdn import FQDN
 
 from ..equality import EqualityTupleMixin
 from .geo import GeoCodes
@@ -757,6 +758,9 @@ class _TargetValue(object):
             reasons.append('empty value')
         elif not data:
             reasons.append('missing value')
+        elif not FQDN(data, allow_underscores=True).is_valid:
+            reasons.append('{} value "{}" is not a valid FQDN'
+                           .format(_type, data))
         elif not data.endswith('.'):
             reasons.append('{} value "{}" missing trailing .'
                            .format(_type, data))
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -416,14 +416,8 @@ class AzureProvider(BaseProvider):
             :type  azrecord: azure.mgmt.dns.models.RecordSet
 
             :type  return: dict
-
-            CNAME and PTR both use the catch block to catch possible empty
-            records. Refer to population comment.
         '''
-        try:
-            return {'value': _check_endswith_dot(azrecord.cname_record.cname)}
-        except:
-            return {'value': '.'}
+        return {'value': _check_endswith_dot(azrecord.cname_record.cname)}
 
     def _data_for_MX(self, azrecord):
         return {'values': [{'preference': ar.preference,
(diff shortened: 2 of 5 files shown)


octodns/record/__init__.py
```diff
@@ -757,6 +758,9 @@ def validate(cls, data, _type):
             reasons.append('empty value')
         elif not data:
             reasons.append('missing value')
+        elif not FQDN(data, allow_underscores=True).is_valid:
+            reasons.append('{} value "{}" is not a valid FQDN'
+                           .format(_type, data))
```
ross, line 763: Does the `FQDN` method provide any information about what's wrong with the value? It might be nice (if it's easy) to provide what the user needs to do it fix the problem. If it's not easy or doesn't support it I think this is still ship-able.

ross (commented):
Looks pretty good. Comment/question inline.

octodns/record/__init__.py
PeterDaveHello: Looks like it doesn't have that rich information, just true or false about if the string is valid:

https://fqdn.readthedocs.io/en/latest/

ross (approved):

PR #632, opened 2020-11-25
Fix _is_valid_dkim_key for Python 3.9 compatibility in OVH provider

mtearle:
base64.decodestring was deprecated and removed in Python 3.9 in favour of
decodebytes (See https://bugs.python.org/issue39351 )

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


mtearle:
Coverage tests were passing under 3.9.   Will have a think overnight how to fix this, (for earlier versions), suggestions welcome.

ross (commented):
Probably makes sense to go ahead and add a 3.9 build onto the actions/ci list?

ross (approved):
Thanks!

Issue #633, opened 2020-11-27
Alias zones sync fails with KeyError


rissson:
Hi everyone!

First, thanks for this great project!

When syncing aliased zones, let's say a zone B. that is aliased to a zone A., if you only sync zone B., without also syncing zone A., it fails with:

```
$ octodns-sync --config-file=./config/config.yaml B.
Traceback (most recent call last):
  File "/home/risson/.cache/pypoetry/virtualenvs/lama-corp-dns-iy2hlWpG-py3.8/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/home/risson/.cache/pypoetry/virtualenvs/lama-corp-dns-iy2hlWpG-py3.8/lib/python3.8/site-packages/octodns/cmds/sync.py", line 38, in main
    manager.sync(eligible_zones=args.zone, eligible_sources=args.source,
  File "/home/risson/.cache/pypoetry/virtualenvs/lama-corp-dns-iy2hlWpG-py3.8/lib/python3.8/site-packages/octodns/manager.py", line 383, in sync
    desired=desired[zone_source],
KeyError: 'B.'
```

However, `octodns-sync --config-file=./config/config.yaml A. B.` succeeds (the order in which the zones are provided doesn't matter).

A basic example of it failing is already in this repo: `YAML_TMP_DIR=/tmp octodns-sync --config-file=tests/config/simple-alias-zone.yaml alias.tests.`

I'm currently in the process of adding a unit test for this case, please let me know if this behavior is intended. If it isn't, maybe it should be documented that you have to sync the _origin_ zone as well as the aliased one.

A few information:
Python 3.8.6
octoDNS 0.9.11

cc. @ross who added support for this feature

ross:
Nice catch.

> I'm currently in the process of adding a unit test for this case, please let me know if this behavior is intended.

I'd says the behavior is expected, but not having a better error message is an oversight. 

I think the fix would be to throw an error message when an aliased zone is being synced without it companion, e.g. something like:

"A. cannot be synced without B. since it's aliased."

I'm guessing a `try`/`except` around the `KeyError` would be the place to do it.

> documented that you have to sync the _origin_ zone as well as the aliased one.

Yeah. Documenting it would be a good idea as well.

rissson:
Should I add this to #634 then?

ross:
> Should I add this to #634 then?

Yeah. Would be happy to take a look at that.

rissson:
[...]

PR #634, opened 2020-11-27
manager: error when an alias zone is synced without its source

rissson:
See #633 for details

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -375,7 +375,6 @@ class Manager(object):
         futures = []
         for zone_name, zone_source in aliased_zones.items():
             source_config = self.config['zones'][zone_source]
-            print(source_config)
             try:
                 desired_config = desired[zone_source]
             except KeyError:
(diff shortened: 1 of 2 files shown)


tests/test_octodns_manager.py
```diff
@@ -191,6 +191,11 @@ def test_aliases(self):
                               'zone alias.tests. is an alias zone',
                               text_type(ctx.exception))
 
+            # Sync an alias without the zone it refers to
+            tc = Manager(get_config_filename('simple-alias-zone.yaml')) \
+                .sync(eligible_zones=["alias.tests."])
+            self.assertEquals(0, tc)
```
ross, line 197: Looks good. As mentioned in https://github.com/github/octodns/issues/633#issuecomment-735903849, I think the desired behavior here would be to throw a `ManagerException` with an error message about syncing an aliased zone without it's companion.

octodns/manager.py
```diff
@@ -375,12 +375,19 @@ def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
         futures = []
         for zone_name, zone_source in aliased_zones.items():
             source_config = self.config['zones'][zone_source]
+            print(source_config)
```
ross, line 378:
```suggestion
```

Removing debug print.

ross (approved):

PR #636, opened 2020-12-01
CloudflareProvider: Add account id and subscription parameters

lancehudson:
Adds these params to the provider:
* account_id
* subscription_id

`account_id` is needed in order to create zones in any account other than your default account. I also ran into a bug (started on the 24th of Nov) with Cloudflare, where if you have access to more than one account, the zone list api point doesn't page properly. This is a workaround while adding additional features.

**From our support case:**
```
According to our engineers, the issue here is due to incorrect traversal of lists of zones for API Tokens given access to all zones in one account, when the underlying user has access to more than one account.
They are working on a fix, however it may take a while to be deployed, due to the change freeze.

In the meanwhile, if possible, you should start to consider specifying an account ID or name (via the account.id or account.name filters, e.g. /zones?account.name=Test Account&per_page=50), which uses a more performant code-path in the majority of use-cases.
```

`subscription_id` is used when the provider creates a zone. This allows it to us to use octodns to create zones instead of secondary scripts, as zones by default use the 'Free' plan, and you have to subscribe them to a paid plan.


ross:
Hi @lancehudson looks like CI isn't happy with the changes in this PR. You can run the same things locally with something like:

```console
$ ./script/bootstrap
# ^ only required once
$ ./script/lint
...
$ ./script/coverage
...
```

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-cloudflare/. Feel free to rework/reopen this PR there.

PR #641, opened 2020-12-08
fix: error in gandi.py

0xflotus:


yzguy:
@0xflotus Thanks for the spelling fix! Looks like the test for the provider will need updated to reflect the new error message

https://github.com/github/octodns/blob/master/tests/test_octodns_provider_gandi.py#L177-L178

0xflotus:
It seems to be a good test, when it covers this small fix. Thank you ;)

ross (approved):

Issue #644, opened 2020-12-10
Support UltraDNS's Apex Alias records


nikolay-te:
It seems that currently `octodns` does not support the Apex Alias records that one can configure on UltraDNS.
E.g. trying to dump a zone from UltraDNS with such records causes `octodns-dump` to crash:

```
octodns-dump --config-file ./config/staging.yaml --output-dir ./ stg.XXX.com. ultradns                                                                                                                   
2020-12-10T16:27:50  [4492312000] INFO  Manager __init__: config_file=./config/staging.yaml
2020-12-10T16:27:50  [4492312000] INFO  Manager __init__:   max_workers=2
2020-12-10T16:27:50  [4492312000] INFO  Manager __init__:   include_meta=False
2020-12-10T16:27:51  [4492312000] INFO  Manager dump: zone=stg.XXX.com., sources=()
Traceback (most recent call last):
  File "/Users/ndenev/te/octodns/venv/bin/octodns-dump", line 8, in <module>
    sys.exit(main())
  File "/Users/ndenev/te/octodns/venv/lib/python3.9/site-packages/octodns/cmds/dump.py", line 34, in main
    manager.dump(args.zone, args.output_dir, args.lenient, args.split,
  File "/Users/ndenev/te/octodns/venv/lib/python3.9/site-packages/octodns/manager.py", line 469, in dump
    source.populate(zone, lenient=lenient)
  File "/Users/ndenev/te/octodns/venv/lib/python3.9/site-packages/octodns/provider/ultra.py", line 290, in populate
    _type = self.RECORDS_TO_TYPE[record['rrtype']]
KeyError: 'APEXALIAS (65282)'
```

ross:
Likely 2 things there:

1. The Ultra DNS provider should ignore types it/octoDNS doesn't understand/support
2. I suspect `APEXALIAS` can be mapped to octoDNS's `ALIAS` type, but I'm not familiar with Ultra DNS so can't say for sure.

/cc @phelpsw who wrote the provider

nikolay-te:
I've opened https://github.com/octodns/octodns/pull/646 which should make it so records with unknown types are ignored and are also logged as warnings.

sbrinkerhoff:
[...]

PR #646, opened 2020-12-10
Ignore UltraDNS records with unsupported rrtypes and log warning.

nikolay-te:
This is a workaround for the issue in https://github.com/octodns/octodns/issues/644
In case we encounter an unsupported record type like APEXALIAS, print a warning with the record and continue.

### DIFF
--- a/octodns/provider/ultra.py
+++ b/octodns/provider/ultra.py
@@ -291,7 +291,7 @@ class UltraProvider(BaseProvider):
                     _type = self.RECORDS_TO_TYPE[record['rrtype']]
                 except KeyError:
                     self.log.warning('populate: ignoring record with '
-                                     'unsupported rrtype=%s', record)
+                                     'unsupported rrtype, %s  %s', name, record['rrtype'])
                     continue
                 values[name][_type] = record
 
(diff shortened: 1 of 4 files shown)


octodns/provider/ultra.py
```diff
@@ -287,7 +287,12 @@ def populate(self, zone, target=False, lenient=False):
                 name = zone.hostname_from_fqdn(record['ownerName'])
                 if record['rrtype'] == 'SOA (6)':
                     continue
-                _type = self.RECORDS_TO_TYPE[record['rrtype']]
+                try:
+                    _type = self.RECORDS_TO_TYPE[record['rrtype']]
+                except KeyError:
+                    self.log.warning('populate: ignoring record with '
+                                     'unsupported rrtype=%s', record)
```
ross, line 294: Seems like this will print out the `dict` of the record which for big/involved records might be noisy and hard to read. It may be preferable to do something like

```suggestion
                                     'unsupported rrtype, %s  %s', name, record['rrtype'])
```

Feel free to tweak or change that.

ross (commented):
Looks good, thanks. Minor comment on the warning message inline.

nikolay-te:
@ross Thanks for the suggestion, I agree it could get noisy if we print the whole dict, so I've applied the proposed changes.

ross (approved):

Issue #647, opened 2020-12-11
Issue with the Route53 provider handling of CNAME records without trailing dot.


nikolay-te:
So, we have some records in Route53 which are missing the trailing dot when fetching them via the API, however these DNS queries work as expected because R53 assumes them to be FQDN:
From https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-basic.html
```
CNAME — Canonical name
The fully qualified domain name (for example, www.example.com) that you want Route 53 to return in response to DNS queries for this record. A trailing dot is optional; Route 53 assumes that the domain name is fully qualified. This means that Route 53 treats www.example.com (without a trailing dot) and www.example.com. (with a trailing dot) as identical.
```

But `octodns` is failing when pulling zones with such records:
```
octodns-dump --config-file ../dns-automation/config/staging.yaml --output-dir ./ XXX.com. route53                                                                                
2020-12-11T19:07:55  [4652064192] INFO  Manager __init__: config_file=../dns-automation/config/staging.yaml
2020-12-11T19:07:55  [4652064192] INFO  Manager __init__:   max_workers=2
2020-12-11T19:07:55  [4652064192] INFO  Manager __init__:   include_meta=False
2020-12-11T19:07:56  [4652064192] INFO  Manager dump: zone=XXX.com., sources=()
Traceback (most recent call last):
  File "/Users/ndenev/te/octodns/venv/bin/octodns-dump", line 33, in <module>
    sys.exit(load_entry_point('octodns', 'console_scripts', 'octodns-dump')())
  File "/Users/ndenev/te/octodns/octodns/cmds/dump.py", line 34, in main
    manager.dump(args.zone, args.output_dir, args.lenient, args.split,
  File "/Users/ndenev/te/octodns/octodns/manager.py", line 469, in dump
    source.populate(zone, lenient=lenient)
  File "/Users/ndenev/te/octodns/octodns/provider/route53.py", line 993, in populate
    record = Record.new(zone, name, data, source=self,
  File "/Users/ndenev/te/octodns/octodns/record/__init__.py", line 120, in new
    raise ValidationError(fqdn, reasons)
octodns.record.ValidationError: Invalid record YYY._domainkey.XXX.com.
  - CNAME value "abcdef.dkim.YYY.io" missing trailing .
```

nikolay-te:
Potential fix in https://github.com/octodns/octodns/pull/648

ross:
[...]

Issue #179, opened 2018-01-24
NS1 values for CNAME, SRV, MX, etc should be seen as absolute


yzguy:
So I think this is worth a discussion.

I brought up to NS1 support that they allow values without the `.` and they accept/return these values in the UI/API however if you query for the record via `dig` or `nslookup`, they put the `.`. It was a while ago but they seemed not receptive to it or that it wasn't an issue.

I'm proposing we add to the OctoDNS provider for NS1 that it takes any values that should have a `.`, and adds it if it's missing. So that a value with a missing `.` is seen as equivalent to a value with the `.`

This would make it so that if the `.` was missing during a dump it doesn't error saying it's missing, and if you did something like NSOne sync to Dyn, it would see the two values as equivalent and not say that Dyn is different and want to try and recreate the records without the `.`, which results in an error as that's not allowed.

Similar functionality is [already implemented in the DNSMadeEasy provider](https://github.com/yzguy/octodns/blob/564bee93f2b71e72541c0c991a3e8124315f80a2/octodns/provider/dnsmadeeasy.py#L129-L136). Focusing around dealing with actual relative values, whereas the NS1 one would just deal with "relative" values that are known to be absolute.

Would love to hear feedback about if it's a good idea or not.

**Examples:**

If you put in `relative` as the value for a CNAME, you would expect `relative.yzguy.io.` as the answer, however if you query you get the following:

```
-> dig test.yzguy.io @dns1.p06.nsone.net +short
relative.
```

If you put `relative.yzguy.io` as the value, you would expect `relative.yzguy.io.yzguy.io.`, but you get

```
-> dig test.yzguy.io @dns1.p06.nsone.net +short
relative.yzguy.io.
```

markmpeterson:
FYI - @yzguy thank you for the suggestion. I'll raise this with the lead engineers to discuss internally. :)

yzguy:
@markmpeterson Do you think it's likely this would be a fix in the NS1 package or a backend change in NS1's platform itself.

I'd imagine it might be a big deal to just force all current values to have a `.` to make them absolute in the database, etc. However that would just fix it without any code changes in anyone's code for the most part.

Other part is the validation of inputs via UI/API, so that if they don't put it then it just automatically does before saving it.

yzguy:
I think this is worth moving ahead with. It’s blocking me in some respects from using OctoDNS on a zone that has a lot of records without periods, because the input validation did not require them, and query-wise/zone transfer it gets the correct absolute value.

[...]

PR #648, opened 2020-12-11
Route53 assumes some records to be always fqdn.

nikolay-te:
Route53 assumes some records to be always fqdn even if it can store them without the trailing dot. Append the trailing dot in such cases when pulling data from their API.

This attempts to fix https://github.com/octodns/octodns/issues/647

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -944,6 +944,12 @@ class Route53Provider(BaseProvider):
                 if record_type not in self.SUPPORTS:
                     # Skip stuff we don't support
                     continue
+                # Route53 treats some records always as fully qualified
+                if record_type in ['CNAME', 'SRV', 'NS']:
+                    rrs = rrset["ResourceRecords"]
+                    rrs = [{k: v if v.endswith('.') else v + '.'
+                            for k, v in r.items()} for r in rrs]
+                    rrset["ResourceRecords"] = rrs
                 if record_name.startswith('_octodns-'):
                     # Part of a dynamic record
                     try:


yzguy:
Do you know if those are the only records with that requirement? With a few other providers it was any records that the value was pointing at another DNS record, for example [Constellix](https://github.com/octodns/octodns/blob/master/octodns/provider/constellix.py#L132-L139)

In that case it was `'ALIAS', 'CNAME', 'MX', 'NS', 'SRV'`

nikolay-te:
I think you are right. Just checked [here](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html) and it seems that for all records that point to domain names the trailing dot is optional.

ross:
See https://github.com/octodns/octodns/issues/647#issuecomment-744479445 for more info, but I believe this is a use case for `lenient` rather than having the provider(s) transparently fix things. If not we can reopen.

yzguy:
I think this might be legitimate, the "always look at values returned by API as absolute" has been implemented in a bunch of providers:

* [Constellix](https://github.com/octodns/octodns/blob/master/octodns/provider/constellix.py#L132-L139)
* [DNSMadeEasy](https://github.com/octodns/octodns/blob/master/octodns/provider/dnsmadeeasy.py#L122-L126)
* [NS1](https://github.com/octodns/octodns/blob/master/octodns/provider/ns1.py#L731-L737)
* [Gandi](https://github.com/octodns/octodns/blob/master/octodns/provider/gandi.py#L90-L96)

They aren't really fixing anything, OctoDNS validation rules ensure in YAML, etc that every thing has to be absolute. This means when new records are added they must be absolute, and are then passed to the provider as absolute in the API call.

When dumping records, or getting the records to compare to a source (Provider, etc) the values from the provider's API have to be absolute for OctoDNS to accurately compare it's expected absolute value to a relative value in a provider.

For example:

If `www` is returned from a provider API, the API on their side will automatically append `example.com.` when the record is queried, all is good. However `www` to OctoDNS is not an absolute FQDN therefore invalid, if you put `--lenient` it will ignore it, but then tell you `www.example.com.` in your YAML != `www`, therefore it must delete `www`, and add a `www.example.com.`

Issue about this issue in NS1: https://github.com/octodns/octodns/issues/179

The solution is it is the same issue is that the provider could make the value absolute before passing it on to the `_data_for_` method

ross:
> If `www` is returned from a provider API, the API on their side will automatically append `example.com.`

I didn't realize this was talking about "relative" values. I thought the associated PR was just adding `.` on the end, https://github.com/octodns/octodns/pull/648/files#diff-855a5ab1ea3cc150c71e81cf5752481b436dcdc723bfcf86c4c8c4e1bc0bf8a3R950-R951.  Relative values will definitely need to be fixed by the provider, they'd be different from just adding `.` to the end. If that's what this is about it can be reopened.

nikolay-te:
In this specific case, Route53 does not support relative values. Everything is assumed to be absolute/FQDN, be it with or without a dot at the end. Even if I create a CNAME with a target just `www` the query responses will be just `www.`
So this is similar, but not the same, and probably `--lenient` should help.

yzguy:
Ah, I misunderstood. Thanks for the clarification, definitely looks to be similar but not the same.

All good, no need to reopen.

PR #649, opened 2020-12-12
Updated document for unsafe thresholds

### LINKED ISSUE #638: Document unsafe threshold configuration
#120 added the ability to configure the unsafe percentage thresholds, but there is no documentation on how to change this. I had to blame the code to find the PR and found this helpful comment:

This PR should allow you to pass an update and/or delete threshold argument to the provider to override the default threshold of 30%. 

That would look something like:

```
  dyn:
    class: octodns.provider.dyn.DynProvider
    update_pcent_threshold: .4
    delete_pcent_threshold: .4
```

_Originally posted by @joewilliams in https://github.com/github/octodns/issues/120#issuecomment-335617349_

siddharths2710:
Resolves #638

### DIFF
--- a/docs/records.md
+++ b/docs/records.md
@@ -120,3 +120,18 @@ If you'd like to enable lenience for a whole zone you can do so with the followi
     targets:
     - ns1
 ```
+
+#### Restrict Record manipulations
+
+OctoDNS currently provides us the ability to limit the frequency of update/deletes on 
+DNS records by allowing us to configure a percentage of the allowed operations as a
+threshold parameter. If left unconfigured, suitable defaults take over instead.
+In the below example, the Dynamic provider configured accomodates only 40% of both
+update and delete operations over all the records present.
+
+````yaml
+dyn:
+    class: octodns.provider.dyn.DynProvider
+    update_pcent_threshold: 0.4
+    delete_pcent_threshold: 0.4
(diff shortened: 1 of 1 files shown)


docs/records.md
```diff
@@ -120,3 +120,18 @@ If you'd like to enable lenience for a whole zone you can do so with the followi
...
     - ns1
 ```
+
+#### Restrict Record manipulations
+
+OctoDNS currently provides us the ability to limit the frequency of update/deletes on 
+DNS records by allowing us to configure a percentage of the allowed operations as a
+threshold parameter. If left unconfigured, suitable defaults take over instead.
+In the below example, the Dynamic provider configured accomodates only 40% of both
+update and delete operations over all the records present.
```
ross, line 130:
```suggestion
OctoDNS currently provides the ability to limit the number of updates/deletes on 
DNS records by configuring a percentage of allowed operations as a threshold.
If left unconfigured, suitable defaults take over instead. In the below example, 
the Dyn provider is configured with limits of 40% on both update and 
delete operations over all the records present.
```
Minor tweaks to remove the first person perspective (use of "us") so that it matches the way the documentation is written elsewhere and a few other specific words.

ross (approved):
Thanks!

PR #654, opened 2020-12-26
Fix indent of config/example.com.yaml in README.md

PeterDaveHello:
Reference: https://github.com/octodns/octodns/blob/master/docs/records.md


ross (approved):
Both are valid YAML, but the spaces to appear to be used in other places so 👍 to making it more consistent.

PeterDaveHello:
I believe the revised one is more strict to some linters 😅 but yes, it's triggered by consistent issue here 👍

Issue #657, opened 2021-01-06
zonefile source fails under windows since files can't end in dot


ttk:
Due to a limitation in windows, filenames can't end in a dot.  The default behaviour of the `ZoneFileSource` loader is to include the dot at the end of the zone filename as per the comment:

`# Filenames should match zone name (eg. example.com.)`

My solution was to subclass the ZoneFileSource class and override the `_load_zone_file()` function to discover the zonefiles as per our naming convention.

Perhaps the default `ZoneFileSource` should support zone file lookups without the dot as well?

ross:
Interesting. Guess best bet may be to support an optional `.zone` or something like that and then try loading both. Looks like the relevant bit of code is https://github.com/octodns/octodns/blob/16241d29a6ed0f72f52381334d3c1ef66e446a45/octodns/source/axfr.py#L227-L232

ttk:
Yeah, supporting `.zone` would be helpful as that is a common naming convention and it would solve the problem on windows without any code changes.  However, supporting a custom naming convention for the zone files would still require subclassing `ZoneFileSource`, as far as I can tell.

ross:
> require subclassing `ZoneFileSource`, as far as I can tell.

I was suggesting that the `.zone` support be added to the existing (linked) provider. Would be happy to :eye: a PR to that effect.

yzguy:
I can take a look at this. It's a good catch, surprised it didn't come up before.

yzguy:
@ttk Just added this functionality to the ZoneFileSource, it's been merged into master.

PR #660, opened 2021-01-09
ZoneFileSource: allow users to specify file extension

yzguy:
Addresses #657 

The Source previously just looked for a file with the same name as the zone, eg `example.com.` in a specified directory. Windows does not allow files to end in a period, which means the default file naming does not work with Windows.

This change allows a user to specify an optional `file_extension` option in the configuration of the source, by doing this OctoDNS will look for a file with the format of `zonename+extension`, eg `example.com.zone`. This allows users to use any extension they want, which commonly `.db`, `.zone` are used with BIND zone files.

```yaml
zonefile:
  class: octodns.source.axfr.ZoneFileSource
  directory: ./zonefiles
  file_extension: zone
```

When specifying the extension, it can contain the leading `.` or not, it will be stripped off regardless, then appended to the zonename, which already contains a period.

/cc @ttk

### DIFF
--- a/octodns/source/axfr.py
+++ b/octodns/source/axfr.py
@@ -206,17 +206,24 @@ class ZoneFileSource(AxfrBaseSource):
         class: octodns.source.axfr.ZoneFileSource
         # The directory holding the zone files
         # Filenames should match zone name (eg. example.com.)
+        # with optional extension specified with file_extension
         directory: ./zonefiles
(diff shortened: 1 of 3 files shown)


ross (approved):

ttk:
Thanks @yzguy !  I'm glad you made the file extension configurable.

Issue #662, opened 2021-01-21
git checkout fails under Windows


ttk:
Similar to what was reported in #657 , because Windows doesn't support files that end in a dot/period, the octodns project fails to checkout a working branch.

```
git clone git@github.com:octodns/octodns.git
Cloning into './octodns'...
done.
error: invalid path 'tests/config/split/dynamic.tests./a.yaml'
fatal: unable to checkout working tree
```

Here are the list of all files and directories that end in a dot.

```
./tests/zones/invalid.records.
./tests/zones/invalid.zone.
./tests/zones/unit.tests.
./tests/config/split/subzone.unit.tests.
./tests/config/split/unordered.
./tests/config/split/empty.
./tests/config/split/unit.tests.
./tests/config/split/dynamic.tests.
```

My workaround is to clone the project in WSL2.

ross:
That's annoying. The new functionality can mostly work around that, but then it'll be stuck not being able to test the default/main case if we rename them all. Not sure what the best option is...

ttk:
My plan is to rename the files with the `.zone` extension so that the checkout works in Windows.   Then I'll modify the unit tests to temporarily copy the `.zone` file to a `.` file if the detected os is not Windows.  Does that sound reasonable?

ross:
> Then I'll modify the unit tests to temporarily copy the `.zone` file to a `.` file if the detected os is not Windows. Does that sound reasonable?

If you want they can just copy them and leave them on platforms that allow it and we can `.gitignore` the copies so that git won't try and add them back. That's probably easier than trying to clean up after things if something goes wrong during the tests.

ross:
https://github.com/octodns/octodns/pull/665 should  get this addressed (with the fix an underlying bug in https://github.com/octodns/octodns/pull/664.) 

@ttk if you can to pull down that branch and give it a try on a windows box please do. You can probably grab a tarball of that branch to avoid having to git clone (since that won't yet work) I don't have a windows box so I can't test it.

ross:
> ./tests/config/split/subzone.unit.tests.
> ./tests/config/split/unordered.
> ./tests/config/split/empty.
> ./tests/config/split/unit.tests.
> ./tests/config/split/dynamic.tests.

That's going to be annoying. Those are directories rather than files :-/

ross:
Reworked https://github.com/octodns/octodns/pull/665 a bit to what I believe should function on windows and still have full code coverage.

ttk:
[...]

PR #663, opened 2021-01-23
Windows Support

ttk:
Most of the changes are affect the unit tests which used several files and directories that end in a dot.  These aren't supported on Windows, so all of these files were renamed to remove the dot, or end with a valid extension.

In order to keep backward compatibility, I decided to add a new `SplitYamlAltProvider` as opposed to adding a new parameter to the constructor on `SplitYamlProvider`. 

Also includes a bug fix for the newly added custom extension support for Zone File Source, plus it's more optimal because it doesn't get the entire directory listing.

The unit tests complete successfully on Windows and Ubuntu (WSL2).  Only tested with python 3.x, not with 2.7.

I added an exclude for the code coverage, as I didn't know how to get 100% with a conditional on the `os.name` variable.  I hope this is acceptable.

This fixes issue #662 .


ross (commented):
Hi @ttk thanks for taking a pass at this. I think I'd prefer to try and keep the changes isolated to the tests rather than letting the windows file `.` problems make their way into a provider. 

Perhaps a setUp for the azure unit tests that copy the files from .zone versions over to `.` or something so that the tests work as written for non-problematic oses.

ttk:
I don't think it's possible to have the unit tests pass under Windows without changes to the core code.  For example, the Manager.dump() function will attempt to write to an invalid filename in Windows without my proposed change.  I really don't like having `if os.name == 'nt':` in the code, but it's a necessary evil if backward compatibility is important.

ross:
Addressed in https://github.com/octodns/octodns/pull/665

PR #664, opened 2021-01-25
Rename extention zonefile test to avoid existing unit.tests.

ross:
The extension handling stuff was getting tricked into working since both `unit.tests.` and `unit.tests.extension` exist.  This was due to looking for `zone_name` in the directory listing which will never exist if using extensions. We need to look for the zone_name + extension in the directory listen and then use that if it exists.

/cc https://github.com/octodns/octodns/pull/663 that lead me to discovering this problem when I tried to fix that one
/cc @yzguy

### DIFF
--- a/octodns/source/axfr.py
+++ b/octodns/source/axfr.py
@@ -229,14 +229,16 @@ class ZoneFileSource(AxfrBaseSource):
         self._zone_records = {}
 
     def _load_zone_file(self, zone_name):
+
+        zone_filename = zone_name
+        if self.file_extension:
+            zone_filename = '{}{}'.format(zone_name,
+                                          self.file_extension.lstrip('.'))
+
         zonefiles = listdir(self.directory)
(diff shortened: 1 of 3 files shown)


yzguy (approved):

octodns/source/axfr.py
```diff
@@ -229,14 +229,16 @@ def __init__(self, id, directory, file_extension=None, check_origin=True):
         self._zone_records = {}
 
     def _load_zone_file(self, zone_name):
+
+        zone_filename = zone_name
+        if self.file_extension:
+            zone_filename = '{}{}'.format(zone_name,
+                                          self.file_extension.lstrip('.'))
+
         zonefiles = listdir(self.directory)
```
ttk, line 238: Why get the listing of all files?  What if there are thousands of files in this directory?  This operation would be slow.  My suggestion is to check if the file exists  `if isfile(zone_filename):`

PR #665, opened 2021-01-25
Fix windows filename problems

ross:
Builds on https://github.com/octodns/octodns/pull/664 to fix https://github.com/octodns/octodns/issues/662.

This takes the tact of renaming all the problematic files and then copying one back to a version that ends with a `.` to test the default handling works correctly. 

/cc https://github.com/octodns/octodns/pull/663#issuecomment-766884745 which is another pass at a solution


tests/test_octodns_source_axfr.py
```diff
@@ -53,6 +54,19 @@ def test_zonefiles_with_extension(self):
         source.populate(valid)
         self.assertEquals(1, len(valid.records))
 
+    def test_zonefiles_without_extension(self):
+        try:
+            copyfile('./tests/zones/unit.tests.tst',
+                     './tests/zones/unit.tests.')
+        except:
```
ross, line 61: @ttk I'd appreciate it if you could provide the exact exception that's thrown by copyfile when it tries to copy to a filename with a `.` on the end in windows so that we don't have to catch all exceptions here. 

I would consider getting a better/tighter exception here a prerequisite for shipping this so that we can catch other/unrelated problems and fail the test rather than skip.

yzguy:
Looks good to me, easy change. Definitely agree with the exception concern. I'll keep an eye out for that, then can approve.

tests/test_octodns_source_axfr.py
ttk: Unfortunately the `copyfile()` doesn't raise any exception, instead it creates the following file: `./tests/zones/unit.tests` (transparently removes the trailing dot).

I had to do something like this to raise an exception:
```
--- a/tests/test_octodns_source_axfr.py
+++ b/tests/test_octodns_source_axfr.py
@@ -17,6 +17,7 @@ from octodns.source.axfr import AxfrSource, AxfrSourceZoneTransferFailed, \
     ZoneFileSource, ZoneFileSourceLoadFailure
 from octodns.zone import Zone
 from octodns.record import ValidationError
+import os


 class TestAxfrSource(TestCase):
@@ -58,6 +59,10 @@ class TestZoneFileSource(TestCase):
         try:
             copyfile('./tests/zones/unit.tests.tst',
                      './tests/zones/unit.tests.')
+
+            if os.path.exists('./tests/zones/unit.tests'):
+                raise Exception("Files ending in a . not supported")
+
         except:
             self.skipTest('Unable to create unit.tests. (ending with .) so '
                           'skipping default filename testing.')
```
This isn't great.  I first tried `if not os.path.exists('./tests/zones/unit.tests.'):` but that returned `True`.

ross: > This isn't great. I first tried `if not os.path.exists('./tests/zones/unit.tests.'):` but that returned `True`.

Yuk.

I'll probably rework this a bit and get rid of the try/catch then since it's not going to except and just skip test if the file without the `.` exists.

ttk:
I tested and it's almost perfect.  One small change in `def test_zone_directory(self):` to make the assert pass under windows:
```
--- a/tests/test_octodns_provider_yaml.py
+++ b/tests/test_octodns_provider_yaml.py
@@ -213,7 +213,7 @@ class TestSplitYamlProvider(TestCase):

         self.assertEqual(
-            join(dirname(__file__), 'config/split/unit.tests.tst'),
+            join(dirname(__file__), 'config/split','unit.tests.tst'),
             source._zone_directory(zone))

     def test_apply_handles_existing_zone_directory(self):
```
This is so that the directory separators are consistent.

yzguy (approved):

ross:
Looking to see if we can add windows builds over in https://github.com/octodns/octodns/pull/672. Right now it seems like it's going to slow things down a lot (it's just waiting to build for a long time,) but we'll see how that goes.

Otherwise going to ship this...

patcon:
Seems that the workaround on windows is to have domain-specific folders ending in `.tst` instead of `.`, yes? I love that this is already resolved, but hoping to make it a tiny bit clearer. (I'm looking to have a loose group of people manage DNS via public GitHub repo, and the less things that provoke "whys", the better 🙂 )

Would it be ok to submit a change that:
- looks for dot-ending folders
- looks for `.tst` ending folders
- any folder has dot-ending implied (enabled via root config, perhaps)

This would allow this:

```
.
├── domains
│   ├── example.com
│   │   └── example.com.yaml
│   └── example.net
│       └── example.net.yaml
└── octodns-config.yaml
```

ross:
> domain-specific folders ending in `.tst` instead of `.`, yes

The `.tst` was chosen, short for "test", for no particular reasons. If I was picking I wouldn't make it the standard or anything. If I was picking/defining a standard I'd probably go with something more like `.zone` or `.dns` as the standard ending when windows support is required with the `SplitYamlProvider` or other providers that want directories that end in `.` by default.

> Would it be ok to submit a change that:

Don't think that should be necessary here/for your use case. It seems like you'd just need the standard `YamlProvider` which doesn't rely on things ending in `.` and will work on all OSs without any special care. There's only a couple of providers that want directories/files with `.` on the end and they're fairly esoteric and wouldn't seem to make sense for what you're describing.

patcon:
Thanks for the response @ross! Might I be able to submit a PR to make `.zone` and/or `.dns` work as well? These seem clear enough (esp `.dns` as that seems less an implementation term than "zone", that kinda feels like it floated into a layer of abstraction where non-sysops users might engage)

> It seems like you'd just need the standard YamlProvider which doesn't rely on things ending in . and will work on all OSs without any special care

I'm really liking the different folders/files per subdomain, as a separation of concern that we get from `SplitYamlProvider`, esp when we later hit the point where we start using a `CODEOWNERS` file in our repo usage pattern:

https://github.com/g0v-network/domains

ross:
> Might I be able to submit a PR to make `.zone` and/or `.dns` work as well?

The value used is configurable, e.g. 

```yaml
providers:
  config:
    class: octodns.provider.yaml.SplitYamlProvider
    directory: ./config
    extension: .zone
```

So those are already options.

patcon:
Thank you thank you thank you thank you thank you thank you 🙌

Issue #640, opened 2020-12-05
Correct handling of NULL SRV records in Cloudflare Provider (possibly elsewhere)


mtearle:
Try into to sync the following entry from the YamlProvider to the Cloudflare provider

```
_pop3._tcp:
  octodns:
    cloudflare:
      proxied: false
  type: SRV
  value:
    port: 0
    priority: 0
    target: .
    weight: 0
```

Relevant part of trace here:

```
2020-12-05T23:01:56  [3074897664] DEBUG CloudflareProvider[cloudflare] _request: method=POST, path=/zones/f6c330b6c72b895026886d48e30ee7ec/dns_records
2020-12-05T23:01:56  [3074897664] DEBUG urllib3.connectionpool https://api.cloudflare.com:443 "POST /client/v4/zones/f6c330b6c72b895026886d48e30ee7ec/dns_records HTTP/1.1" 400 None
2020-12-05T23:01:56  [3074897664] DEBUG CloudflareProvider[cloudflare] _request:   status=400
2020-12-05T23:01:56  [3074897664] DEBUG CloudflareProvider[cloudflare] _request:   data={'data': {'service': '_pop3', 'proto': '_tcp', 'name': 'ucc.gu.uwa.edu.au.', 'priority': 0, 'weight': 0, 'port': 0, 'target': ''}, 'name': '_pop3._tcp.ucc.gu.uwa.edu.au', 'type': 'SRV', 'ttl': 3600}
Traceback (most recent call last):
  File "/home/wheel/mtearle/hack/mooneye/octodns/env/bin/octodns-sync", line 11, in <module>
    load_entry_point('octodns==0.9.11', 'console_scripts', 'octodns-sync')()
  File "/home/wheel/mtearle/hack/mooneye/octodns/env/lib/python3.9/site-packages/octodns-0.9.11-py3.9.egg/octodns/cmds/sync.py", line 38, in main
  File "/home/wheel/mtearle/hack/mooneye/octodns/env/lib/python3.9/site-packages/octodns-0.9.11-py3.9.egg/octodns/manager.py", line 418, in sync
  File "/home/wheel/mtearle/hack/mooneye/octodns/env/lib/python3.9/site-packages/octodns-0.9.11-py3.9.egg/octodns/provider/base.py", line 95, in apply
[...]

PR #668, opened 2021-01-30
Fix handling of NULL SRV records in Cloudflare and Digital Ocean providers, add warnings to DNS Made Easy and DNSimple providers

mtearle:
[RFC2782 ](https://tools.ietf.org/html/rfc2782)permits a SRV record target of "." to "." mean that a service is decidedly not available at this domain.

[RFC6186 ](https://tools.ietf.org/html/rfc6186)provides an example of this to indicate that non-TLS versions of POP3 and IMAP are unavailable at a domain.

This pull request fixes the handling of SRV records in Cloudflare and Digital Ocean providers, add warnings to DNS Made Easy and DNSimple providers.   Support across the other providers has been checked.

Fixes: #640

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -239,11 +239,13 @@ class CloudflareProvider(BaseProvider):
     def _data_for_SRV(self, _type, records):
         values = []
         for r in records:
+            target = ('{}.'.format(r['data']['target'])
+                      if r['data']['target'] != "." else ".")
             values.append({
                 'priority': r['data']['priority'],
                 'weight': r['data']['weight'],
                 'port': r['data']['port'],
-                'target': '{}.'.format(r['data']['target']),
+                'target': target,
             })
         return {
             'type': _type,
@@ -405,6 +407,8 @@ class CloudflareProvider(BaseProvider):
             name = subdomain
 
         for value in record.values:
+            target = value.target[:-1] if value.target != "." else "."
+
             yield {
                 'data': {
                     'service': service,
@@ -413,7 +417,7 @@ class CloudflareProvider(BaseProvider):
                     'priority': value.priority,
                     'weight': value.weight,
                     'port': value.port,
-                    'target': value.target[:-1],
+                    'target': target,
                 }
             }
 
--- a/octodns/provider/digitalocean.py
+++ b/octodns/provider/digitalocean.py
@@ -186,10 +186,14 @@ class DigitalOceanProvider(BaseProvider):
     def _data_for_SRV(self, _type, records):
         values = []
         for record in records:
+            target = (
+                '{}.'.format(record['data'])
+                if record['data'] != "." else "."
+            )
             values.append({
                 'port': record['port'],
                 'priority': record['priority'],
-                'target': '{}.'.format(record['data']),
+                'target': target,
                 'weight': record['weight']
             })
         return {
--- a/octodns/provider/dnsimple.py
+++ b/octodns/provider/dnsimple.py
@@ -280,7 +280,6 @@ class DnsimpleProvider(BaseProvider):
         values = defaultdict(lambda: defaultdict(list))
         for record in self.zone_records(zone):
             _type = record['type']
-            data_for = getattr(self, '_data_for_{}'.format(_type), None)
             if _type not in self.SUPPORTS:
                 self.log.warning(
                     'populate: skipping unsupported %s record',
@@ -325,7 +324,7 @@ class DnsimpleProvider(BaseProvider):
                 )
                 return False
 
-        return record._type in self.SUPPORTS
+        return super(DnsimpleProvider, self).supports(record)
 
     def _params_for_multiple(self, record):
         for value in record.values:
--- a/octodns/provider/dnsimple.py
+++ b/octodns/provider/dnsimple.py
@@ -218,12 +218,23 @@ class DnsimpleProvider(BaseProvider):
             try:
                 weight, port, target = record['content'].split(' ', 2)
             except ValueError:
-                # see _data_for_NAPTR's continue
+                # their api/website will let you create invalid records, this
+                # essentially handles that by ignoring them for values
+                # purposes. That will cause updates to happen to delete them if
+                # they shouldn't exist or update them if they're wrong
+                self.log.warning(
+                    '_data_for_SRV: unsupported %s record (%s)',
+                    _type,
+                    record['content']
+                )
                 continue
+
+            target = '{}.'.format(target) if target != "." else "."
+
             values.append({
                 'port': port,
                 'priority': record['priority'],
-                'target': '{}.'.format(target),
+                'target': target,
                 'weight': weight
             })
         return {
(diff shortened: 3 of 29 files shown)


octodns/provider/dnsmadeeasy.py
```diff
@@ -265,9 +265,18 @@ def populate(self, zone, target=False, lenient=False):
...
-            if _type not in self.SUPPORTS:
-                self.log.warning('populate: skipping unsupported %s record',
-                                 _type)
+            data_for = getattr(self, '_data_for_{}'.format(_type), None)
+            if not data_for or not self.supports(
+                Record.new(
+                    zone,
+                    record['name'],
+                    data_for(_type, [record, ])
+                )
```
ross, line 274: Creating the full `Record` here will mean that any validation problems with the data that comes down will cause exceptions and won't be ignored as they normally would be lenient. From what I can tell by looking at `supports()` there doesn't seem to be anything that strictly requires it to be a `Record` object. Can that method be refactor to just pass in the necessary data, probably the output of `data_for`, and work with that?

This also results in double-running `data_for` and create the `Record` object twice.  It may make sense to just leave this block as it originally was and then go down into the for that's creating records below and add in a call to `supports()` around here https://github.com/octodns/octodns/pull/668/files#diff-f9649e542e179d4617a1512b9cedc54e1fd1b0d4e680a19150c8f2458e1208c9R289. It should be fine to "skip" records down there too.

octodns/provider/dnsmadeeasy.py
```diff
@@ -284,6 +293,29 @@ def populate(self, zone, target=False, lenient=False):
...
+        # octodns.provider.dnsmadeeasy.DnsMadeEasyClientBadRequest:
+        #      - Record value may not be a standalone dot.
+        #
+        # Skip the record and continue
+        def __extract_targets(data):
+            if 'value' in data:
+                return [data['value']['target'], ]
+            return [value['target'] for value in data['values']]
+
+        if record._type == "SRV" and "." in __extract_targets(record.data):
```
ross, line 310: Don't believe this local helper function is needed and it makes the code a little harder to follow:

```suggestion
        if value in data:
            targets = (data['value']['target'],)
        else:
            targets = [value['target'] for value in data['values']]

        if record._type == "SRV" and "." in targets:
```

mtearle:
Thanks @ross for the review, slowly working through the necessary changes (out of time this evening, but will continue tomorrow)

octodns/provider/cloudflare.py
```diff
@@ -243,7 +243,10 @@ def _data_for_SRV(self, _type, records):
                 'priority': r['data']['priority'],
                 'weight': r['data']['weight'],
                 'port': r['data']['port'],
-                'target': '{}.'.format(r['data']['target']),
+                'target': (
+                    '{}.'.format(r['data']['target'])
+                    if r['data']['target'] != "." else "."
+                ),
```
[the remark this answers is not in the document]
mtearle: Noted, will move the assignment outside the dict

octodns/provider/cloudflare.py
```diff
@@ -413,7 +416,10 @@ def _contents_for_SRV(self, record):
                     'priority': value.priority,
                     'weight': value.weight,
                     'port': value.port,
-                    'target': value.target[:-1],
+                    'target': (
+                        value.target[:-1]
+                        if value.target != "." else "."
+                    ),
```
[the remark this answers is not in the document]
mtearle: Noted, will moved the assignment outside the dict

octodns/provider/digitalocean.py
```diff
@@ -189,7 +189,10 @@ def _data_for_SRV(self, _type, records):
             values.append({
                 'port': record['port'],
                 'priority': record['priority'],
-                'target': '{}.'.format(record['data']),
+                'target': (
+                    '{}.'.format(record['data'])
+                    if record['data'] != "." else "."
+                ),
```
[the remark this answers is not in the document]
mtearle: Noted, will moved the assignment outside the dict

octodns/provider/dnsimple.py
```diff
@@ -218,12 +218,20 @@ def _data_for_SRV(self, _type, records):
...
+                    '_data_for_SRV: unsupported %s record (%s)',
+                    _type,
+                    record['content']
+                )
                 continue
             values.append({
                 'port': port,
                 'priority': record['priority'],
-                'target': '{}.'.format(target),
+                'target': ('{}.'.format(target) if target != "." else "."),
```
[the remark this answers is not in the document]
mtearle: Noted, will moved the assignment outside the dict

octodns/provider/dnsmadeeasy.py
```diff
@@ -265,9 +265,18 @@ def populate(self, zone, target=False, lenient=False):
...
+                Record.new(
+                    zone,
+                    record['name'],
+                    data_for(_type, [record, ])
+                )
+            ):
+                self.log.warning(
+                    'populate: skipping unsupported %s record',
+                    _type
+                )
```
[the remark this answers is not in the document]
mtearle: In the updated/rebased  commits I have refactored the warning

changes() in zone.py will emit the warning about the unsupported record

octodns/provider/dnsmadeeasy.py
mtearle: changes() in zone.py will emit the warning about the unsupported record at a later stage, so that code has now been refactored away.

ross:
The functionality that would let me merge master in is a checkbox, 

Screen Shot 2021-02-16 at 7 51 49 AM

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I've removed 2.7 CI to just punt on this problem so once this is updated to master one way or another it should be unblocked.

mtearle:
Hi @ross 

Apologies for the slow reply, have been unwell for the past couple of days.    As that repo is under an org account (as opposed to personal), I've bumped up the level of access to maintainer (as there was no checkbox), so hopefully that should work this time.

ross:
> Apologies for the slow reply, have been unwell for the past couple of days. As that repo is under an org account (as opposed to personal), 

Ah.

> I've bumped up the level of access to maintainer (as there was no checkbox), so hopefully that should work this time.

That appears to have done it.

octodns/provider/dnsimple.py
```diff
@@ -269,7 +280,12 @@ def populate(self, zone, target=False, lenient=False):
         values = defaultdict(lambda: defaultdict(list))
         for record in self.zone_records(zone):
             _type = record['type']
+            data_for = getattr(self, '_data_for_{}'.format(_type), None)
```
ross, line 283: This `data_for` appears unused in the block up here. 

```suggestion
```

octodns/provider/dnsimple.py
```diff
@@ -290,6 +306,27 @@ def populate(self, zone, target=False, lenient=False):
...
+                targets = [value['target'] for value in record.data['values']]
+
+            if "." in targets:
+                self.log.warning(
+                    'supports: unsupported %s record with target (%s)',
+                    record._type, targets
+                )
+                return False
+
+        return record._type in self.SUPPORTS
```
ross, line 328: It's preferable to go ahead and call up to the super method here rather than duplicate it's check in case that ever changes and gets more involved in the future. 

```suggestion
        return super(DnsimpleProvider, self).supports(record)
```

octodns/provider/dnsmadeeasy.py
```diff
@@ -284,6 +284,30 @@ def populate(self, zone, target=False, lenient=False):
...
+                targets = [value['target'] for value in record.data['values']]
+
+            if "." in targets:
+                self.log.warning(
+                    'supports: unsupported %s record with target (%s)',
+                    record._type, targets
+                )
+                return False
+
+        return record._type in self.SUPPORTS
```
ross, line 309: Same here.

```suggestion
        return super(DnsMadeEasyProvider, self).supports(record)
```

ross (approved):
Looks good. Few more minor things inline w/suggestions. Going to make them and then commit to move this forward.

PR #669, opened 2021-02-02
Modify Azure DNS Client logic to lazy load

rreichel3:
Lazy loading the Azure DNS client ensures that the necessary network calls only occur when it is time to actually do something with the client.

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -327,19 +327,19 @@ class AzureProvider(BaseProvider):
         if not self._dns_client_handle:
             # Not initialized yet, we need to do that.
             credentials = ServicePrincipalCredentials(
-                self._dns_client_client_id, 
-                secret=self._dns_client_key, 
+                self._dns_client_client_id,
+                secret=self._dns_client_key,
                 tenant=self._dns_client_directory_id
             )
             self._dns_client_handle = DnsManagementClient(
-                credentials, 
+                credentials,
                 self._dns_client_subscription_id
             )
-            
+
         return self._dns_client_handle
+
     def _set_dns_client(self, client)
         self.dns_client_handle = client
-            
 
     def __init__(self, id, client_id, key, directory_id, sub_id,
                  resource_group, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


octodns/provider/azuredns.py
```diff
@@ -330,13 +330,31 @@ def __init__(self, id, client_id, key, directory_id, sub_id,
...
-            client_id, secret=key, tenant=directory_id
-        )
-        self._dns_client = DnsManagementClient(credentials, sub_id)
+        # Store necessary initialization params
+        self._dns_client_handle = None
+        self._dns_client_client_id = client_id
+        self._dns_client_key = key
+        self._dns_client_directory_id = directory_id
+        self._dns_client_subscription_id = sub_id
+        self._dns_client = None
```
ross, line 339: Since the original property is `_dns_client` we'll need to double `_` here to avoid collision.

```suggestion
        self.__dns_client = None
```

octodns/provider/azuredns.py
```diff
@@ -330,13 +330,31 @@ def __init__(self, id, client_id, key, directory_id, sub_id,
...
+
         self._resource_group = resource_group
         self._azure_zones = set()
 
+    @property
+    def _dns_client(self)
+        if self._dns_client is None:
```
ross, line 346:
```suggestion
        if self.__dns_client is None:
```

octodns/provider/azuredns.py
```diff
@@ -330,13 +330,31 @@ def __init__(self, id, client_id, key, directory_id, sub_id,
...
+            credentials = ServicePrincipalCredentials(
+                self._dns_client_client_id,
+                secret=self._dns_client_key,
+                tenant=self._dns_client_directory_id
+            )
+            self._dns_client = DnsManagementClient(
```
ross, line 352:
```suggestion
            self.__dns_client = DnsManagementClient(
```

octodns/provider/azuredns.py
```diff
@@ -330,13 +330,31 @@ def __init__(self, id, client_id, key, directory_id, sub_id,
...
+                credentials,
+                self._dns_client_subscription_id
+            )
+        return self._dns_client
```
ross, line 356:
```suggestion
        return self.__dns_client
```

ross (commented):
Looks good, minor fix to avoid collision suggested inline.

ross (approved):

ross (approved):

PR #670, opened 2021-02-04
Force the value passed to FQDN to be a str

ross:
When working with python 2.7 and `unicode` strings FQDN complains that what's passed to it isn't a `str`. This should work around that for as long as we support python 2.7.

/cc @rreichel3 for 👍


rreichel3 (approved):
Verified this fixes the issue!

PR #672, opened 2021-02-05
Try out building things on windows too

ross:


ross:
This might not be as useful as I'd hope. I intentionally started this branch before the windows fixes were merged hoping that it'd fail and a merge in of master would fix it, but that doesn't seem to be the case.

Windows all passed, it looks like the checkout worked fine, guess maybe it's silently removing the `.` from the filenames?

https://github.com/octodns/octodns/pull/672/checks?check_run_id=1841691490#step:2:381

But `./script/cibuild` seems to have passed without actually doing anything. It is a bash script, but I'd expect it to fail and not just do nothing...

https://github.com/octodns/octodns/pull/672/checks?check_run_id=1841691490#step:5:2

/cc @ttk

Issue #656, opened 2021-01-03
Plan should order changes consistently, e.g. Deletes before Adds etc. 


ross:
Some providers have to manually ensure that deletes come first to avoid problems like their providers throwing errors about duplicate A/CNAME when making a type change.

We should look into a default behavior of consistently ordering `Plan` `changes` to make it easier on providers/more likely that they do the right thing out of the gate. If not by default we should add helpers to the `Plan` object to allow them to request the ordered version: e.g. `plan.ordered_changes`. 

/cc https://github.com/octodns/octodns/pull/587#pullrequestreview-560726591

mtearle:
@ross This was the approach I took on the Cloudflare provider when I hit this issue when putting together my other pull request.  https://github.com/octodns/octodns/pull/635/commits/0266b3e452f6bfe7669cc44ecc3c18b9f64d7e25

deric:
@mtearle That's great (works fine, at least in my case), but it doesn't change the plan. It would be better if the plan would be the same as the ordering of the actual changes.

PR #675, opened 2021-02-11
PowerDNS: Delete records before creating or modifying records

snh:
This PR is related to the overall ordering issue discussed in https://github.com/octodns/octodns/issues/656, though solves this specifically for the PowerDNS provider.

When replacing an `A` record with a `CNAME` record or vice-versa, the existing record must be removed in a PowerDNS `DELETE` change first, otherwise the PowerDNS `REPLACE` change will fail, as a `CNAME` and `A` record cannot co-exist, even briefly.

/cc @ross @dbussink

### DIFF
--- a/octodns/provider/powerdns.py
+++ b/octodns/provider/powerdns.py
@@ -386,7 +386,7 @@ class PowerDnsBaseProvider(BaseProvider):
         # Ensure that any DELETE modifications always occur before any REPLACE
         # modifications. This ensures that an A record can be replaced by a
         # CNAME record and vice-versa.
-        mods = sorted(mods, key=itemgetter('changetype'))
+        mods.sort(key=itemgetter('changetype'))
 
         self.log.debug('_apply:   sending change request')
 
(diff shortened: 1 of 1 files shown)


octodns/provider/powerdns.py
```diff
@@ -381,6 +382,12 @@ def _apply(self, plan):
         for change in changes:
             class_name = change.__class__.__name__
             mods.append(getattr(self, '_mod_{}'.format(class_name))(change))
+
+        # Ensure that any DELETE modifications always occur before any REPLACE
+        # modifications. This ensures that an A record can be replaced by a
+        # CNAME record and vice-versa.
+        mods = sorted(mods, key=itemgetter('changetype'))
```
snh, line 389: I'm relying on the fact that PowerDNS only has two valid values for `changetype`, being `DELETE` or `REPLACE`, and `DELETE` is ahead of `REPLACE` alphabetically.

Ref: https://doc.powerdns.com/authoritative/http-api/zone.html#rrset

octodns/provider/powerdns.py
```diff
@@ -7,6 +7,7 @@
 
 from requests import HTTPError, Session
 import logging
+from operator import itemgetter
```
ross, line 10:
```suggestion
from operator import itemgetter
import logging
```

ross (approved):
LGTM.

PR #676, opened 2021-02-16
Test 2.7 CI

mtearle:
test test


mtearle:
Oops, my bad did the request against the wrong target :(

PR #679, opened 2021-03-10
Update branch name

xt0rted:
Looks like the default branch name has changed since adding this link to the readme.


ross (approved):

ross:
Weird. Looks like the actions jobs didn't kick off on this and I can't find a way to re-queue them...

xt0rted:
@ross the workflow is set to ignore markdown files https://github.com/octodns/octodns/blob/765766b1bf2bcdbe47edb876ce59c461ee2aba30/.github/workflows/main.yml#L2-L5

ross:
> @ross the workflow is set to ignore markdown files

Ah. That makes sense. Annoying the PR process still requires them then.. I'll probably remove that bit and just build them regardless. Thanks.

PR #686, opened 2021-03-31
Change the auth value to have a proper string not a binary rep

lksoft:
I was unable to actually authorize to EasyDNS because the Authorization Header was sending data as b'[encoded_data]' instead of the actual string needed. So I changed that so that I could authorize properly.


ross:
Doesn't look like I have permissions to update this branch to master (so that it can run CI and be merged) so you'll need to do so.

lksoft:
Hi Ross, I think I updated properly. This is the first time I have done a Pull Request here in GitHub. I updated my branch to have the latest commits from master here and merged it into master on my fork. Is that correct?

PR #687, opened 2021-03-31
Enable NS support for non-root NS records

maikelpoot:
Transip api support NS type for non-root records.

This PR will just enable NS support , provider was prepared for supporting NS type records.


ross (approved):

ross:
LGTM.

PR #689, opened 2021-04-07
Capture plan output in File as argument to sync

cfunkhouser:
Capture plan output in File as argument to sync.

The `Manager.sync` method accepts a `plan_output_fh` argument, which is passed
to `_PlanOutput.run` as `fh`. This latter functionality already exists. The
goal is to make it easier to capture plan output in a human-readable format
as part of CI/CD automation.

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -270,8 +270,9 @@ class Manager(object):
     def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
              dry_run=True, force=False, plan_output_fh=stdout):
         self.log.info('sync: eligible_zones=%s, eligible_targets=%s, '
-                      'dry_run=%s, force=%s', eligible_zones, eligible_targets,
-                      dry_run, force)
+                      'dry_run=%s, force=%s, plan_output_fh=%s',
+                      eligible_zones, eligible_targets, dry_run, force,
+                      plan_output_fh)
 
         zones = self.config['zones'].items()
         if eligible_zones:
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -272,7 +272,7 @@ class Manager(object):
         self.log.info('sync: eligible_zones=%s, eligible_targets=%s, '
                       'dry_run=%s, force=%s, plan_output_fh=%s',
                       eligible_zones, eligible_targets, dry_run, force,
-                      plan_output_fh)
+                      getattr(plan_output_fh, 'name', plan_output_fh.__class__.__name__))
 
         zones = self.config['zones'].items()
         if eligible_zones:
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -9,6 +9,7 @@ from concurrent.futures import ThreadPoolExecutor
 from importlib import import_module
 from os import environ
 from six import text_type
+from sys import stdout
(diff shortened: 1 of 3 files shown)


octodns/manager.py
```diff
@@ -267,16 +268,17 @@ def _populate_and_plan(self, zone_name, sources, targets, desired=None,
...
 
     def sync(self, eligible_zones=[], eligible_sources=[], eligible_targets=[],
-             dry_run=True, force=False):
+             dry_run=True, force=False, plan_output_fh=stdout):
         self.log.info('sync: eligible_zones=%s, eligible_targets=%s, '
-                      'dry_run=%s, force=%s', eligible_zones, eligible_targets,
-                      dry_run, force)
+                      'dry_run=%s, force=%s, plan_output_fh=%s',
+                      eligible_zones, eligible_targets, dry_run, force,
+                      plan_output_fh)
```
ross, line 275: Looks like the default `__repr__` of fh objects is fairly verbose:

```
>>> with open('/tmp/foo', 'w') as fh:
...     print(f'{fh}')
...
<_io.TextIOWrapper name='/tmp/foo' mode='w' encoding='UTF-8'>
>>> print(f'{stdout}')
<_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
```

Maybe preferable to log `plan_output_fh.name` here? Not 100% sure that file handle objects always have a `.name` though...

```suggestion
                      'dry_run=%s, force=%s, plan_output_fh=%s',
                      eligible_zones, eligible_targets, dry_run, force,
                      plan_output_fh.name)
```

tests/test_octodns_manager.py
```diff
@@ -8,14 +8,17 @@
...
 from octodns.record import Record
 from octodns.manager import _AggregateTarget, MainThreadExecutor, Manager, \
     ManagerException
 from octodns.yaml import safe_load
 from octodns.zone import Zone
 
+from mock import MagicMock, patch
+from unittest import TestCase
+
+
```
ross, line 21:
```suggestion
```

ross (commented):
Looks reasonable. One suggestion for consideration inline.

octodns/manager.py
cfunkhouser: Unfortunately, `.name` isn't reliable. One of my primary use-cases is `File`-compatible types like `StringIO`, which does not have that attribute.

Alternatively, I could leave this out of the log line altogether, or indicate whether it's set to the default value at call time. Do either of those options seem better?

ross: Maybe something like this:

```suggestion
                      'dry_run=%s, force=%s, plan_output_fh=%s',
                      eligible_zones, eligible_targets, dry_run, force,
                      getattr(plan_output_fh, 'name', plan_output_fh.__class__.__name__))
```

```
>>> fh = StringIO()
>>> getattr(fh, 'name', fh.__class__.__name__)
'StringIO'
>>> with open('/tmp/foo', 'w') as fh:
...     getattr(fh, 'name', fh.__class__.__name__)
...
'/tmp/foo'
```

Otherwise probably make sense to just omit it.

ross (approved):
:cool:, this looks good.

PR #691, opened 2021-04-12
Hetzner provider

ricardbejarano:
Commited initial work.

Pull request status:
- [x] Implement `HetznerClient` API client class (no extra dependencies required)
- [x] Implement `HetznerProvider` provider class
- [x] Write tests
- [x] Add to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,7 @@ The above command pulled the existing data out of Route53 and placed the results
 | [EnvVarSource](/octodns/source/envvar.py) | | TXT | No | read-only environment variable injection |
 | [GandiProvider](/octodns/provider/gandi.py) | | A, AAAA, ALIAS, CAA, CNAME, DNAME, MX, NS, PTR, SPF, SRV, SSHFP, TXT | No | |
 | [GoogleCloudProvider](/octodns/provider/googlecloud.py) | google-cloud-dns | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, TXT  | No | |
+| [HetznerProvider](/octodns/provider/hetzner.py) | | A, AAAA, CAA, CNAME, MX, NS, SRV, TXT | No | |
 | [MythicBeastsProvider](/octodns/provider/mythicbeasts.py) | Mythic Beasts | A, AAAA, ALIAS, CNAME, MX, NS, SRV, SSHFP, CAA, TXT | No | |
 | [Ns1Provider](/octodns/provider/ns1.py) | ns1-python | All | Yes | Missing `NA` geo target |
 | [OVH](/octodns/provider/ovh.py) | ovh | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT, DKIM | No | |
--- a/octodns/provider/hetzner.py
+++ b/octodns/provider/hetzner.py
@@ -2,7 +2,16 @@
 #
 #
 
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import logging
+from collections import defaultdict
+
 from requests import Session
+
+from ..record import Record
+from .base import BaseProvider
 
 
 class HetznerClientException(Exception):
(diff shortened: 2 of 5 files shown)


ricardbejarano:
I have a working implementation of `HetznerProvider` but I'm yet to manually test it more throroughly.

Basic `A`, `TXT` and `CNAME` creates/updates/deletes work just fine, as well as zone creation, but I want to see where it breaks with more complicated records such as `MX` or `CAA`.

Formal tests are still a big unknown right now.

yzguy:
@ricardbejarano You can use DigitalOceanProvider as a reference for tests. When I wrote providers I didn't start the tests till after the provider was completely done. I did enough testing against the actual provider that it worked as expected. The tests just confirmed it working correctly or occasionally would make a bug apparent.

ricardbejarano:
Yup, I'm using de DO provider as a reference for everything else, might as well do that.

ricardbejarano:
@yzguy I think there's a bug [here](https://github.com/octodns/octodns/blob/master/octodns/provider/digitalocean.py#L265). CAA records are formated as:

```
flags tag "value"
```

Where `value` is a regular string that may or may not end with a hostname, [examples](https://en.wikipedia.org/wiki/DNS_Certification_Authority_Authorization#Examples).

Since you're dotting the end of the string, if, for instance, you try to deny all issuance by creating a record like:
```
0 issue ";"
```
The `;` will have a `.` appended at the end of the string in DigitalOcean and will end up like:
```
0 issue ";."
```

ricardbejarano:
Raised #692 for ^

ricardbejarano:
Pushed a fully functional `HetznerProvider` implementation. Pretty much a copy+paste+adapt from `DigitalOceanProvider`.
Tested manually for all supported record types.

Given that Hetzner is pretty lax about restricting the value of records I've added some extra whitespace handling for multi-field records (`CAA`, `MX`, `SRV`) that we may find useful somewhere else across OctoDNS. See [this](https://github.com/octodns/octodns/pull/691/files#diff-14acda1a2539c898ecc2df2ebc52dcb9fa21a933e23be6cac4f2de92dc30c417R160), [this](https://github.com/octodns/octodns/pull/691/files#diff-14acda1a2539c898ecc2df2ebc52dcb9fa21a933e23be6cac4f2de92dc30c417R186) and [this](https://github.com/octodns/octodns/pull/691/files#diff-14acda1a2539c898ecc2df2ebc52dcb9fa21a933e23be6cac4f2de92dc30c417R212).
I'm open to contributing those somewhere available for all providers to reuse. Same with [HetznerProvider._append_dot](https://github.com/octodns/octodns/pull/691/files#diff-14acda1a2539c898ecc2df2ebc52dcb9fa21a933e23be6cac4f2de92dc30c417R143), which given how picky (and rightfully so) OctoDNS is about dots on hostnames I think should be provided for all providers to reuse.

Will begin work on tests whenever possible.

ross:
> Given that Hetzner is pretty lax about restricting the value of records I've added some extra whitespace handling for multi-field records (`CAA`, `MX`, `SRV`) that we may find useful somewhere else across OctoDNS

For the most part we've taken the approach of letting octoDNS "clean up" and "standardize" things when providers are more lax. Don't recall anything re those specific cases though. E.g. I would expect that if a provider allows multiple trailing whitespaces after a CNAME value, `foo.bar.com.   ` and someone pulled that down into octoDNS and then ran a plan that it could/would show a change to `foo.bar.com.` were it applied.

ricardbejarano:
Yup that's precisely what the `_data_for` methods do. Which is somewhat tricky for multi-field records because unlike DO's API, Hetzner returns record values in the format they have over the wire.

This means you could have an `MX` record like `  10   whatever.com  `, which internally we want to split into `priority` and `exchange` and ignore whitespace.

Additionally, if both values remain the same as in OctoDNS' zone config, whitespace differences don't produce a change, if for some reason the user needs all that whitespace there.

ricardbejarano:
Keepaliving this thread, haven't forgot about it, but I haven't had time to dive into it just yet.

ricardbejarano:
Coverage at 72%, only missing `apply()` tests.

One last stretch! 💪

ricardbejarano:
@yzguy I'm having trouble understanding how `apply(...)` [here](https://github.com/octodns/octodns/blob/master/tests/test_octodns_provider_digitalocean.py#L168) is mocked.

Does `apply` there not make any calls?
Otherwise, how are they mocked? It isn't until some lines below where you specify the calls to mock.

My `apply` calls `_apply_Create` and thus makes calls that must be mocked somewhere. Where do you mock yours?

```python
    def test_apply(self):
        provider = HetznerProvider('test', 'token')

        resp = Mock()
        resp.json = Mock()
        provider._client._do_raw = Mock(return_value=resp)

        # non-existent domain, create everything
        resp.json.side_effect = [
            HetznerClientNotFound,  # no zone in populate
            HetznerClientNotFound,  # no zone during apply
            {"zone": {"id": "string"}}
        ]
        plan = provider.plan(self.expected)

        # No root NS, no ignored, no excluded, no unsupported
        n = len(self.expected.records) - 9
        self.assertEquals(n, len(plan.changes))
        self.assertEquals(n, provider.apply(plan))  # call to provider.apply(...) fails (see traceback below)
        self.assertFalse(plan.exists)
```

```
======================================================================
ERROR: test_apply (test_octodns_provider_hetzner.TestHetznerProvider)
----------------------------------------------------------------------
  File "/Users/caretaker/Downloads/octodns/tests/test_octodns_provider_hetzner.py", line 107, in test_apply
    self.assertEquals(n, provider.apply(plan))
  File "/Users/caretaker/Downloads/octodns/octodns/provider/base.py", line 98, in apply
    self._apply(plan)
  File "/Users/caretaker/Downloads/octodns/octodns/provider/hetzner.py", line 380, in _apply
    getattr(self, '_apply_{}'.format(class_name))(zone_id, change)
  File "/Users/caretaker/Downloads/octodns/octodns/provider/hetzner.py", line 348, in _apply_Create
    self._client.zone_record_create(zone_id, params['name'],
  File "/Users/caretaker/Downloads/octodns/octodns/provider/hetzner.py", line 118, in zone_record_create
    record = self._do('POST', '/records', data=data)['record']
  File "/Users/caretaker/Downloads/octodns/octodns/provider/hetzner.py", line 60, in _do
    return self._do_raw(method, path, params, data).json()
  File "/Users/caretaker/Downloads/octodns/env/lib/python3.9/site-packages/mock/mock.py", line 1100, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/Users/caretaker/Downloads/octodns/env/lib/python3.9/site-packages/mock/mock.py", line 1104, in _mock_call
    return _mock_self._execute_mock_call(*args, **kwargs)
  File "/Users/caretaker/Downloads/octodns/env/lib/python3.9/site-packages/mock/mock.py", line 1163, in _execute_mock_call
    result = next(effect)
StopIteration: 
```

yzguy:
So the following code bit: 

```
        resp.json.side_effect = [
            DigitalOceanClientNotFound,  # no zone in populate
            DigitalOceanClientNotFound,  # no domain during apply
            domain_after_creation
        ]
```
Every time resp.json is called in your client, the response specified here is returned. In the case of the DigitaloceanProvider:

* `populate()` is called when a `provider.plan` is run, the `self.zone_records(zone)` is called, ending up a `self._client.records()` being called, which calls `.json()`. This is where the first `DigitalOceanClientNotFound` is returned

* `_apply()` is called when `provider.apply()` is called, ending up in `self._client.domain(domain_name)` being called, which calls `.json()`, and the mock returns the second `DigitalOceanClientNotFound`

* After the above `DigitalOceanClientNotFound` is returned and then caught by the try/except in `_apply`, it then calls `self._client.domain_create(domain_name)`, which ends up in it creating the domain, then immediately calling `self.records(name)`, which is where the last `.json()` is, with the mock returning `domain_after_creation`.

ricardbejarano:
Ooooh it's on `.json()` that it mocks!!

You're a legend!
Aight, back to work. Thanks!

ricardbejarano:
Alright, leaving coverage at 94% for today.
I'm missing thorough testing of record create/update/delete, but I now know how to do so.

ricardbejarano:
Alright, it's ready!

yzguy:
Seeing as this is a very close copy of DigitaloceanProvider, and this is passing all tests with 100% coverage. I'd say it's good to go. I'll let @ross do a quick look over, then merge.

ross (approved):
LGTM. Thanks for the contribution.

ricardbejarano:
Thanks again to both of you guys for the support!

🎉 🎉 🎉

ricardbejarano:
@ross how long until it's released? Do you have a timeline we can track?

ross:
> @ross how long until it's released? Do you have a timeline we can track?

No specific plans/timeline at the moment, but I can throw it on the TODO list. In the meantime you can install specific SHA's with something like https://github.com/octodns/octodns/#installing-a-specific-commit-sha

ricardbejarano:
👌👌

PR #693, opened 2021-04-21
Change DNSimple's Provider to query zones instead domains

duduribeiro:
A quick summary of a problem with the current DNSimple provider:
Let's suppose that I have the following config:

zones:
  30.100.194.in-addr.arpa.:
    sources:
      - config
    targets:
      - dnsimple

Even if a customer has this Reverse zone configured in DNSimple, this
fails with:

400 Bad Request for url: https://api.sandbox.dnsimple.com/v2/x/domains

because it is trying to create a domain because the zone wasn't found.
octodns.provider.dnsimple.DnsimpleClientNotFound: Not found

This happens because the GET /domains endpoint at DNSimple does not
bring Reverse Zones. To make this work nice, we should use /zones/
instead making it return properly the reverse zones.

### DIFF
--- a/octodns/provider/dnsimple.py
+++ b/octodns/provider/dnsimple.py
@@ -51,8 +51,8 @@ class DnsimpleClient(object):
         resp.raise_for_status()
         return resp
 
-    def domain(self, name):
-        path = '/domains/{}'.format(name)
+    def zone(self, name):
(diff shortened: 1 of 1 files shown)


ross (approved):
Makes sense to me... 

/cc https://developer.dnsimple.com/v2/zones/

Issue #694, opened 2021-04-22
Unused pools error for fallback-only pools


viranch:
I have a use-case to create a dynamic record with different fallback pools for each rule. Here's an example:

```yaml
example:
  ttl: 300
  type: A
  values: [4.4.4.4]
  dynamic:
    pools:
      one:
        fallback: two
        values:
          - value: 1.1.1.1
      two:
        values:
          - value: 2.2.2.2
      three:
        values:
          - value: 3.3.3.3
    rules:
      - geos:
        - NA-US
        pool: one
      - geos:
        - AS-SG
        pool: three
```

In other words, my default/fallback is different for each region. These fallbacks are not used as primary pool for any rule. This causes octoDNS to throw the 'unused pool' error. I'd like to skip this validation altogether. Having an unused pool doesn't seem all that practically harmful.

I looked at the code, and it seems pretty straight-forward to skip this validation. I can submit a PR but I wanted to open this issue to check what the authors think first.

viranch:
fixed by #695

PR #695, opened 2021-04-23
pools used as fallbacks should count as seen

ross:
If a pool is referenced as a fallback that should count as seen.

Note that after some quick testing with the NS1 provider I don't believe the fallback-only case is handled with it so this may not be enough to get the use case working depending on what provider(s) you're using. I haven't tested with Route53 yet...

/cc https://github.com/octodns/octodns/issues/694 & @viranch

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -531,6 +531,7 @@ class _DynamicMixin(object):
 
         pools_exist = set()
         pools_seen = set()
+        pools_seen_as_fallback = set()
         if not isinstance(pools, dict):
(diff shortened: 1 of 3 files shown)


ross:
Route53 does appear to work so there's a separate bug with NS1 not supporting the case.

ross:
> Note that after some quick testing with the NS1 provider I don't believe the fallback-only case is handled with it so this may not be enough to get the use case working depending on what provider(s) you're using.

Filed https://github.com/octodns/octodns/issues/696 to track

yeled (approved):

Issue #696, opened 2021-04-23
NS1 fails to sync fallback-only pools


ross:
NS1 has issues syncing when fallback-only pools are used. E.g. the following config:

```yaml
---
a:
  dynamic:
    pools:
      one:
        fallback: two
        values:
          - value: 1.1.1.1
      three:
        values:
          - value: 3.3.3.3
      two:
        values:
          - value: 2.2.2.2
    rules:
      - geos:
        - NA-US
        pool: one
      - geos:
        - AS-SG
        pool: three
  ttl: 300
  type: A
  values: [4.4.4.4]
```

The record appears to be created correctly:

<img width="856" alt="Screen Shot 2021-04-22 at 6 54 47 PM" src="https://user-images.githubusercontent.com/12789/115806769-3ab38000-a39c-11eb-857c-a4d7062ce9ce.png">

but when a sync is created the fallback pool isn't materalized and thus octoDNS thinks what's in NS1 needs updating:

```
2021-04-22T18:52:35  [4484468224] INFO  Ns1Provider[ns1] plan: desired=exxampled.com.
2021-04-22T18:52:35  [4484468224] WARNING Record Invalid record a.exxampled.com.
  - undefined fallback "two" for pool "one"
2021-04-22T18:52:35  [4484468224] INFO  Ns1Provider[ns1] populate:   found 2 records, exists=True
2021-04-22T18:52:36  [4484468224] INFO  Ns1Provider[ns1] plan:   Creates=0, Updates=1, Deletes=0, Existing Records=2
2021-04-22T18:52:36  [4484468224] INFO  Manager
********************************************************************************
* exxampled.com.
********************************************************************************
* ns1 (Ns1Provider)
*   Update
*     <ARecord A 300, a.exxampled.com., ['4.4.4.4'], {'one': {'fallback': 'two', 'values': [{'value': '1.1.1.1', 'weight': 1}]}, 'three': {'fallback': None, 'values': [{'value': '3.3.3.3', 'weight': 1}]}}, [{'pool': 'one', 'geos': ['NA-US']}, {'pool': 'three', 'geos': ['AS-SG']}]> ->
[...]

PR #698, opened 2021-04-24
Put pool name and fallback in answer notes

viranch:
Addresses #696

Removes fallback from region's notes and adds pool names and fallback in answer's notes. This enables the parser to correctly populate names and failovers of any failover-only pools. Tested working on NS1.


ross:
Created https://github.com/octodns/octodns/pull/699 half to work/think though things myself so that I fully grok the setup and other half to cleanly handle the upgrade/downgrade with setups that existed before these changes. The solution I landed on is pretty similar to what you have here, main difference is the upgrade/downgrade details.

viranch:
Closing in favor of #699

PR #699, opened 2021-04-26
Ns1 fallback only support

ross:
This PR reworks NS1 pool/value handling to support fallback-only pools in a backwards (and forwards) compatible way (including multiple fallback chains.) 

The toughest bit was getting it so that things could cleanly come to this version (without any changes when they didn't require it) and then cleanly go back (in case the upgrade/test wasn't perminent.  

The tests that are there currently pass, but some more tests need to be added to validate the new cases (thus coverage won't yet be complete.)

/cc https://github.com/octodns/octodns/pull/698 for a PR on the same issue @viranch
/cc Fixes https://github.com/octodns/octodns/issues/696

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -542,7 +542,7 @@ class Ns1Provider(BaseProvider):
                 rules[rule_order] = rule
 
             # The group notes field in the UI is a `note` on the region here,
-            # that's where we can find our pool's fallback.
+            # that's where we can find our pool's fallback in < v0.9.11 anyway
             if 'fallback' in notes:
                 # set the fallback pool name
                 pools[pool_name]['fallback'] = notes['fallback']
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -406,7 +406,7 @@ class Ns1Provider(BaseProvider):
             for piece in note.split(' '):
                 try:
                     k, v = piece.split(':', 1)
-                    data[k] = v
+                    data[k] = v if v != '' else None
                 except ValueError:
                     pass
         return data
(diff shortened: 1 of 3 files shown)


viranch:
Shouldn't `pool-only-in-fallback` from `tests/config/dynamic.tests.yaml` added in #695 test for the new test case?

ross:
> Shouldn't `pool-only-in-fallback` from `tests/config/dynamic.tests.yaml` added in #695 test for the new test case?

Yeah that would cover one of the cases that needs testing here. The other stuff will have to do with the upgrade/downgrade cases and will be harder to hit/test. May even require some refactoring... will have to see when I get a chance to sit down and work on it, hopefully this afternoon or tomorrow.

ross:
Testing/coverage is now complete and there's some bits hitting the new code paths during unit tests.

I've also run the case in https://github.com/octodns/octodns/issues/696 as well as some "real world" examples that I have that are close to complicated stuff I've used in the past. I've also round-tripped things from current `master`, this branch, and back to verify that things seem to work as expected/stay sane. 

/cc @pavantc @theojulienne @yeled do you all want to do any testing with this and/or review? Some of my test cases are based on things I created to act like production records so I'm fairly sure it'll work and worse case once you all upgrade it'll tell you if it's going to change a bunch of stuff... :grin:

theojulienne (approved):

viranch:
can this be merged?

ross:
> can this be merged?

Totally thought it was 😦

viranch:
looks like this needs attention

ross:
> looks like this needs attention

Yeah. I started to look into the conflicts locally and they weren't simple and I haven't gotten back to them yet.

ross:
Think that got it.

viranch:
💥 thank you!

PR #700, opened 2021-04-26
fixed missing whitespace in plan() debug logging

ricardbejarano:
When debug logging is enabled, the plan() function logs lines like:

```
<previous output>
Plan: DEBUG: __init__: Creates=13, Updates=0, Deletes=0Existing=0
<following output>
```

Where a space between "Deletes=0" and "Existing=0" is missing.
This adds it.


ross (approved):
Thanks for catching & fixing.

PR #704, opened 2021-04-30
Add support for dynamic CNAME records in NS1

viranch:


viranch:
Monitor creation is failing due to trailing dots in CNAME answers, will work on it later

ross:
Excited to see this getting some attention. It's a frequently asked for thing, but not something I ever had the use for so didn't get to put time into. Will keep an eye on this for when it's ready to review and might take a peek at Route53 support as well.

viranch:
Alright I think it's ready for review now @ross. I tested on my side with NS1 and seems to work as expected.

octodns/provider/ns1.py
```diff
@@ -646,6 +652,10 @@ def _data_for_CAA(self, _type, record):
         }
 
     def _data_for_CNAME(self, _type, record):
+        if record.get('tier', 1) > 1:
+            # Advanced dynamic record
+            return self._data_for_dynamic_A(_type, record)
```
ross, line 657: We should probably go ahead and rename this method now that it supports A, AAAA,  and CNAME. `_data_for_dynamic` is probably sufficient. (It already did A and AAAA so the name wasn't great...)

ross (commented):
Overall looks good. One minor rename/clarification suggestion inline and then I think this is 👍

octodns/provider/ns1.py
viranch: Went ahead and renamed both _data_for_dynamic_A and _params_for_dynamic_A methods.

ross (approved):

PR #705, opened 2021-05-11
Update the AzureProvider to support azure-mgmt-dns 8.0.0 and azure-identity.

omarkilani:
Looks like MS broke the API again... 🤷‍♂️


omarkilani:
There's a very weird test failure that I'm still looking into, which I think is a bug in the test itself.

ross (approved):
> Looks like MS broke the API again... 🤷‍♂️

Good grief ☹️

PR #706, opened 2021-05-11
Support dynamic records in Azure DNS

viranch:
Adds support for dynamic records in Azure DNS by using Azure Traffic Managers (ATMs).

An ATM can be one of 'Geographic', 'Priority' or 'Weighted' (among others). So a single ATM can only do one of geo-fencing, failover and weighted round-robin. All types support TCP, HTTP and HTTPS health-checks with customizable host, port and path.

To achieve the full functionality of dynamic records, multiple ATMs need to be created (one for each pool, each rule and failover chains) and nested together. This PR implements dynamic records by creating 3 layers of nested ATMs: DNS record -> Geographic -> Priority -> Weighted. Each pool gets a Weighted ATM, which are nested within another Priority ATM for failovers corresponding to each rule, which are further nested inside a top-level Geographic ATM which does geo-fencing. The top-level ATM then gets aliased to a record inside the DNS zone (this is not an ALIAS record, it's a Azure-specific "nesting" which is called 'alias' on the portal).

Here's my attempt at a graphical representation:
```yaml
www:
  ttl: 60
  type: CNAME
  value: default.com.
  dynamic:
    pools:
      one:
        fallback: two
        values:
        - value: one1.com.
          weight: 11
        - value: one2.com.
          weight: 12
      two:
        values:
        - value: two.com.
    rules:
    - geos:
      - EU
      pool: one
    - pool: two
```
becomes:
```
www.site.com
  -> www-site-com (Geographical ATM)
     Europe -> rule-one--www-site-com (Priority ATM)
               Priority 1 -> pool-one--www-site-com (Weighted ATM)
                             Weight 11 -> one1.com
                             Weight 12 -> one2.com
               Priority 2 -> two.com
               Priority 3 -> default.com
     WORLD  -> rule-two--www-site-com (Priority ATM)
               Priority 1 -> two.com
               Priority 2 -> default.com
```
Each ATM gets the heathcheck config as defined in octodns. This works fairly well in my testing.

Now some caveats:
* Nested ATMs actually result in `foo.trafficmanager.net` CNAMEs to send DNS clients to each other. This introduces extra CNAME hops in DNS resolution of user's dynamic records, and poses all risks associated with single-DNS due to `*.trafficmanager.net` CNAMEs. The above example will result in this CNAME chain for Europe:
```
$ dig +short +noshort www.site.com
www.site.com.                               60  IN  CNAME   www-site-com.trafficmanager.net.
www-site-com.trafficmanager.net.            60  IN  CNAME   rule-one--www-site-com.trafficmanager.net.
rule-one--www-site-com.trafficmanager.net.  60  IN  CNAME   pool-one--www-site-com.trafficmanager.net.
pool-one--www-site-com.trafficmanager.net.  60  IN  CNAME   one1.com.
```
* Because of above, it is impossible to alias A/AAAA records to ATMs that have other nested ATMs. As a result, Azure DNS cannot fully support A/AAAA dynamic records.
* ATMs cannot have health checking disabled on any of its member endpoints. As a result, the default answer will also be health-checked. If all answers in the pools are unhealthy along with the default one, Azure DNS may not hand out the default answer.

I believe this is still somewhat functional if someone wanted to pursue dynamic records on Azure DNS.


viranch:
alright done with force pushes (sorry!) and more commits, this is ready for review.

ross:
I originally looked at Azure dynamic support and bailed b/c they didn't support transparently having `www.foo.com.` backed by a Traffic Manager setup when the TM used geo rules. Looked like you skirted that by using a CNAME and making the trafficmanager.net. hop visible?  I wasn't super keen on that setup since it was SO different from how everthing else would work and I didn't want to mix NS1/Dyn/Route53 records where it'd be direct and Azure where there'd be an extra hop/CNAME. 

That was more of a specific product decision at the time (didn't want it visible for github.com etc.) and Azure lost out in the selection as a result. I really wish they'd address that, but I'm not opposed having support that does it (I just wouldn't use it for "big" things.) :grin:

> alright done with force pushes (sorry!) and more commits, this is ready for review.

Cool. Will aim to dig into this today.

viranch:
> using a CNAME and making the trafficmanager.net. hop visible

I am not CNAMEing to the traffic manager, I am using 'Nested' endpoint to select the target traffic manager from a dropdown. This makes the CNAME hop visible, no way around it. Azure might will rollout a feature to hide that CNAME hop at some point.

> mix NS1/Dyn/Route53 records where it'd be direct and Azure where there'd be an extra hop/CNAME

Agreed, it's not a uniform behavior across providers and you'd be better off without the extra CNAME hop.

ross:
> > using a CNAME and making the trafficmanager.net. hop visible
> 
> I am not CNAMEing to the traffic manager, I am using 'Nested' endpoint to select the target traffic manager from a dropdown. This makes the CNAME hop visible, no way around it. Azure might rollout a feature to hide that CNAME hop at some point.

Oh that would have been a blocker if I'd known/realized it back when I evaled things. At the time I couldn't point it at a Geo TM so I stopped there. Amazingly I could point it at something else that was supported and then go in and change the TM to be a Geo and it seemed to work, but I wasn't actually going to try hijinks like that with a real/production setup.

README.md
```diff
@@ -192,7 +192,7 @@ The above command pulled the existing data out of Route53 and placed the results
 
 | Provider | Requirements | Record Support | Dynamic | Notes |
 |--|--|--|--|--|
-| [AzureProvider](/octodns/provider/azuredns.py) | azure-identity, azure-mgmt-dns | A, AAAA, CAA, CNAME, MX, NS, PTR, SRV, TXT | No | |
+| [AzureProvider](/octodns/provider/azuredns.py) | azure-identity, azure-mgmt-dns, azure-mgmt-trafficmanager | A, AAAA, CAA, CNAME, MX, NS, PTR, SRV, TXT | Yes (CNAMEs only) | |
```
ross, line 195: What does the provider do when a dynamic `A` or `AAAA` is defined? Given that dynamic is a yes/no and not a per type thing it seems like it'd probably just repeatedly show a change that can't be applied.

We may need to extend that to optionally support dynamic types if we want to ship this to prevent that.

Would it maybe make sense to do something like the following to call out that it's a little "non-standard" in terms of the intermediate CNAME? 

```suggestion
| [AzureProvider](/octodns/provider/azuredns.py) | azure-identity, azure-mgmt-dns, azure-mgmt-trafficmanager | A, AAAA, CAA, CNAME, MX, NS, PTR, SRV, TXT | Alpha (CNAMEs only) | |
```

We probably need to find a place to put the caveats you mention in https://github.com/octodns/octodns/pull/706#issue-642259778 so they're easily found if people go to look at using it.  There isn't a great place for that sort of thing ATM, so maybe just comments in the provider's docstring. As we break each provider out into it's own repo we'll have better options for that sort of thing.

octodns/provider/azuredns.py
```diff
@@ -94,21 +105,23 @@ def __init__(self, resource_group, record, delete=False):
...
         key_name = '{}_records'.format(self.record_type).lower()
         if record._type == 'CNAME':
-            key_name = key_name[:len(key_name) - 1]
+            key_name = key_name[:-1]
         azure_class = self.TYPE_MAP[self.record_type]
 
-        self.params = getattr(self, '_params_for_{}'.format(record._type))
-        self.params = self.params(record.data, key_name, azure_class)
+        params_for = getattr(self, '_params_for_{}'.format(record._type))
+        self.params = params_for(record.data, key_name, azure_class)
```
ross, line 124: Thanks for cleaning these bits up. Not sure how they were missed in the original PR review.

octodns/provider/azuredns.py
```diff
@@ -336,24 +407,45 @@ def __init__(self, id, client_id, key, directory_id, sub_id,
...
-                client_id=self._dns_client_client_id,
-                client_secret=self._dns_client_key,
-                tenant_id=self._dns_client_directory_id
-            )
+            # Azure's logger spits out a lot of debug messages at 'INFO'
+            # level, override it by re-assigning `info` method to `debug`
+            # (ugly hack until I find a better way)
+            logger_name = 'azure.core.pipeline.policies.http_logging_policy'
+            logger = logging.getLogger(logger_name)
+            logger.info = logger.debug
```
ross, line 424: 😦  👍

ross (commented):
Made a first pass through and added a few comments/questions. Will look more deeply :soon:

README.md
viranch: > it'd probably just repeatedly show a change that can't be applied

The provider throws an exception in `_extra_changes` if there's a non-CNAME dynamic record detected. This way we tell the user early on during the plan stage instead of giving up only later during the apply stage.

> extend that to optionally support dynamic types

not sure what this means or is required given above.

Agreed with your suggestion and about mentioning caveats in the doc-string. Commit coming up!

octodns/provider/azuredns.py
viranch: The caveats text is too long for the doc-string, I've instead linked to this PR suggesting the reader to read it over.

README.md
viranch: The caveats text is too long for the doc-string, I've instead linked to this PR suggesting the reader to read it over.

octodns/record/__init__.py
```diff
@@ -429,6 +429,10 @@ def __init__(self, _id, data):
         ]
         values.sort(key=lambda d: d['value'])
 
+        # normalize weight of a single-value pool
+        if len(values) == 1:
+            values[0]['weight'] = 1
```
ross, line 434: Having a hard time figuring out the path that would lead to this being required. https://github.com/octodns/octodns/pull/706/files#diff-7f4d8ad4d1143b84eb1f75b328f01cf688b9421abc8d0295077ac94922814020R427 in the list comprehension for `data['values']` above seems like it should set the weight to `1` when it's not define. Is this trying to "fix" when people put a non-`1` weight on a single value? If so 👍, but we should probably stick a `logger.warn` here telling them that it's happening so they'll have a chance to learn their config doesn't make sense.

octodns/provider/azuredns.py
```diff
@@ -495,6 +624,324 @@ def _data_for_TXT(self, azrecord):
...
+                    # Throw exception otherwise, it should not happen if the
+                    # profile was generated by octoDNS
+                    if 'GEO-AS' not in geo_map:
+                        msg = '_data_for_dynamic: Profile={}: '.format(
+                            geo_profile.name)
+                        msg += 'Middle East (GEO-ME) is not supported by ' + \
+                               'octoDNS. It needs to be either paired ' + \
+                               'with Asia (GEO-AS) or expanded into ' + \
+                               'individual list of countries.'
+                        raise AzureException(msg)
```
ross, line 653: Fwiw we transparently mapped `OC` (Oceania) into it's list of countries and back in the NS1 provider, 

https://github.com/octodns/octodns/blob/8be24f10a104c1f238bade6fa71554336bef776e/octodns/provider/ns1.py#L358-L362

It was a PITA and took a lot of care to get all the details right, but just wanted to call that out in case it's of interest to go that route here.

ross (commented):
I think things are good here if it's working for you and you're happy with it. I don't really have an Azure setup to test this with currently so sort of going to take your word for it :grin:

Couple more questions inline I'll let you review/ack. Once that's done we can :shipit:

octodns/record/__init__.py
viranch: > "fix" when people put a non-1 weight on a single value?

yep exactly. will add the `logger.warn`.

octodns/provider/azuredns.py
viranch: Yep I saw that, I found it cleaner to use Asia instead of checking for a static list of countries. Plus we won't run into this unless someone has fiddled with octoDNS-generated Traffic Managers. `octodns-dump` will anyway not work due to this provider expecting certain naming conventions.

octodns/record/__init__.py
viranch: Added the warning.

ross: Sorry it didn't occur to me when I was looking at this and just popped into my head this afternoon when I was AFK. I believe this weight check should actually be a validation rather than an auto-fix & warning as many things of this sort are already pointed out at that phase and people can always configure `lenient` if they have a strong reason to preserve something that doesn't agree.

I've PR'd that over in https://github.com/octodns/octodns/pull/715 and will pull this and the test change below out of this one in favor of that, and to avoid requiring another round trip of suggestion, change, review.

PR #714, opened 2021-05-17
string check support for NS1 monitor

meghashyamps:
Adding support to pass the "string to check" if present. This is backward compatible to checking for the default "200 OK"

```
---
  octodns:
    healthcheck:
      host: my-host-name
      path: /dns-health-check
      port: 443
      protocol: HTTPS
    ns1:
      healthcheck:
        value: "check if this string is present"
```


ross (commented):
Would prefer the key not be provider specific, e.g:

```yaml
octodns:
  healthcheck:
    host: my-host-name
    path: /dns-health-check
    port: 443
    protocol: HTTPS
    response:
      contains: "check if this string is present"
      # this would be a default, that could be overridden
      # status: 200
```

That should allow other providers that support it to implement the concept cleanly.

The status code check should probably be left in place and the new contains string can be an additional check. If you specifically need to get rid of the status check we could add code to leave it out when `status: None` is provided under `response`.

meghashyamps:
I have implemented it like so:

```
  octodns:
    healthcheck:
      path: /pop/admin
      port: 443
      protocol: HTTPS
      response:
        contains: "check if this string is present"
```
Adding multiple `Up` conditions for NS1 TCP monitor was not supported from what I could tell from the schema. The current implementation of HTTP using TCP was anyway using string check to look for `200 OK` in the response. This PR just tries to extend that by providing a way to check a string other than `200 OK`. If the response part is not defined, it would just default to using `200 OK`.

octodns/record/__init__.py
```diff
@@ -217,6 +217,16 @@ def healthcheck_port(self):
...
 
+    @property
+    def healthcheck_response(self):
+        healthcheck = self._octodns.get('healthcheck', {})
+        if healthcheck.get('protocol', None) == 'TCP':
+            return None
+        try:
+            return healthcheck['response']
+        except KeyError:
+            return {'contains': '200 OK'}
```
ross, line 228: This is specific to NS1 so doesn't belong in the non-provider specific/generic Record object

ross (commented):
> Adding multiple Up conditions for NS1 TCP monitor was not supported from what I could tell from the schema.

Its definitely possible to have multiple contains:

Screen Shot 2021-05-18 at 6 31 42 AM

meghashyamps:
Alright, taking another shot at this. Since the current implementation is hardcoded to check for `200 OK` as a `string match` (not status code check as in a HTTP response) for TCP monitor, it was starting to get hairy trying to add stuff without breaking existing defaults. But I have done my best.

octodns/record/__init__.py
```diff
@@ -217,6 +217,27 @@ def healthcheck_port(self):
...
+    # string to check in response body
+    @property
+    def healthcheck_responsebody(self):
+        healthcheck = self._octodns.get('healthcheck', {})
+        if healthcheck.get('protocol', None) == 'TCP':
+            return None
+        try:
+            return healthcheck['response_body']
+        except KeyError:
+            return None
```
ross, line 239: Body implies that it's matching the full response body, but this is just searching for a string in the response (body) and thus `contains` is more appropriate. 


```suggestion
    # string to check for in response body
    @property
    def healthcheck_response_contains(self):
        healthcheck = self._octodns.get('healthcheck', {})
        if healthcheck.get('protocol', None) == 'TCP':
            return None
        try:
            return healthcheck['response_contains']
        except KeyError:
            return None
```

octodns/provider/ns1.py
```diff
@@ -933,12 +933,19 @@ def _monitor_gen(self, record, value):
...
+                    'comparison': 'contains',
+                    'key': 'output',
+                    'value': '200 OK'
+                })
+            if record.healthcheck_responsebody is not None:
+                rules.append({
+                    'comparison': 'contains',
+                    'key': 'output',
+                    'value': record.healthcheck_responsebody,
+                })
```
ross, line 948:
```suggestion
            if record.healthcheck_response_contains is not None:
                rules.append({
                    'comparison': 'contains',
                    'key': 'output',
                    'value': record.healthcheck_response_contains,
                })
```

octodns/provider/ns1.py
```diff
@@ -933,12 +933,19 @@ def _monitor_gen(self, record, value):
...
                 r'User-agent: NS1\r\n\r\n'.format(path=path, host=host)
             ret['config']['send'] = request
-            # We'll also expect a HTTP response
-            ret['rules'] = [{
-                'comparison': 'contains',
-                'key': 'output',
-                'value': '200 OK',
-            }]
+            rules = ret.setdefault('rules', [])
+            if record.healthcheck_statuscode is not None:
```
ross, line 937: This is now allowing the config to contain an arbitrary status code, and if anything is defined it's adding `200 OK` to the list of requirements. If someone were to set `201` this wouldn't work as expected.

Relatedly since this capability is being added I took a look at how it would be implemented with Route53 and this bit is interesting:

> HTTP : Route 53 tries to establish a TCP connection. If successful, Route 53 submits an HTTP request and waits for an HTTP status code of 200 or greater and less than 400.

So it doesn't allow things to specify a status code as-is, but there is a way to include a single contains string via a `SearchString` parameter to the healthcheck config.

Since Route53 doesn't support specifying a status code and this implementation and this implementation doesn't match the schema you've defined I'd recommend either:

1. Dropping the ability to configure the status code and just leaving NS1 as-is with a `200 OK` always there and the response contain if that's configured. 
2. Implementing a lookup table in NS1 to map the status code to the expected string, e.g. `200: '200 OK'`,  `201: '201 Created'`, ...

I'd leant towards 1 based on Route53's lack of configurability and the fact that I'm not aware of anyone requesting/needing the ability to change the response status code.

meghashyamps: > 1. Dropping the ability to configure the status code and just leaving NS1 as-is with a `200 OK` always there and the response contain if that's configured.

This is doable except it doesn't solve for our use case where certain legacy endpoints monitored are network/firewall devices which do not really return a "200 OK" in the response. Instead it just returns a login form containing string such as `Welcome to Cisco` or `please enter administrator username` etc. In such cases multiple `UP` conditions having to match `200 OK` as well as `Welcome to` will not pass. 

I am going to look at those legacy devices to see if anything else can be monitored on them. But as it stands, do you think there is any other way to achieve this?

ross: OK. In that case there would probably need to be a translation between status code and the expected text, though it doesn't really line up with what Route53 does/supports. 

We could maybe include the `200 OK` if no `healthcheck_response_contains` is defined and if it is only do the contains, skipping the 200. That wouldn't line up exactly with Route53, which would still require `200<=status<400`, but that's a static requirement for it and it would at least would be consistent with what's there currently.

meghashyamps: @ross  I was able to make things work with existing implementation of checking for "200 OK". We can close this PR as "not needed" at this time.

Issue #594, opened 2020-08-06
Route53: dynamic record triggers update even with no changes to the record


ttk:
I have the following dynamic record:
```
www:
  dynamic:
    pools:
      eu:
        values:
        - value: aloadofELBstuffhere.amazonaws.com.
    rules:
    - geos:
      - EU
      pool: eu
  ttl: 300
  type: CNAME
  value: web.example.com.
```

When I sync to route53, it creates the record(s) the first time. All good. When I run a sync the second time, it reports an update, even though nothing has changed:

```
* route53 (Route53Provider)
*   Update
*     <CnameRecord CNAME 300, www.testing123.com., web.example.com., {'eu': {'fallback': None, 'values': [{'value': 'aloadofELBstuffhere.amazonaws.com.', 'weight': 1}]}}, [{'pool': 'eu', 'geos': ['EU']}]> ->
*     <CnameRecord CNAME 300, www.testing123.com., web.example.com., {'eu': {'fallback': None, 'values': [{'value': 'aloadofELBstuffhere.amazonaws.com.', 'weight': 1}]}}, [{'pool': 'eu', 'geos': ['EU']}]> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=3
```

I expect the sync to detect that nothing has changed and Updates=0.

warmfusion:
Looks like the issue is with regards to the healthcheck elements;

Longer log output 
```
2020-08-05T20:11:50  [139819830553928] INFO  Route53Provider[route53] plan: desired=example.com.
2020-08-05T20:11:51  [139819830553928] INFO  Route53Provider[route53] populate:   found 53 records, exists=True
2020-08-05T20:11:51  [139819830553928] INFO  Route53Provider[route53] _extra_changes_update_needed: health-check caused update of www.example.com.:CNAME
2020-08-05T20:11:51  [139819830553928] INFO  Route53Provider[route53] _extra_changes_dynamic_needs_update: health-check caused update of www.example.com.:CNAME
2020-08-05T20:11:51  [139819830553928] INFO  Route53Provider[route53] plan:   extra changes
  Update
[...]

PR #716, opened 2021-05-18
Do not trigger change for health checks on cname dynamic records

ttk:
Fixes issue #594 

For dynamic CNAME records, route53 will by default assign the health check host to be the value of the CNAME for that pool.  In octodns, the default value of the health check host is the fqdn for the record.  These two defaults don't match up and therefore octodns detects a change.  When the apply runs, `get_health_check_id()` detects that the health check host is not an IP address, and basically reverts back to the original health check host and effectively no change is done.

As a quick fix, I replicated the logic in `get_health_check_id()` in `_extra_changes_update_needed()`.  This effectively makes the change detection logic match what apply would do.  I think a proper fix would involve configurable health checks per dynamic pool, but this would involve a lot more work. 

Added unit test, 100% code coverage.

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -1253,15 +1253,11 @@ class Route53Provider(BaseProvider):
         return self._gen_mods('DELETE', existing_records, existing_rrsets)
 
     def _extra_changes_update_needed(self, record, rrset):
-        value = rrset['ResourceRecords'][0]['Value']
-
-        try:
-            ip_address(text_type(value))
-            # We're working with an IP
+        if record._type == 'CNAME':
+            # For CNAME, healthcheck host by default points to the CNAME value
+            healthcheck_host = rrset['ResourceRecords'][0]['Value']
+        else:
             healthcheck_host = record.healthcheck_host
-        except (AddressValueError, ValueError):
-            # This isn't an IP, host is the value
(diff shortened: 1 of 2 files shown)


octodns/provider/route53.py
```diff
@@ -1253,7 +1253,16 @@ def _mod_Delete(self, change, zone_id, existing_rrsets):
...
-        healthcheck_host = record.healthcheck_host
+        value = rrset['ResourceRecords'][0]['Value']
+
+        try:
+            ip_address(text_type(value))
+            # We're working with an IP
+            healthcheck_host = record.healthcheck_host
+        except (AddressValueError, ValueError):
+            # This isn't an IP, host is the value
+            healthcheck_host = value
```
ross, line 1264: Would it be more direct to check `record._type` here and return `value` if it's `CNAME` and `record.healthcheck_host` otherwise? I believe this would work, but it took me staring at it and thinking about the cases for a couple minutes before I thought so. The if type route might be a little more explicit and immediately clear (unless I'm missing something.)

ttk: Yes, that should work too and I agree it would be more explicit.  I'll prepare an update to this PR with the changes.  Thanks for reviewing.

ross (approved):

PR #717, opened 2021-05-18
map Oceania to Australia/Pacific in Azure

viranch:
per https://docs.microsoft.com/en-us/azure/traffic-manager/traffic-manager-geographic-regions


ross (approved):
:cool:

PR #722, opened 2021-05-28
Adding URLFWD type to Record handler and NS1, YAML, CloudFlare Providers

brianeclow:
# Goal

Creating initial `URLFWD` support and testing, targeting NS1 and YAML providers.

## Implementation

- Adding `URLFWD` record type and basic checking to Record class
- Adding `URLFWD` to supported types and handler to NS1 Provider
- Adding `URLFWD` to supported types to YAML Provider
- Adding `URLFWD` to supported types and handlers to CloudFlare Provider based on CloudFlare's PageRules `forwarding_url` functionality
- Testing 💯 

## Example

```yaml
urlfwd:
  ttl: 300
  type: URLFWD
  values:
  - code: 302
    masking: 2
    path: '/'
    query: 0
    target: 'http://www.example.com'
```

## Testing

- [x] Full test coverage
- [x] Tested with NS1 provider
- [x] Tested YAML and YAML Split Providers
- [x] Tested with CloudFlare provider

### DIFF
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -299,7 +299,7 @@ class CloudflareProvider(BaseProvider):
             })
         return {
             'type': _type,
-            'ttl': 3600,  # ttl does not exist for this type, forcing a setting
+            'ttl': 300,  # ttl does not exist for this type, forcing a setting
             'values': values
         }
 
@@ -376,10 +376,10 @@ class CloudflareProvider(BaseProvider):
                     path = parsed_uri.path
                     _type = 'URLFWD'
                     # assumption, actions will always contain 1 action
-                    values = record['actions'][0]['value']
-                    values['path'] = path
+                    _values = record['actions'][0]['value']
+                    _values['path'] = path
                     # no ttl set by pagerule, creating one
-                    values['ttl'] = 3600
+                    _values['ttl'] = 300
                     values[name][_type].append(_values)
                 # the dns_records branch
                 # elif 'name' in record:
@@ -618,9 +618,10 @@ class CloudflareProvider(BaseProvider):
             uri = data['targets'][0]['constraint']['value']
             uri = '//' + uri if not uri.startswith('http') else uri
             parsed_uri = urlsplit(uri)
-            return '{name} {path} {url} {status_code}'.format(name=parsed_uri.netloc, 
-                                                              path=parsed_uri.path, 
-                                                              **data['actions'][0]['value'])
+            return '{name} {path} {url} {status_code}' \
+                .format(name=parsed_uri.netloc,
+                        path=parsed_uri.path,
+                        **data['actions'][0]['value'])
         return data['content']
 
     def _apply_Create(self, change):
@@ -649,11 +650,11 @@ class CloudflareProvider(BaseProvider):
                 name = zone.hostname_from_fqdn(parsed_uri.netloc)
                 path = parsed_uri.path
                 # assumption, actions will always contain 1 action
-                values = record['actions'][0]['value']
-                values['path'] = path
-                values['ttl'] = 3600
-                values['type'] = 'URLFWD'
-                record.update(values)
+                _values = record['actions'][0]['value']
+                _values['path'] = path
+                _values['ttl'] = 300
+                _values['type'] = 'URLFWD'
+                record.update(_values)
             else:
                 name = zone.hostname_from_fqdn(record['name'])
             # Use the _record_for so that we include all of standard
--- a/octodns/provider/cloudflare.py
+++ b/octodns/provider/cloudflare.py
@@ -373,13 +373,13 @@ class CloudflareProvider(BaseProvider):
                     uri = '//' + uri if not uri.startswith('http') else uri
                     parsed_uri = urlsplit(uri)
                     name = zone.hostname_from_fqdn(parsed_uri.netloc)
-                    _path = parsed_uri.path
+                    path = parsed_uri.path
                     _type = 'URLFWD'
                     # assumption, actions will always contain 1 action
-                    _values = record['actions'][0]['value']
-                    _values['path'] = _path
+                    values = record['actions'][0]['value']
+                    values['path'] = path
                     # no ttl set by pagerule, creating one
-                    _values['ttl'] = 3600
+                    values['ttl'] = 3600
                     values[name][_type].append(_values)
                 # the dns_records branch
                 # elif 'name' in record:
@@ -589,10 +589,7 @@ class CloudflareProvider(BaseProvider):
         # content as things are currently implemented so we need to handle
         # those explicitly and create unique/hashable strings for them.
         # AND... for URLFWD/Redirects additional adventures are created.
-        if 'targets' in data:
-            _type = 'URLFWD'
-        else:
-            _type = data['type']
+        _type = data.get('type', 'URLFWD')
         if _type == 'MX':
             return '{priority} {content}'.format(**data)
         elif _type == 'CAA':
@@ -621,15 +618,9 @@ class CloudflareProvider(BaseProvider):
             uri = data['targets'][0]['constraint']['value']
             uri = '//' + uri if not uri.startswith('http') else uri
             parsed_uri = urlsplit(uri)
-            ret = {}
-            ret.update(data['actions'][0]['value'])
-            ret.update({'name': parsed_uri.netloc, 'path': parsed_uri.path})
-            urlfwd = (
-                '{name}',
-                '{path}',
-                '{url}',
-                '{status_code}')
-            return ' '.join(urlfwd).format(**ret)
+            return '{name} {path} {url} {status_code}'.format(name=parsed_uri.netloc, 
+                                                              path=parsed_uri.path, 
+                                                              **data['actions'][0]['value'])
         return data['content']
 
     def _apply_Create(self, change):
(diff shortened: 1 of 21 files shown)


ross (commented):
Looks reasonable. I don't have URLFWD support in my test account so i've asked that it be enabled so I can take a closer look as some of the details/options.

In general with octoDNS it's best to only add new types with support for at least 2 providers. The main reasons for this is that it prevents the implementation from being too centered around the first/only provider. E.g. NS1's URLFWD may do things that no one else does and/or just have vastly different semantics. If things are initially implemented with only it in mind they may be done in a way that won't be compatible with others... You know of/have access to any other providers with URL redirection support?

octodns/record/__init__.py
```diff
@@ -1471,3 +1472,88 @@ class _TxtValue(_ChunkedValue):
 class TxtRecord(_ChunkedValuesMixin, Record):
     _type = 'TXT'
     _value_type = _TxtValue
+
+
+class UrlfwdValue(EqualityTupleMixin):
+    # TODO: should have defaults for path, code, masking, and query
```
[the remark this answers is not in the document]
brianeclow: I was holding off on the defaults until another provider could be tested. As you pointed out, octoDNS _should_ support more than just one provider's opinionated ways. I did make several assumptions centered on how NS1 handled `URLFWD`s/Redirections with some of my remembered experiences with Dyn. Unfortunately, I do not have access to a Dyn setup to test against.

brianeclow:
@actazen, Thank you.

octodns/record/__init__.py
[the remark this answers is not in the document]
ross: Sorry missed this coming through a while back and only saw the commit earlier today. I no longer have access to Dyn either, but I'm considering it deprecated/fine to ignore since it's slated for shutdown so it should be fine to ignore it for the purpose of octoDNS. It looks like DNSimple supports a form of redirection if that's useful https://support.dnsimple.com/articles/redirector/.

Lmk when you're ready for review.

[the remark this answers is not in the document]
brianeclow: Due to a POC at work, I have access to CloudFlare, and I am in process of building that addition, now.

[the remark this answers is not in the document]
brianeclow: Decided to move away from the type checking, as from my research and dealings with CloudFlare, this would become complicated fast.

brianeclow:
@ross, I have added and tested out CloudFlare as the second provider. Ready for the red pin on my first real python project update. 😅

octodns/provider/cloudflare.py
```diff
@@ -170,21 +171,24 @@ def zones(self):
 
         return self._zones
 
+    def _ttl_data(self, ttl):
+        return 300 if ttl == 1 else ttl
```
ross, line 175: Where does the default of 300 come from? I noticed [down below](https://github.com/octodns/octodns/pull/722/files#diff-7ca0e511d38100edff85700683a6bb4dd6b5f4127fe830f43e2348ffac2f2d29R302) there's a default for urlfwd's of 3600. Should the two values agree? Some providers have a `default_ttl` configuration parameter, e.g. [`YamlProvider`](https://github.com/octodns/octodns/blob/master/octodns/provider/yaml.py#L27-L28), maybe we should go ahead and do that for Cloudflare and use that value here and down below for URLFWD.

octodns/provider/cloudflare.py
```diff
@@ -279,7 +283,23 @@ def _data_for_SRV(self, _type, records):
...
+            values.append({
+                'path': r['path'],
+                'target': r['url'],
+                'code': r['status_code'],
+                'masking': 2,
+                'query': 0,
+            })
+        return {
+            'type': _type,
+            'ttl': 3600,  # ttl does not exist for this type, forcing a setting
```
ross, line 302: If someone puts a TTL in the (yaml) config for a `URLFWD` type record that doesn't match this value it'll probably show up as a "constant" change that can't be resolved. There's some other [special cases of changes around TTL down in the `_include_change` method](https://github.com/octodns/octodns/blob/5e2f7d43845d48fbf2e92b7bbaf2bb5b0e529b33/octodns/provider/cloudflare.py#L411) that may need expanding if so.

Long shot, but is there any sort of "notes" field on the URLFWD record where we could stuff the TTL and get it back out of the api? Might save some PITA if so.

octodns/provider/cloudflare.py
```diff
@@ -338,10 +365,29 @@ def populate(self, zone, target=False, lenient=False):
...
+                    uri = '//' + uri if not uri.startswith('http') else uri
+                    parsed_uri = urlsplit(uri)
+                    name = zone.hostname_from_fqdn(parsed_uri.netloc)
+                    _path = parsed_uri.path
+                    _type = 'URLFWD'
+                    # assumption, actions will always contain 1 action
+                    _values = record['actions'][0]['value']
+                    _values['path'] = _path
+                    # no ttl set by pagerule, creating one
+                    _values['ttl'] = 3600
```
ross, line 382: I don't think most of these need the `_` on the front. It's just used for `type` since that's a python function.
```suggestion
                    path = parsed_uri.path
                    _type = 'URLFWD'
                    # assumption, actions will always contain 1 action
                    values = record['actions'][0]['value']
                    values['path'] = path
                    # no ttl set by pagerule, creating one
                    values['ttl'] = 3600
```

octodns/provider/cloudflare.py
```diff
@@ -512,7 +588,11 @@ def _gen_key(self, data):
         # BUT... there are exceptions. MX, CAA, LOC and SRV don't have a simple
         # content as things are currently implemented so we need to handle
         # those explicitly and create unique/hashable strings for them.
-        _type = data['type']
+        # AND... for URLFWD/Redirects additional adventures are created.
+        if 'targets' in data:
+            _type = 'URLFWD'
+        else:
+            _type = data['type']
```
ross, line 595: If `"type"` doesn't exist in `data` for URLFWD this can probably be simplified to

```suggestion
        _type = data.get('type', 'URLFWD')
```

octodns/provider/cloudflare.py
```diff
@@ -537,12 +617,28 @@ def _gen_key(self, data):
...
+            parsed_uri = urlsplit(uri)
+            ret = {}
+            ret.update(data['actions'][0]['value'])
+            ret.update({'name': parsed_uri.netloc, 'path': parsed_uri.path})
+            urlfwd = (
+                '{name}',
+                '{path}',
+                '{url}',
+                '{status_code}')
+            return ' '.join(urlfwd).format(**ret)
```
ross, line 632: If I'm reading this correctly I think it can be simplified to:

```suggestion
            return '{name} {path} {url} {status_code}'.format(name=parsed_uri.netloc, 
                                                              path=parsed_uri.path, 
                                                              **data['actions'][0]['value'])
```

octodns/provider/cloudflare.py
```diff
@@ -555,14 +651,27 @@ def _apply_Update(self, change):
...
+                uri = '//' + uri if not uri.startswith('http') else uri
+                parsed_uri = urlsplit(uri)
+                name = zone.hostname_from_fqdn(parsed_uri.netloc)
+                _path = parsed_uri.path
+                # assumption, actions will always contain 1 action
+                _values = record['actions'][0]['value']
+                _values['path'] = _path
+                _values['ttl'] = 3600
+                _values['type'] = 'URLFWD'
+                record.update(_values)
```
ross, line 665:
```suggestion
                path = parsed_uri.path
                # assumption, actions will always contain 1 action
                values = record['actions'][0]['value']
                values['path'] = path
                values['ttl'] = 3600
                values['type'] = 'URLFWD'
                record.update(values)
```

octodns/record/__init__.py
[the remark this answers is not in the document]
ross: Does that mean this todo & code validation stuff should be removed or just that the TODO doesn't apply.

ross (commented):
Overall looks good. Few comments/questions inline and a thing or to two check.

octodns/provider/cloudflare.py
brianeclow: Oops. That default check is from https://github.com/octodns/octodns/pull/210

CloudFlare uses a TTL of 1 as an alias to a TTL of 300. Our test domain was littered with the default TTLs, and I was getting frustrated fighting the another teams automation for their POC.

The hard coded 3600 TTL for CloudFlare `URLFWD` was to handle CloudFlare's pagerules do not have TTLs. While NS1, and potentially others, might have TTLs associated with their redirection service. However, the point is taken, if I was following pattern, I should use a faux TTL of 300 instead of 3600.

octodns/provider/cloudflare.py
brianeclow: There is no notes field in the CloudFlare API. Compared to the zone API, it is missing a number of things.

I thought I checked the TTL change occurrence. I apparently, didn't. I add the handling in the referred def.

octodns/record/__init__.py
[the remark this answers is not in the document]
brianeclow: The TODO doesn't apply. And apparently, I missed removing the _this_ TODO comment, but I got the other one from this file. 🤦

octodns/provider/cloudflare.py
brianeclow: Actually, this was to build a faux DNS record to populate. Applying this change causes problems. And I didn't realize it until after I applied.

octodns/provider/cloudflare.py
brianeclow: I think I handled this in the prescribed manor via the `_include_change` def. Should I leave the magic numbers in place?

octodns/provider/cloudflare.py
ross: Weird. Wonder why i never shipped that. guess it just got forgotten.

ross (approved):
I think we're good here. Unfortunately I don't have Cloudflare access (set up) atm so I can't really kick the tires. My only/main concern is around the ttl default handling changes, but afaict this makes sense. If there are problems with it should just be ttl hiccups so that's not super serious...

PR #724, opened 2021-06-10
Ensure that dynamic records map to unique Azure TM profiles

viranch:
... and throw exception with a collision is detected.

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -632,6 +632,10 @@ class AzureProvider(BaseProvider):
             else:
                 # dynamic record alias is broken, return dummy value and apply
                 # will likely overwrite/fix it
+                self.log.warn('_data_for_CNAME: Missing Traffic Manager '
+                              'alias for dynamic CNAME record %s, forcing '
+                              're-link by setting an invalid value',
+                              azrecord.fqdn)
                 return {'value': 'iam.invalid.'}
 
         return {'value': _check_endswith_dot(azrecord.cname_record.cname)}
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -259,8 +259,21 @@ def _parse_azure_type(string):
     return string.split('/')[-1]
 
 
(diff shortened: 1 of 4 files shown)


octodns/provider/azuredns.py
```diff
@@ -259,8 +259,21 @@ def _parse_azure_type(string):
     return string.split('/')[-1]
 
 
-def _traffic_manager_suffix(record):
-    return record.fqdn[:-1].replace('.', '-')
+def _root_traffic_manager_name(record):
+    # ATM names can only have letters, numbers and hyphens
+    # replace dots with double hyphens to ensure unique mapping,
+    # hoping that real life FQDNs won't have double hyphens
+    return record.fqdn[:-1].replace('.', '--')
```
ross, line 266: Naming character restrictions like this are so annoying. Computers don't care what characters are in there... :grin:

octodns/provider/azuredns.py
```diff
@@ -613,8 +626,13 @@ def _data_for_CNAME(self, azrecord):
...
             :type  return: dict
         '''
-        if azrecord.cname_record is None and azrecord.target_resource.id:
-            return self._data_for_dynamic(azrecord)
+        if azrecord.cname_record is None:
+            if azrecord.target_resource.id:
+                return self._data_for_dynamic(azrecord)
+            else:
+                # dynamic record alias is broken, return dummy value and apply
+                # will likely overwrite/fix it
```
ross, line 634: Probably worth a warning log here with a bit more info about what's happening, like the comment.

ross (commented):
Looks good overall. Comment/suggestion inline.

octodns/provider/azuredns.py
viranch: pushed new commit for this

octodns/provider/azuredns.py
viranch: There go my plans of using emojis for traffic manager names 👎

ross (approved):

PR #726, opened 2021-06-16
Add alt text to images

eilla1:
This PR adds alt text to the images in the README for accessibility. 
Also removes the `=` from the example commands in the README. 

When I did `$ octodns-sync --config-file=./config/production.yaml` it was throwing an error. 
Then, when i did `$ octodns-dump`, it prompted me to put it in this format: `$ octodns-sync --config-file CONFIG_FILE` (without the = sign). Thus, I am making a PR to reflect this change in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<img src="https://raw.githubusercontent.com/octodns/octodns/master/docs/logos/octodns-logo.png?" height=251 width=404>
+<img src="https://raw.githubusercontent.com/octodns/octodns/master/docs/logos/octodns-logo.png?" alt="OctoDNS Logo" height=251 width=404>
 
 ## DNS as code - Tools for managing DNS across multiple providers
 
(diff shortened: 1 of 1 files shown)


ross (commented):
👍, thanks for getting the ALT tags.

Did you run into problems using `=` in the cmdline? It should work either way.

eilla1:
Ah yes I did run into problems when using the `=` but now I realize I might have done something else wrong and the `=` was not the cause of the error.

Just tried it with the `=` and it was successful. Is using the `=` preferable? If so, should I revert the changes?

ross:
> Just tried it with the `=` and it was successful. Is using the `=` preferable? If so, should I revert the changes?

Yeah. There's no hard rules about it, but generally it's most common to use `=` with `--foo` options and spaces with `-f`. Not finding great references on that, but https://unix.stackexchange.com/a/573378 at least says as much.

The python `ArgParse` stuff works either way, but given that and the fact that examples elsewhere in issues and blog posts generally use `=` I'd prefer to leave them in place without a specific reason to make the switch.

Will :eyes: and ✅ once the `=` bits are reverted. Thanks again.

eilla1:
oh interesting! TIL; thanks for the information 😄

ross (approved):

PR #729, opened 2021-06-21
Feature: optionally merge records values

sielaq:
Hi fellows,

Let me introduce to you small "merge" feature.
This might be useful for simple DNS providers,
that does not support dynamic DNS.
With that feature, enabling/disabling traffic for a particular region/datacenter
becomes very simple and code duplication is being avoided.

What do you think about it ?

### DIFF
--- a/docs/records.md
+++ b/docs/records.md
@@ -135,3 +135,121 @@ dyn:
     update_pcent_threshold: 0.4
     delete_pcent_threshold: 0.4
 ````
+
+#### Merge records of defined types
+
+For some simple providers (like Google DNS) it is not possible to do dynamic DNS.
+If you wish to keep your configuration simple and regions (or datacenters) separated
+(so you can manually switch regions quite fast) this feature might help you.
+
+Merging is done in a very simple way: ONLY values are being appended,
+the rest of the parameters (like ttl) are overwritten
+
+Which means you should be able to make a bit sophisticated configuration for edge cases:
+
+common/octodns.com.yaml
+````yaml
+---
+'':
+  - ttl: 60
+    type: TXT
+    value: "this stays untouched"
+````
+
+
+ams/octodns.com.yaml
+
+````yaml
+---
+'':
+  - ttl: 60
+    type: A
+    values:
+      - 1.1.1.1
+'www':
+  - ttl: 60
+    type: A
+    values:
+      - 2.2.2.2
+  - ttl: 60
+    type: TXT
+    value: "can be skipped if desired"
+````
+
+fra/octodns.com.yaml
+````yaml
+---
+'':
+  - ttl: 60
+    type: A
+    values:
+      - 1.1.1.2
+'www':
+  - ttl: 60
+    type: A
+    values:
+      - 2.2.2.3
+  - ttl: 60
+    type: TXT
+    value: "can be skipped if desired too"
+````
+
+dual_dc.yaml
+```yaml
+---
+providers:
+  common:
+    class: octodns.provider.yaml.YamlProvider
+    directory: ./common
+
+  ams:
+    class: octodns.provider.yaml.YamlProvider
+    directory: ./ams
+    merge_types:
+      - A
+    merge_skip_types:
+      - TXT
+
+  fra:
+    class: octodns.provider.yaml.YamlProvider
+    directory: ./fra
+    merge_types:
+      - A
+    merge_skip_types:
+      - TXT
+
+zones:
+  octodns.com.:
+    sources:
+      - common
+      - ams
+      - fra
+    targets:
+      - pdns
+
+````
+
+single_dc_ams.yml
+```yaml
+---
+providers:
+  common:
+    class: octodns.provider.yaml.YamlProvider
+    directory: ./common
+
+  ams:
+    class: octodns.provider.yaml.YamlProvider
+    directory: ./ams
+    merge_types:
+      - A
+    merge_skip_types:
+      - TXT
+zones:
+  octodns.com.:
+    sources:
+      - common
+      - ams
+    targets:
+      - pdns
+
+````
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -30,8 +30,22 @@ class YamlProvider(BaseProvider):
         # (optional, default True)
         enforce_order: true
         # Whether duplicate records should replace rather than error
-        # (optiona, default False)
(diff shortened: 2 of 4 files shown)


sbogun (approved):
LGTM though I cannot judge the functionality of the change.

ross:
Interesting concept and thanks for the PR to show the details. I don't think I would have been able to grok it without the code to :eyes:. My main concern here is that it would only work with the Yaml provider, but is implemented across both Yaml and Zone. There are a number of use cases that don't utilize YamlProvider as the source. The main one I can imagine that's relevant to this concept would be syncing from a dynamic provider, e.g. Route53 to a non-dynamic one, e.g. PowerDNS. 

Also I may be misunderstanding the details, but since the logic is happening at the source level it's not clear how it'd be possible to have both dynamic and non-dynamic (merged values) targets driven off the same set of sources, since zones can't be repeated in a config. 

Based on the above you might have a look at https://github.com/octodns/octodns/pull/637. It's a mostly functional PR that just needs a reason to :ship:. I should probably go ahead and get done. Anyway, I believe a Processor as envisioned by that PR may provide a cleaner way to go about this that would work regardless of source. In theory a DynamicToMerged processor could be created that looks at whether the target provider `SUPPORTS_DYNAMIC`. If `True` it could leave things untouched and if `False` it could merge them as you're looking for here. See `process_target_zone` https://github.com/octodns/octodns/pull/637/files#diff-46999229b2f08e09df048a40dcb20ea7ad1c94e777de344c9525c2c58bcd5437R23-R24. 

Anyway, have a look and let me know what you think.

sielaq:
Hi @ross , thanks for taking a look into the code.

> My main concern here is that it would only work with the Yaml provider, but is implemented across both Yaml and Zone. There are a number of use cases that don't utilize YamlProvider as the source.

Yes, the idea is for the yaml provider currently [ especially to help those, who keep everything as a code (IaaC, DNSaaC) and must maintain that]. And yes, it uses Zone as a place for merging since this is the only place where you can have a big picture about the whole zone ( you can have many configuration sources files for a one zone already ). And I spot this is a good place to spread the idea for other source providers.

> The main one I can imagine that's relevant to this concept would be syncing from a dynamic provider, e.g. Route53 to a non-dynamic one, e.g. PowerDNS

Currently it is agnostic from other source providers, it just won’t be used yet in other than yaml provider, until somebody will have a need to – we could implement it.

> ... it's not clear how it'd be possible to have both dynamic and non-dynamic (merged values) targets driven off the same set of sources, ...

If you treat source provider as a pre-processed level – where zones can be repeated from multiple sources – so on zone level you can decide: skip or merge. I would rather not mix dynamic and non-dynamic  usecases and just using this feature for some edge cases.

> since zones can't be repeated in a config

yes and no. This is not fully true.  You can have multiple source configs already (I’m talking about current code in master branch) and each zone can be repeated without any error – until the **type** (A,AAAA,TXT…)  in the same zone is not repeated.  So the current code already does concatenation – a simple merging - which is fine and cool and I really like it!
You can have already repeated zones in a configs that after processing generates non-repeatable zones configuration in the end.  Mine PR just put it to next level - where values of same type are appended if someone have a wish.

I hope I have clarified the concept.

Regarding the mentioned PR it is a little bit to fuzzy for me to understand that idea,
I would need more code examples how to use it :(

best regards
@sielaq 

ps. greeting from Europe
sorry for delay - a different timezone

ross:
> Currently it is agnostic from other source providers, it just won’t be used yet in other than yaml provider, until somebody will have a need to – we could implement it.

So in general the providers should not be exposed to this sort of detail and each be required to implement identical logic to get the functionality. It violates the object-oriented principal of encapsulation. The logic about how to do something should not be spread across multiple places, and in this case both zone and providers are having to support it, and for provider every individual subclass would need to do so.

Thus my suggestion of the `Processors` route. It should be able to implement similar functionality (that is also more inline with octoDNS's primary design goal (discussed in the next response below) in a way that's done once, self contained, and can be used for any existing or future providers. 

This is pseudocode as I'd have to sit down and write up the whole thing to get actual details, but something like the following would be a rough sketch of what the Provider might look like:

```python
class DynamicMerging(BaseProcessor):

    def _maybe_merge_values(self, record):
        # logic for when to merge values and when to leave them alone goes here. 
        # if we're leaving it alone we just return it straight away
        if doing_nothing:
            return record
        # if we're merging values then we do that here
        ...
        return munged_record

    def process_target_zone(self, zone, target):
        if target.SUPPORTS_DYNAMIC:
            # this provider can handle dynamic records, give it the zone as-is
            return zone

        # this zone does not support dynamic records, we need to run through the zone 
        # and merge values based on our settings/config
        ret = self._clone_zone(zone)
        for record in zone.records:
            if record._type in self.allowlist:
                ret.add_record(self._maybe_merge_values(record))

        return ret
```

Lmk if that helps illustrate the potential processor route. 

> > since zones can't be repeated in a config
> 
> yes and no. This is not fully true. You can have multiple source configs already (I’m talking about current code in master branch) and each zone can be repeated without any error – until the **type** (A,AAAA,TXT…) in the same zone is not repeated. So the current code already does concatenation – a simple merging - which is fine and cool and I really like it!
> You can have already repeated zones in a configs that after processing generates non-repeatable zones configuration in the end. Mine PR just put it to next level - where values of same type are appended if someone have a wish.

Sorry, I don't think I clearly explained what I was trying to say. The primary use case of octoDNS is to push source data to multiple targets and keep them in sync, e.g:

```yaml
foo.com.:
  sources:
  - config
  targets:
  - route53
  - dnsimple
```

This is normally done with a single `config-file` for all sources & targets in an environment and not one for simple providers and another for dynamic ones as the example shows. It's best to be able to make a change and run a single thing and see all the changes that are going to be made in ~one place rather than having to run multiple plans and work through/compare the outputs of each. 

If we take an example from this PR into a setup like that we'd get something in the general shape of:

```yaml
foo.com.:
  sources:
   - common
   - ams
  targets:
  - route53
  - dnsimple
```

I believe the issue is that both `route53` and `dnsimple` would get the "merged" values even though `route53` can support dynamic records. There'd be no way to use support both `dynamic` and non providers with a single config which violates the fundamental expected/required behavior of octoDNS.

> ps. greetings from Europe
> sorry for delay - a different timezone

No worries. Half my team at github were in the EU so I'm used to it and I normally get on things earlier in the day, but was AFK the for the most part the past couple days which delayed my responses.

PR #730, opened 2021-06-21
Partial support for dynamic A+AAAA records on Azure 

viranch:
Azure DNS A/AAAA records cannot be aliased to Traffic Managers that contain other nested Traffic Managers. Due to this limitation, A/AAAA dynamic records can only do one of geo-fencing, fallback and weighted RR.

This PR implements the limited functionality, with checks in place to prevent unsupported scenarios.

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -299,6 +299,47 @@ def _get_monitor(record):
     return monitor
 
 
+def _check_valid_dynamic(record):
+    typ = record._type
+    dynamic = record.dynamic
+    if typ in ['A', 'AAAA']:
+        # A/AAAA records cannot be aliased to Traffic Managers that contain
+        # other nested Traffic Managers. Due to this limitation, A/AAAA
+        # dynamic records can do only one of geo-fencing, fallback and
+        # weighted RR. So let's validate that the record adheres to this
+        # limitation.
+        data = dynamic._data()
+        values = set(record.values)
+        pools = data['pools'].values()
+        seen_values = set()
+        rr = False
+        fallback = False
+        for pool in pools:
+            vals = pool['values']
+            if len(vals) > 1:
+                rr = True
+            pool_values = set(val['value'] for val in vals)
+            if pool.get('fallback'):
+                fallback = True
+            seen_values.update(pool_values)
+
+        if values != seen_values:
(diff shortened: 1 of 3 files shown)


octodns/provider/azuredns.py
```diff
@@ -805,23 +841,58 @@ def _data_for_dynamic(self, azrecord):
...
+                geo = any(r.get('geos') for r in data['rules'])
+
+                if [rr, fallback, geo].count(True) > 1:
+                    msg = ('{} {}: A/AAAA dynamic records must use at most '
+                           'one of round-robin, fallback and geo-fencing')
+                    msg = msg.format(record.fqdn, record._type)
+                    raise AzureException(msg)
+            elif typ != 'CNAME':
+                msg = ('{}: Dynamic records in Azure must be of type '
+                       'A/AAAA/CNAME').format(record.fqdn)
```
ross, line 895: This method is getting pretty long. May be wroth breaking up some of these branches/bits into helper functions to aid readability. Doesn't have to happen right now if you don't want. It can be in a dedicated PR or something if you'd prefer (if done at all.)

ross (approved):
Will give this a day to hear back whether you want to break up the extra changes bit here, in another PR, or not at all.

octodns/provider/azuredns.py
viranch: 👍 good call. updated the PR.

ross:
You ready for a final :eye: on this or still want to tweak it more?

viranch:
This is ready now.

octodns/provider/azuredns.py
```diff
@@ -263,7 +269,10 @@ def _root_traffic_manager_name(record):
     # ATM names can only have letters, numbers and hyphens
     # replace dots with double hyphens to ensure unique mapping,
     # hoping that real life FQDNs won't have double hyphens
-    return record.fqdn[:-1].replace('.', '--')
+    name = record.fqdn[:-1].replace('.', '--')
+    if record._type != 'CNAME':
+        name += '-{}'.format(record._type)
```
ross, line 274: Heh. Danger of implementing one thing at a time, I guess. There are other places where I ended up having to stuff the type into the names for this same reason.

ross (approved):

PR #731, opened 2021-06-23
fix for NA continent geo target limitation on NS1

meghashyamps:
image

This is to fix the limitation when the continent `NA` is used in geos. It was translating to `us-west`, `us-east` and `us-central` and not really taking into account all the other countries that are part of `NA` like `CA`, `MX` etc.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -363,8 +363,7 @@ class Ns1Provider(BaseProvider):
         'NA': {'DO', 'DM', 'BB', 'BL', 'BM', 'HT', 'KN', 'JM', 'VC', 'HN',
                'BS', 'BZ', 'PR', 'NI', 'LC', 'TT', 'VG', 'PA', 'TC', 'PM',
                'GT', 'AG', 'GP', 'AI', 'VI', 'CA', 'GD', 'AW', 'CR', 'GL',
-               'CU', 'MF', 'SV', 'US', 'UM', 'MQ', 'MS', 'KY', 'MX', 'CW',
-               'BQ', 'SX'}
+               'CU', 'MF', 'SV', 'US', 'MQ', 'MS', 'KY', 'MX', 'CW', 'BQ'}
     }
 
     def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -341,6 +341,9 @@ class Ns1Provider(BaseProvider):
         'ASIAPAC': 'AS',
         'EUROPE': 'EU',
         'SOUTH-AMERICA': 'SA',
+        # continent NA has been handled as part of Geofence Country filter
+        # starting from v0.9.13. These below US-* just need to continue to
+        # exist here so it doesn't break the ugrade path
         'US-CENTRAL': 'NA',
         'US-EAST': 'NA',
         'US-WEST': 'NA',
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -350,8 +350,6 @@ class Ns1Provider(BaseProvider):
         'AS': ('ASIAPAC',),
         'EU': ('EUROPE',),
         'SA': ('SOUTH-AMERICA',),
-        # TODO: what about CA, MX, and all the other NA countries?
-        'NA': ('US-CENTRAL', 'US-EAST', 'US-WEST'),
     }
 
     # Necessary for handling unsupported continents in _CONTINENT_TO_REGIONS
@@ -359,6 +357,11 @@ class Ns1Provider(BaseProvider):
         'OC': {'FJ', 'NC', 'PG', 'SB', 'VU', 'AU', 'NF', 'NZ', 'FM', 'GU',
                'KI', 'MH', 'MP', 'NR', 'PW', 'AS', 'CK', 'NU', 'PF', 'PN',
                'TK', 'TO', 'TV', 'WF', 'WS'},
+        'NA': {'DO', 'DM', 'BB', 'BL', 'BM', 'HT', 'KN', 'JM', 'VC', 'HN',
+               'BS', 'BZ', 'PR', 'NI', 'LC', 'TT', 'VG', 'PA', 'TC', 'PM',
+               'GT', 'AG', 'GP', 'AI', 'VI', 'CA', 'GD', 'AW', 'CR', 'GL',
+               'CU', 'MF', 'SV', 'US', 'UM', 'MQ', 'MS', 'KY', 'MX', 'CW',
+               'BQ', 'SX'}
     }
 
     def __init__(self, id, api_key, retry_count=4, monitor_regions=None,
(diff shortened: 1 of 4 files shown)


ross (commented):
This looks good and seems like a positive improvement/change. Can you please write up a CHANGELOG entry in this PR mentioning this change since it's a fairly noteworthy one.

You can just add a new header at the top like:

    v0.9.13 - 2021-0*-** -

and otherwise include it in a [`Noteworthy Changes` section like this](https://github.com/octodns/octodns/blob/master/CHANGELOG.md#noteworthy-changes).

ross:
I shipped https://github.com/octodns/octodns/pull/637 which added a new changelog section so this can just be added to that.

meghashyamps:
> I shipped #637 which added a new changelog section so this can just be added to that.

Done

octodns/provider/ns1.py
```diff
@@ -350,15 +350,18 @@ def _FILTER_CHAIN_WITH_REGION_AND_COUNTRY(self, with_disabled):
         'AS': ('ASIAPAC',),
         'EU': ('EUROPE',),
         'SA': ('SOUTH-AMERICA',),
-        # TODO: what about CA, MX, and all the other NA countries?
-        'NA': ('US-CENTRAL', 'US-EAST', 'US-WEST'),
```
ross, line 354: Will `_REGION_TO_CONTINENT` above still need it's `US-*` cases to handle the upgrade path? I believe that's the case. Probably makes sense to add a comment up there to that effect to explain why they need to stick around, for now.

ross (commented):
Overall looks good. I want to do a bit of integration tessting doing upgrades/downgrades to make sure everything is happy before final review of this one since it's touching some fairly sensitive areas.

octodns/provider/ns1.py
meghashyamps: Yes, thats why I left this block untouched:
```
    _REGION_TO_CONTINENT = {
        'AFRICA': 'AF',
        'ASIAPAC': 'AS',
        'EUROPE': 'EU',
        'SOUTH-AMERICA': 'SA',
        'US-CENTRAL': 'NA',
        'US-EAST': 'NA',
        'US-WEST': 'NA',
    }
```
Adding a comment.

meghashyamps: Done

ross:
So testing the upgrade paths here:

## From master to this PR

Looks like post update doesn't plan any changes when there's a NA rule:

```yaml
---
? ''
: dynamic:
    pools:
      fra:
        fallback: iad
        values:
        - value: 1.1.1.2
      iad:
        values:
        - value: 1.1.1.1
      sea:
        fallback: iad
        values:
        - value: 1.1.1.3
    rules:
    - geos:
      - EU
      pool: fra
    - geos:
      - NA
      pool: sea
    - pool: iad
  ttl: 60
  type: A
  values:
  - 1.1.1.4
```

```
2021-06-24T15:57:36  [4798594560] INFO  Manager sync:   0 total changes
(env) coho:octodns ross$ hub pr checkout 731
remote: Enumerating objects: 46, done.
remote: Counting objects: 100% (40/40), done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 22 (delta 17), reused 17 (delta 13), pack-reused 0
Unpacking objects: 100% (22/22), 3.62 KiB | 92.00 KiB/s, done.
From github.com:octodns/octodns
 * [new ref]         refs/pull/731/head -> NA-limitation-fix
Switched to branch 'NA-limitation-fix'
(env) coho:octodns ross$ PYTHONPATH=. ./octodns/cmds/sync.py --config-file=config/dev.yaml
2021-06-24T16:03:58  [4622114304] INFO  Manager __init__: config_file=config/dev.yaml
2021-06-24T16:03:58  [4622114304] INFO  Manager __init__:   max_workers=1
2021-06-24T16:03:58  [4622114304] INFO  Manager __init__:   include_meta=False
2021-06-24T16:03:59  [4622114304] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=True, force=False, plan_output_fh=<stdout>
2021-06-24T16:03:59  [4622114304] INFO  Manager sync:   zone=exxampled.com.
2021-06-24T16:03:59  [4622114304] INFO  Manager sync:   sources=['config'] -> targets=['ns1', 'route53']
2021-06-24T16:03:59  [4622114304] INFO  YamlProvider[config] populate:   found 1 records, exists=False
2021-06-24T16:03:59  [4622114304] INFO  Ns1Provider[ns1] plan: desired=exxampled.com.
2021-06-24T16:03:59  [4622114304] INFO  Ns1Provider[ns1] populate:   found 2 records, exists=True
2021-06-24T16:04:00  [4622114304] INFO  Ns1Provider[ns1] plan:   No changes
2021-06-24T16:04:00  [4622114304] INFO  Route53Provider[route53] plan: desired=exxampled.com.
2021-06-24T16:04:01  [4622114304] INFO  Route53Provider[route53] populate:   found 2 records, exists=True
2021-06-24T16:04:01  [4622114304] INFO  Route53Provider[route53] plan:   No changes
2021-06-24T16:04:01  [4622114304] INFO  Manager
No changes were planned
```

When i revert versions back to master there's obviously no further changes... This seems OK, but users won't get the "improvement" until they otherwise make a change to the record and it updates for that reason. 

If that's not the desired behavior here something can be placed in an `_extra_changes` method to see if it needs updated. This is one of the frequent complications with these sorts of improvements.

## From this PR to master

Going the other direction, e.g. someone upgrades and finds a problem and they need to revert... I deleted the record and then pushed it up with this PR's code and then checked out master and tried to run a sync, unfortunately that resulted in an error:

```
(env) coho:octodns ross$ PYTHONPATH=. ./octodns/cmds/sync.py --config-file=config/dev.yaml
2021-06-24T16:06:42  [4501724672] INFO  Manager __init__: config_file=config/dev.yaml
2021-06-24T16:06:42  [4501724672] INFO  Manager __init__:   max_workers=1
2021-06-24T16:06:42  [4501724672] INFO  Manager __init__:   include_meta=False
2021-06-24T16:06:42  [4501724672] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=True, force=False, plan_output_fh=<stdout>
2021-06-24T16:06:42  [4501724672] INFO  Manager sync:   zone=exxampled.com.
2021-06-24T16:06:42  [4501724672] INFO  Manager sync:   sources=['config'] -> targets=['ns1', 'route53']
2021-06-24T16:06:42  [4501724672] INFO  YamlProvider[config] populate:   found 1 records, exists=False
2021-06-24T16:06:42  [4501724672] INFO  Ns1Provider[ns1] plan: desired=exxampled.com.
  File "/Users/ross/octodns/octodns/./octodns/cmds/sync.py", line 44, in <module>
    main()
  File "/Users/ross/octodns/octodns/./octodns/cmds/sync.py", line 38, in main
    manager.sync(eligible_zones=args.zone, eligible_sources=args.source,
  File "/Users/ross/octodns/octodns/octodns/manager.py", line 373, in sync
    ps, d = future.result()
  File "/Users/ross/octodns/octodns/octodns/manager.py", line 58, in result
    return self.func(*self.args, **self.kwargs)
  File "/Users/ross/octodns/octodns/octodns/manager.py", line 263, in _populate_and_plan
    plan = target.plan(zone)
  File "/Users/ross/octodns/octodns/octodns/provider/base.py", line 51, in plan
    exists = self.populate(existing, target=True, lenient=True)
  File "/Users/ross/octodns/octodns/octodns/provider/ns1.py", line 790, in populate
    data = data_for(_type, record)
  File "/Users/ross/octodns/octodns/octodns/provider/ns1.py", line 630, in _data_for_A
    return self._data_for_dynamic(_type, record)
  File "/Users/ross/octodns/octodns/octodns/provider/ns1.py", line 572, in _data_for_dynamic
    con = country_alpha2_to_continent_code(country)
  File "/Users/ross/octodns/octodns/env/lib/python3.9/site-packages/pycountry_convert/convert_country_alpha2_to_continent_code.py", line 265, in country_alpha2_to_continent_code
    raise KeyError("Invalid Country Alpha-2 code: '{0}'".format(country_2_code))
KeyError: "Invalid Country Alpha-2 code: 'SX'"
```

It looks like you ran into the fact that they're unknown while working on this PR https://github.com/octodns/octodns/pull/731/files#diff-5a9a9700f1db072c33e6c3051652f72653eba706ffb0f3f48ea828d8cc131450R578-R579.

This is a complicated one since changes would need to exist to master to be able to cope with what this PR is doing... I think the change will need to be split across 2 rounds/PS across two version releases to cope with that cleanly. I can see two options:

1. This PR ships w/o those 2 special case countries included in `NA`, but including the special handling for them and after a release is made a 2nd PR adds those countries to `NA`. 
2. The reverse of 1), first ship support for those countries, wait for a release and then ship the `NA` support.

I'd lean towards 1, but could be convinced to take the other route if people prefer.

meghashyamps:
## Master to PR

> When i revert versions back to master there's obviously no further changes... This seems OK, but users won't get the "improvement" until they otherwise make a change to the record and it updates for that reason.

Yes, this is the desired behavior since if they were OK all this while with the current implementation of `NA` then there is no reason to forcefully hand them the improvement. But of course if they were to touch any relevant parts of their config in the future, they will discover about `NA` continent's improvement, at which point Octo will prompt for an update of the record. 

## From this PR to master

> Going the other direction, e.g. someone upgrades and finds a problem and they need to revert... I deleted the record and then pushed it up with this PR's code and then checked out master and tried to run a sync, unfortunately that resulted in an error:

I can think of 3 scenarios here:

1. Record created with master --> Octo updated to PR version --> SYNC shows no changes --> they discover a problem and revert back to master --> SYNC again shows no changes. (so all good here)
2. Record created with master --> Octo updated to PR version --> some related change caused Octo to update `NA` on that record --> they discover a problem and revert back to master --> SYNC fails with country `SX`/`UM` error. 
3. Record created with PR version --> they discover a problem and revert back to master --> SYNC fails with country `SX`/`UM` error. 

> I think the change will need to be split across 2 rounds/PS across two version releases to cope with that cleanly.

Yup, seems so. 

> 1. This PR ships w/o those 2 special case countries included in `NA`, but including the special handling for them and after a release is made a 2nd PR adds those countries to `NA`.
> 2. The reverse of 1), first ship support for those countries, wait for a release and then ship the `NA` support.

I am also leaning towards 1 mainly because that means having the support for `NA` sooner than later and also since `SX` and `UM` are not show-stopping countries, so we can live with **not** having them in NA for a while. 

I can make the change if it is a go-ahead.

ross:
> I can make the change if it is a go-ahead.

Got for it. Please go ahead and file the 2nd PR at some point, maybe easiest after this one is merged, and I'll hopefully remember to include it in the follow on release.

meghashyamps:
Removing those 2 countries from `_CONTINENT_TO_LIST_OF_COUNTRIES['NA']` basically skips the special handling `elif` block:
```
                elif country in ['SX', 'UM']:
                    con = 'NA'
```
which in turn is showing my coverage as 99% for `ns1.py` file
```
octodns/provider/ns1.py              689      1    278      1    99%   578
```

I cannot add a testcase to satisfy this lack of coverage since the `elif` block is never hit. How do I get get 100% coverage here?

ross:
> I cannot add a testcase to satisfy this lack of coverage since the `elif` block is never hit. How do I get get 100% coverage here?

You can likely directly test that in https://github.com/octodns/octodns/blob/9e085e9f4ef83f6ce721c521989e706a89c81501/tests/test_octodns_provider_ns1.py#L1364

I believe you should be able to hit that branch by adding one or both of those countries explicitly to the list in https://github.com/octodns/octodns/blob/9e085e9f4ef83f6ce721c521989e706a89c81501/tests/test_octodns_provider_ns1.py#L1574-L1581, but I haven't tried it but my guess is that'll hit it.

meghashyamps:
Ok, that worked. 

I will open the second PR once this is merged...

ross (approved):

meghashyamps:
@ross I opened https://github.com/octodns/octodns/pull/736 for the second part for next release

PR #743, opened 2021-07-26
AzureProvider: Support private zones

travislikestocode:
This change adds support for private zones in the Azure DNS provider. My approach was to add `private` as a parameter to the provider, rather than trying to do it on a per-zone basis. NS and CAA records are unsupported (by Azure, not just this provider). If a dynamic record is passed when private is enabled, the dynamic data will be ignored. This is because from what I can tell traffic managers for private zones are not a thing, so dynamic won't work, since it's implemented via traffic manager.

Need to update tests, but would love to get feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ It is similar to [Netflix/denominator](https://github.com/Netflix/denominator).
 - [Table of Contents](#table-of-contents)
 - [Getting started](#getting-started)
   - [Workspace](#workspace)
+    - [Installing a specific commit SHA](#installing-a-specific-commit-sha)
   - [Config](#config)
   - [Noop](#noop)
   - [Making changes](#making-changes)
@@ -192,7 +193,7 @@ The above command pulled the existing data out of Route53 and placed the results
 
 | Provider | Requirements | Record Support | Dynamic | Notes |
 |--|--|--|--|--|
-| [AzureProvider](/octodns/provider/azuredns.py) | azure-identity, azure-mgmt-dns, azure-mgmt-trafficmanager | A, AAAA, CAA, CNAME, MX, NS, PTR, SRV, TXT | Alpha (A, AAAA, CNAME) | |
+| [AzureProvider](/octodns/provider/azuredns.py) | azure-identity, azure-mgmt-dns, azure-mgmt-privatedns, azure-mgmt-trafficmanager | A, AAAA, CAA, CNAME, MX, NS, PTR, SRV, TXT | Alpha (A, AAAA, CNAME) | |
 | [Akamai](/octodns/provider/edgedns.py) | edgegrid-python | A, AAAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT | No | |
 | [CloudflareProvider](/octodns/provider/cloudflare.py) | | A, AAAA, ALIAS, CAA, CNAME, LOC, MX, NS, PTR, SPF, SRV, TXT | No | CAA tags restricted |
 | [ConstellixProvider](/octodns/provider/constellix.py) | | A, AAAA, ALIAS (ANAME), CAA, CNAME, MX, NS, PTR, SPF, SRV, TXT | No | CAA tags restricted |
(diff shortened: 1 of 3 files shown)


travislikestocode:
@viranch may be interested in this, since you've done so much work on the provider

README.md
```diff
@@ -15,6 +15,7 @@ It is similar to [Netflix/denominator](https://github.com/Netflix/denominator).
 - [Table of Contents](#table-of-contents)
 - [Getting started](#getting-started)
   - [Workspace](#workspace)
+    - [Installing a specific commit SHA](#installing-a-specific-commit-sha)
```
travislikestocode, line 18: Looks like my VS code extension added this without me noticing

octodns/provider/azuredns.py
```diff
@@ -538,8 +570,19 @@ def _check_zone(self, name, create=False):
...
+                create_method = 'begin_create_or_update'
+            else:
+                zone_model = Zone
+                create_method = 'create_or_update'
+
+            create_zone = attrgetter('_dns_client.{}.{}'.format(
+                                     self._zone_type, create_method))(self)
+            poller = create_zone(self._resource_group, name,
+                                 zone_model(location='global'))
+            self._private and poller.result()  # wait for zone creation
```
travislikestocode, line 585: Blocks on each zone creation. Less than ideal, obviously, but acceptable for now IMO.

viranch:
I am not fully aware of this Azure offering and how it differs from the public Azure DNS one. Depending on the differences, this product may not be a good fit or in scope of octoDNS.

If it is, I would keep it as a separate provider that can perhaps be a subclass of the current `AzureProvider`. That would help keep different products isolated if things get hairy in future. Again, I'm not sure if this path of subclassing one provider into another works with the octoDNS paradigm of independent providers.

ross:
Sorry. Been super occupied as of late and haven't really gotten a chance to dig into this PR yet.

> Again, I'm not sure if this path of subclassing one provider into another works with the octoDNS paradigm of independent providers.

If you have a look at the PowerDNS provider there's some prior art for providers that inherit from a base to add support for specific situations. If adding private zone support for azure is getting hairy and complicating things then it may well make sense to refactor things here a bit and pull out the "core" stuff into a base class from which the private and dynamic (if I'm understanding the discussion correctly) could inherit. Again no clue if that makes sense without having had a chance to dig in, but it's definitely an option if it does.

README.md
ross: Guess it's not wrong.

ross (commented):
Had a chance to look this over and think about it a bit. It's really annoying how Azure has done PrivateDNS as a separate thing with an similar api and just slight changes here and there. 

It probably makes sense to have a shared base class, `AzureBaseProvider` here and then separate `AzureProvider`  and `AzurePrivateProvider`.  

The base class would likely be all the preexisting `AzureProvider` code and then the two child classes would cover the `if`'s added in this PR. Lmk if you have a feel for what that'd look like. If not I can probably take a stab at it, though I don't really have a setup to test it out atm.

travislikestocode:
> I am not fully aware of this Azure offering and how it differs from the public Azure DNS one. Depending on the differences, this product may not be a good fit or in scope of octoDNS.

Oops forgot to include some background info. Private zones are a brand of DNS zone that is only accessible on Azure internal vnets. They're nearly identical to public DNS from a data perspective, only a few small differences.

> Had a chance to look this over and think about it a bit. It's really annoying how Azure has done PrivateDNS as a separate thing with an similar api and just slight changes here and there.

Agreed!

> It probably makes sense to have a shared base class, AzureBaseProvider here and then separate AzureProvider and AzurePrivateProvider.

> The base class would likely be all the preexisting AzureProvider code and then the two child classes would cover the if's added in this PR. Lmk if you have a feel for what that'd look like. If not I can probably take a stab at it, though I don't really have a setup to test it out atm.

This makes total sense, and after reading a bit more I think I can run with it. Thanks for the feedback, @ross!

ross:
Closing this as effectively stale and the provider has been moved out into its own repo https://github.com/octodns/octodns-azure/. Feel free to rework/reopen this PR there.

Issue #702, opened 2021-04-30
Allow failover configuration in rules


viranch:
I have a use-case for configuring a different failover path/order per region.

For example, I have local presence in LHR (UK), FRA (Germany) and IAD (US). I want UK to failover to FRA, while Germany should failover to LHR (both eventually failover to IAD). If I try to configure this in the pools, it will run into a failover loop. I could create copies of `lhr` and `fra` but that's a long workaround with same IPs in multiple pools, making any changes hard and error prone:

```yaml
  dynamic:
    pools:
      lhr-fra:
        fallback: fra
        values:
        - value: 1.1.1.1
      fra-lhr:
        fallback: lhr
        values:
        - value: 2.2.2.2
      lhr:
        fallback: iad
        values:
        - value: 1.1.1.1
      fra:
        fallback: iad
        values:
        - value: 2.2.2.2
      iad:
        values:
        - value: 3.3.3.3
    rules:
    - geos:
      - EU-UK
      pool: lhr-fra
    - geos:
      - EU-DE
      pool: fra-lhr
    - pool: iad
```

Instead, it will be much cleaner to be able to configure the failover in rules section, something like so:

```yaml
  dynamic:
    pools:
      lhr:
        values:
        - value: 1.1.1.1
      fra:
        values:
        - value: 2.2.2.2
      iad:
        values:
        - value: 3.3.3.3
    rules:
    - geos:
      - EU-UK
      pools: [lhr, fra, iad]
    - geos:
      - EU-DE
      pools: [fra, lhr, iad]
    - pool: iad
```

This is how its represented in Dyn anyway. NS1 can also be setup to do this. I'm not sure about Route53. Is it something octoDNS can add support for?

ross:
> This is how its represented in Dyn anyway. NS1 can also be setup to do this. I'm not sure about Route53. Is it something octoDNS can add support for?

It would be a fairly significant change to the details of dynamic records likely resulting in a fairly major chunk of work on the existing dynamic providers and probably some on the core of octoDNS itself. 

As far as why it is the way it is basically it wasn't a use case that was required in my use at the time. I vaguely remember playing around with whether fallbacks should belong to pools (as is currently the case) or rules (as you propose) and I remember picking the way it is now for reasons, but unfortunately I don't appear to remember the specifics of what they were. 

It may have been problems with what Route53 could cleanly support.

viranch:
[...]

PR #746, opened 2021-07-30
check for identical monitor and skip creating one if found

meghashyamps:
```
www:
  dynamic:
    pools:
      apac:
        fallback: na
        values:
        - value: 1.1.1.1      <---- repeats
      na:
        values:
        - value: 2.2.2.2
        - value: 1.1.1.1      <---- repeats
    rules:
    - geos:
      - AS
      pool: apac
    - geos:
      - NA
      pool: na
    - pool: na
  octodns:
    healthcheck:
      host: null
      path: /pop/admin
      port: 443
      protocol: HTTPS
  ttl: 300
  type: A
  value: 4.4.4.4
```
This PR implements monitor reuse instead of creating multiple identical ones. As seen in the example above, creating dynamic record was failing since the answer `1.1.1.1` is present in both pools `apac` and `na`. This PR should fix that.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -1162,15 +1162,21 @@ class Ns1Provider(BaseProvider):
 
         # Build a list of primary values for each pool, including their
         # feed_id (monitor)
+        value_feed = dict()
         pool_answers = defaultdict(list)
         for pool_name, pool in sorted(pools.items()):
             for value in pool.data['values']:
                 weight = value['weight']
                 value = value['value']
-                existing = existing_monitors.get(value)
-                monitor_id, feed_id = self._monitor_sync(record, value,
-                                                         existing)
-                active_monitors.add(monitor_id)
+                feed_id = value_feed.get(value)
+                # check for identical monitor and skip creating one if found
+                if not feed_id:
+                    existing = existing_monitors.get(value)
+                    monitor_id, feed_id = self._monitor_sync(record, value,
+                                                             existing)
+                    value_feed[value] = feed_id
+                    active_monitors.add(monitor_id)
+
                 pool_answers[pool_name].append({
                     'answer': [value],
                     'weight': weight,
--- a/tests/test_octodns_provider_ns1.py
+++ b/tests/test_octodns_provider_ns1.py
@@ -1332,6 +1332,10 @@ class TestNs1ProviderDynamic(TestCase):
         # This indirectly calls into _params_for_dynamic and tests the
         # handling to get there
         record = self.record()
+        # copy an existing answer from a different pool to 'lhr' so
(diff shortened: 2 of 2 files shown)


ross (commented):
Does this correctly handle the case where deleting the value from one of the two locations? That case would need to leave the monitor in place and only delete it if both were gone.

meghashyamps:
Yup, it does.

ross:
Found a general issue with NS1 record cleanup while digging into things related to this PR https://github.com/octodns/octodns/pull/748

meghashyamps:
@ross are we going to merge this? Even though we solved the problem with https://github.com/octodns/octodns/pull/749 (the duplicate notification list the cause of this error) , it still might be useful to **not** have identical duplicate monitors if we can reuse them.

viranch:
This is still needed because octoDNS will try to create a duplicate monitor which will fail due to name collision.

ross:
Main thing slowing this down is having values duplicated across pools wasn't really an intended use case as I couldn't really come up with non-contrived cases where it made sense. In fact before this issue/PR I would have guessed there were validations in place that prevented it. What's the setup that's lead to it being needed?

It's quite possible the other Dynamic providers have issues with it as well. Anyone tried Route53? (don't really care about Dyn at this point.)

viranch:
> What's the setup that's lead to it being needed?

The workaround described in #702 leads to duplicate pool values.

meghashyamps:
Actually, NS1 allows identical monitors to be created, even with same names. What got me working on this was the fact that they both were trying to create identical notification lists, which in fact has validation for name collision and was failing. 

image

viranch:
My bad. This is not functionally blocking then. But I would still prefer avoiding the wastage by duplicate monitors.

ross:
> Route53

Looks like Route53 is OK with shared values and only creates a single health check for each value (in quick/simple testing.)

ross:
There's more going on that needs addressing here as well. Now that duplicate values are possible with `shared_notifylists` the `Ns1Provider` doesn't clean up monitors when a duplicate is removed. The dup gets stranded and even deleting record doesn't get rid of it ☹️

ross:
OH. The monitors that are returned from `_monitors_for` are keyed by value thus is ~random chance whether you get back the one that needs to go away or the one that is still needed. This is going to be a bigger rework/fix than this PR 😞

ross (approved):
So this PR prevents the duplicate monitor from being created and thus the racy deletion problems mentioned in previous problems don't come about in the first place. I'd consider this a bug fix even though it was filed before the bug was possible to hit :grin:

The only people who could hit this are those running against unreleased HEAD sha's from the last few days etc so don't think anything special needs to be done beyond fixing the problem.

PR #748, opened 2021-08-01
NS1 _apply_Update needs to gc monitors w/existing

ross:
monitors_for in the Ns1Provider needs to be passed the old/existing record when doing an update as that's the one that may have had monitors that need to be cleaned up. Prior to this change a dynamic -> normal record update wouldn't find any existing monitors and thus would never clean up the old monitors. 

https://github.com/octodns/octodns/blob/a1ef4206682e10aa818b6fc4f02fa967a33700c2/octodns/provider/ns1.py#L861-L864

/cc @pavantc for review (should be able to assign soon)


pavantc (approved):
LGTM.
Sorry about the delay @ross, was a long weekend here :)

Issue #744, opened 2021-07-27
NS1 Provider: Dynamic Records - NotifyList too many characters


davidporsche:
I have run into an issue while testing out dynamic records using the NS1 provider that manifests itself with the following error.

`"ns1.rest.errors.ResourceException: server error: Input validation failed (Value u'dynamic-routing-record.poc-test-octo.example.com - A - 10.10.10.10' for field '<obj>.name' must have length less than or equal to 64)"`

The error appears to be a limitation of the NS1 API not accepting strings longer than 64 characters for their healthchecks / monitors.  The way that octodns assembles the string used for the name of the healthcheck, more times than not, I'd  run into this issue with the majority of my dynamic services.

When I reduce the number of characters in either the name of the dynamic record itself or the target values in the dynamic pools, I can get past this issue, but this is not ideal or always possible.

ross:
> When I reduce the number of characters in either the name of the dynamic record itself or the target values in the dynamic pools, I can get past this issue, but this is not ideal or always possible.

Not really sure what (good options) we'd have here. Only thing I can think of would be to detect when the name is going to be too long and to replace some or all of it with a consistent hash that is short enough. Note that it would make the UI and/or alerts pretty hard to grok, but might solve the problem.

e.g. maybe:

```
dynamic-routing-record.poc-test-octo.example.com - A - 10.10.10.10
```

could be converted to something in the shape of

```
dynamic-ro/01cf2a9b/ple.com - A - 10.10.10.10
```

where the hex bit in the middle is a hash of the original string with enough entropy to avoid collisions. 

That is a potential approach/solution, but I don't have the BW atm to dig into implementing it.

davidporsche:
For what it's worth, I did some testing and crafted a manual API call for the same resource and was able to pass it directly to NS1.  The issue does not appear to be directly related to their API.  I'm having a hard time tracking down what is actually throwing this error, though I suspect maybe the NS1 SDK that octodns uses?

ross:
Can you provide a stack trace? I was just assuming it was API based on the comment above said. There's also length validation within octoDNS itself, but that's on record names and such, not on healthchecks.

davidporsche:
Here's a fresh trace

`Traceback (most recent call last):
  File "/usr/local/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/site-packages/octodns/cmds/sync.py", line 38, in main
[...]

Issue #718, opened 2021-05-19
NS1: Creating notification list only if it does not exist


meghashyamps:
Current implementation creates a notification list per record per answer as seen below:

![image](https://user-images.githubusercontent.com/13583698/118861705-a7357800-b891-11eb-9af7-b8d61f2cc3a2.png)

The notifiers in each of the notification lists are exactly the same, yet it ends up creating multiple of these lists instead of reusing a single list. 

One of the main use cases for us is having 2 or 3 main notification lists, something like so:
1. production_notification_list
2. test_notification_list
3. other_notification_list

and then reusing these 3 in relevant monitors. 

What do you think about creating a common notification list instead of per record per answer? (I can open a PR for this)

ross:
Having the search my memory, it was a long while ago now and it's not something I've thought about since so I don't remember exactly why it is the way it is currently. 

Other than having a lot of them what downsides are there to having a list per item?

It'd be a fairly big change to the code, mostly in that it would have to preserve backwards compatibility with what's been created for existing uses. It would have to seamlessly switch things over and back (in case someone upgrades and then downgrades...)

Anyway, there are a few cases where this sort of change has been made, but it's only worth it when it's absolutely required, which isn't often.

meghashyamps:
The main concern is the possibility of multiple repeated notifications of the same monitor (answer) event. Consider this:

We have many dynamic records in production, and some of them have same answers in some pools like so:
```
# excerpt from test.com.yaml

 www:
   dynamic:
     pools:
       pop1:
         values:
         - value: pop1-fqdn.test.com
```
and
```
# excerpt from www2.test.com.yaml

 www2:
   dynamic:
     pools:
       pop_rr:
         values:
         - value: pop1-fqdn.test.com
         - value: pop2-fqdn.test.com
```

Now, because of the way it is implemented currently, `pop1-fqdn.test.com` is going to be monitored by two different monitors. One as part of `www.test.com` and the other as part of `www2.test.com`. The current workflow also creates a unique `notify_list` for every monitor. So any status up/down event on `pop1-fqdn.test.com` is going to trigger notifications on two notification lists. And once https://github.com/octodns/octodns/issues/719 (I am working on it) is taken care of, it is going to trigger emails on two notification lists, there by getting 2 emails for the same endpoint status up/down event. And if an endpoint is used in N places, we are going to get N email notifications. 

Yes, this is going to be a big change. So I am going try and see if there is a way around it without having to implement this. 

Thank you

[...]

Issue #747, opened 2021-08-01
Options to fix 64 char limit for notification lists


meghashyamps:
```
www:
  dynamic:
    pools:
      eu:
        fallback: na
        values:
        - value: pop-europe-alpha.api.mydomain-bi.com.      <--- notify list name for this exceeds 64 chars
      na:
        values:
        - value: pop-beta.api.mydomain-bi.com.
        - value: pop-gamma.api.mydomain-bi.com.
    rules:
    - geos:
      - EU
      pool: eu
    - geos:
      - NA
      pool: na
    - pool: na
  octodns:
    healthcheck:
      host: null
      path: /pop/admin
      port: 443
      protocol: HTTPS
  ttl: 300
  type: CNAME
  value: pop-delta.api.mydomain-bi.com.
```

During migration to prod, we noticed that some of the answers were long and their notification list names were exceeding the 64 character limit that is set by NS1 (Octo prepends the answer with host and "CNAME" with spaces and dashes - to push it over 64 chars).  

```
ns1.rest.errors.ResourceException: server error: Input validation failed (Value u'www.sham-test2.com - CNAME - pop-europe-alpha.api.mydomain-bi.com' for field '<obj>.name' must have length less than or equal to 64)
```

First off, changing our answers to shorter FQDNs is not an option since the current FQDNs have some contextual meaning as to why they are the way they are.

Some options that could fix this:
1. Convert the names of all notification lists to its [md5sums](https://stackoverflow.com/a/5297483) before creating them, so their names will be their md5sums string
2. Convert only the ones whose name are > 64 chars to md5sum, where as the rest will follow along what is there currently
3. Have just one global notify list that **all** monitors use, since we are just creating identical notification lists anyway. Our earlier discussion touched on [this](https://github.com/octodns/octodns/issues/718) briefly. 
4. Any other option I have not thought about yet

We do not track notify lists today, nor do we parse its notes (or name) to get other info. So I believe that makes options 1 and 2 (may be even 3) both forward and backward compatible. 

Let me know which way is best, and I will work on getting a PR out.

meghashyamps:
Ah I just saw there was already an issue raised for this by @davidporsche at https://github.com/octodns/octodns/issues/744

ross:
Closing as dup of https://github.com/octodns/octodns/issues/744.

PR #749, opened 2021-08-01
Ns1 shared notifylist

ross:
This PR adds a new configuration option to `Ns1Provider` to support using a shared/single notifylist for all monitors. 

This has several benefits:
* It works around name length issues with NS1's notification list object/api
* It allows people to share a single notification list/behavior and customize it if they wish, though note that ALL records will then funnel through said list.

The setting defaults to `false` which should preserve the existing behavior.

Note that changing the setting won't cause existing records & values to be moved over to a shared notifylist. Only new/changed values will make the transition when updated.

/cc Fixes https://github.com/octodns/octodns/issues/744
/cc Fixes https://github.com/octodns/octodns/issues/718
/cc https://github.com/octodns/octodns/issues/747 which dupped ^
/cc @davidporsche @meghashyamps

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -236,8 +236,9 @@ class Ns1Provider(BaseProvider):
         # Optional. Default: false. true is Recommended, but not the default
         # for backwards compatibility reasons. If true, all NS1 monitors will
         # use a shared notify list rather than one per record & value
-        # combination. See CHANGELOG for more information before enabling this
-        # behavior.
+        # combination. See CHANGELOG,
+        # https://github.com/octodns/octodns/blob/master/CHANGELOG.md, for more
+        # information before enabling this behavior.
         shared_notifylist: false
         # Optional. Default: None. If set, back off in advance to avoid 429s
         # from rate-limiting. Generally this should be set to the number
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -913,7 +913,6 @@ class Ns1Provider(BaseProvider):
     def _feed_create(self, monitor):
         monitor_id = monitor['id']
         self.log.debug('_feed_create: monitor=%s', monitor_id)
-        # TODO: looks like length limit is 64 char
         name = '{} - {}'.format(monitor['name'], self._uuid()[:6])
 
         # Create the data feed
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -236,7 +236,8 @@ class Ns1Provider(BaseProvider):
         # Optional. Default: false. true is Recommended, but not the default
         # for backwards compatibility reasons. If true, all NS1 monitors will
         # use a shared notify list rather than one per record & value
-        # combination.
+        # combination. See CHANGELOG for more information before enabling this
+        # behavior.
         shared_notifylist: false
         # Optional. Default: None. If set, back off in advance to avoid 429s
         # from rate-limiting. Generally this should be set to the number
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -77,8 +77,10 @@ class Ns1Client(object):
         self._datafeed = client.datafeed()
 
         self._datasource_id = None
+
         self._feeds_for_monitors = None
         self._monitors_cache = None
+        self._notifylists_cache = None
 
     @property
     def datasource_id(self):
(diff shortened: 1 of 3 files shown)


meghashyamps:
> * It works around name length issues with NS1's notification list object/api
> * It allows people to share a single notification list/behavior and customize it if they wish, though note that ALL records will then funnel through said list.

Confirmed that both these are working as you described. Thank you for fixing!

octodns/provider/ns1.py
```diff
@@ -216,6 +233,11 @@ class Ns1Provider(BaseProvider):
         # Only required if using dynamic records
         monitor_regions:
           - lga
+        # Optional. Default: false. true is Recommended, but not the default
+        # for backwards compatibility reasons. If true, all NS1 monitors will
+        # use a shared notify list rather than one per record & value
+        # combination.
+        shared_notifylist: false
```
viranch, line 240: Does this need a more obvious warning about forward incompatibility? Not only in the comments/documentation but also the output of sync command?

If I create a record using the new version with shared_notifylist: true then downgrade and delete a pool/value, the old version will blindly knock out the associated notify list, breaking the other pools/records that were sharing it.

viranch:
We should probably drop the TODO about 64 char limit in `_feed_create`. My testing shows the limit is on the notifylist name and not the feed name.

octodns/provider/ns1.py
ross: > Does this need a more obvious warning about forward incompatibility? Not only in the comments/documentation but also the output of sync command?

Yes. I'd decided that `shared_notifylist: false` should at least be a warning saying that it's not the "preferred" way going forward and had been thinking over whether it should be a new required parameter to force people to decide on version update.  I think i've landed on the required parameter being too strong, but I'm open to others thoughts there.

> If I create a record using the new version with shared_notifylist: true then downgrade and delete a pool/value, the old version will blindly knock out the associated notify list, breaking the other pools/records that were sharing it.

Good call. That's a case I'd overlooked. Not coming up with great options thinking about it. Best I can come up with would be to release a version that leaves the shared ones alone on delete, but doesn't create them, but that'd only really allow updates to/from that version and you wouldn't be able to go back further so I'm not sure that buys anything.

Another option would be to declaire that once you enable `shared_notifylist` you can't go back versions.

Either way it might make sense to go ahead and add `extra_changes` bits to detect the state of notifylists and share/unshare them when they don't match the setting. That should give a controlled path to downgrade, though it could still allow someone to paint themselves into a corner if they were determined enough.

viranch: I couldn't think of any good options either. At the very least, I think we should stick a log.warn if shared_notifylist is set to true and may be hyper-link (to this PR?) for a more detailed explanation.

> add extra_changes bits to detect the state of notifylists and share/unshare them

This will want to update every single dynamic record on running the sync command after setting the config to true. But I still like the idea. If someone wanted to knowingly downgrade, they have some way. With current implementation, there's no looking back.

ross: > This will want to update every single dynamic record on running the sync command after setting the config to true. But I still like the idea. If someone wanted to knowingly downgrade, they have some way. With current implementation, there's no looking back.

Yeah. In theory it'd be possible to make it smart enough to just update the monitors, but I haven't dug into how involved/possible that'd be.

ross: Added a CHANGELOG section with the cautions and caveots and linked that to this PR. Added a mention of the CHANGELOG in the parameter doc. Lmk what you think.

octodns/provider/ns1.py
```diff
@@ -216,6 +233,12 @@ class Ns1Provider(BaseProvider):
         # Only required if using dynamic records
         monitor_regions:
           - lga
+        # Optional. Default: false. true is Recommended, but not the default
+        # for backwards compatibility reasons. If true, all NS1 monitors will
+        # use a shared notify list rather than one per record & value
+        # combination. See CHANGELOG for more information before enabling this
```
viranch, line 239: Not super critical, but it'd be nice and convenient to have a direct link to the CHANGELOG.

ross: I can link the doc, but not to the exact section or anything since it's not yet merged into a permanent location.

viranch: Yeah makes sense. I guess anyone reading these code comments knows their way to the CHANGELOG file.

viranch (approved):
🚢

PR #754, opened 2021-08-13
Multi-value PTR records

viranch:
Many providers support multi-value PTR records. octoDNS assumes all PTR records are single value. This PR adds support for multi-value PTRs in a backward compatible way so that providers that don't support it don't break.

Only Azure DNS and NS1 make use of this for now, because I only have access to those two for testing. I've tested the new multi-value as well as current single-value formats on both providers and they seem to work well. Since this is a new feature, multi-value PTR records will not be forward compatible - fortunately, octoDNS will throw a validation error after rolling back to old version.

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -7,7 +7,6 @@ from __future__ import absolute_import, division, print_function, \
 
 from six import text_type
 
-from ..record import Record
 from ..source.base import BaseSource
 from ..zone import Zone
 from .plan import Plan
@@ -45,34 +44,28 @@ class BaseProvider(BaseSource):
         '''
         return []
 
-    def _process_change(self, change):
+    def _process_desired_zone(self, desired):
         '''
-        Process/manipulate each change for feature-specific corner cases
+        Providers can use this method to make any custom changes to the
+        desired zone.
         '''
-        if not self._include_change(change):
-            return False
+        if self.SUPPORTS_MUTLIVALUE_PTR:
+            # nothing do here
+            return desired
 
-        # Multi value PTR records
-        record = change.new
-        if record and record._type == 'PTR' and len(record.values) > 1 and \
-           not self.SUPPORTS_MUTLIVALUE_PTR:
-            # replace with a single-value copy
-            change.new = Record.new(record.zone, record.name, {
-                'type': 'PTR',
-                'ttl': record.ttl,
-                'values': [record.value],
-            }, source=record.source)
+        new_desired = Zone(desired.name, desired.sub_zones)
+        for record in desired.records:
+            if record._type == 'PTR' and len(record.values) > 1:
+                # replace with a single-value copy
+                self.log.warn('does not support multi-value PTR records; '
+                              'will use only %s for %s', record.value,
+                              record.fqdn)
+                record = record.copy()
+                record.values = [record.values[0]]
 
-            existing = change.existing
-            if existing and not existing.changes(change.new, self):
-                # if new single-value replacement turns out to be the same,
-                # skip the change
-                return False
+            new_desired.add_record(record)
 
-            self.log.warn('does not support multi-value PTR records; will '
-                          'use only %s for %s', record.value, record.fqdn)
-
-        return True
+        return new_desired
 
     def plan(self, desired, processors=[]):
         self.log.info('plan: desired=%s', desired.name)
@@ -88,12 +81,15 @@ class BaseProvider(BaseSource):
         for processor in processors:
             existing = processor.process_target_zone(existing, target=self)
 
+        # process desired zone for any custom zone/record modification
+        desired = self._process_desired_zone(desired)
+
         # compute the changes at the zone/record level
         changes = existing.changes(desired, self)
 
         # allow the provider to filter out false positives
         before = len(changes)
-        changes = [c for c in changes if self._process_change(c)]
+        changes = [c for c in changes if self._include_change(c)]
         after = len(changes)
         if before != after:
             self.log.info('plan:   filtered out %s changes', before - after)
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import, division, print_function, \
 
 from six import text_type
 
+from ..record import Record
 from ..source.base import BaseSource
 from ..zone import Zone
 from .plan import Plan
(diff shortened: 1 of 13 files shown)


yzguy:
I'm curious what your use case is, in my experience there was an A record for the IP, and any other names we CNAME'd to the A record for that IP.

ross:
Similarly curious on use cases. When we've had multiple names pointing to a single IP we've always wanted a canonical name in the PTR to avoid confusing reverse lookups where the name changes from run to run or similar. Not completely opposed to this, but I'd kind of consider it to violate best practices unless there are requirements I haven't thought of.

ross:
I have vauge recollections of looking at the RFC for PTRs, 1035. https://datatracker.ietf.org/doc/html/rfc1035. I've never been good at/patient enough to really read them, but my memory & quick skim leads me towards the canonical idea.  

> Both the gateway pointers at network nodes and the normal host pointers at full address nodes use the PTR RR to point back to the primary domain names of the corresponding hosts.

>    - Gateways will often have two names in separate domains, only one of which can be primary.

That said NS1 definitely supports recording multiple values and the DNS protocols doesn't have a problem with it

```
coho:weather ross$ dig +short PTR 43.252.42.192.in-addr.arpa. @dns1.p07.nsone.net.
foo.exxampled.com.
bar.exxampled.com.
```

Do tools like traceroute/mtu just randomly pick one of the values?

viranch:
> what your use case is

Exactly what you described (multiple CNAMEs pointing to one IP) along with a bunch of legacy records that other teams (at our relatively large org) own that we can't refactor yet.

> I'll give it a try on some of the providers

I'm sure you can guess, but just want to reiterate that you'd need to change the provider classes to parse and save multiple PTR values; the PR won't work as is.

>  the DNS protocols doesn't have a problem with it

Exactly why I thought this won't necessarily a bad thing for octoDNS.

octodns/record/__init__.py
```diff
@@ -1256,13 +1256,40 @@ class NsRecord(_ValuesMixin, Record):
...
     _value_type = PtrValue
 
+    # This is for backward compatibility with providers that don't support
+    # multi-value PTR records.
+    @property
+    def value(self):
+        if len(self.values) == 1:
+            return self.data['value']
+
+        raise AttributeError("Multi-value PTR record has no attribute 'value'")
```
ross, line 1291: Will have to think about this a bit more. Not quite feeling right as it seems like the result is that providers that don't support it just blow up on apply if you put multiple values in a `PTR`, but don't indicate that anything is wrong during plan and would just show that it's going to change...

My second thought (first didn't pan out when thinking it through) is that we might have to implement something like `SUPPORTS_MULTIVALUE_PTR` and have the manager check that anytime is sees multiple values in a `PTR` during the `populate` targets phase during planning. It that works then things would just proceed normally if everything is single value or if there's multi-value PTRs and all the target support them. If there's multi-value and one or more of the targets doesn't then it'd blow up immediately with an error saying as much...

ross: As mentioned over in https://github.com/octodns/octodns/pull/755#pullrequestreview-730311028 the blow up when things aren't supported isn't really how octoDNS currently works so it probably doesn't make sense to make that sort of change now. The current behavior is more that it WARNS and continues as was done in https://github.com/octodns/octodns/pull/756.

I think that's the right tact here. I don't think we need to blow up when there's multiple values, instead we should pick the first one and use that after warning that's what is going to happen.

viranch: Ok makes sense. Let me work on this.

viranch: See if you like it now.

viranch: Hold on, this is not going to work. Sync will keep wanting to make an update since config's multi-value differs from target's single-value. Let me re-think this.

ross: May be a case where `_include_change` is useful. You can grep around for examples of what it's currently doing.

viranch: `_include_change` in base provider is my primary candidate right now. But it relies on providers calling `super(..)._include_change`. A new provider in future can easily omit the super call and sync will keep showing plans with multiple values.

viranch: I've added something around _include_change. Tested with mutli value support turned on and off, seems to work as expected.

octodns/provider/base.py
```diff
@@ -63,7 +93,7 @@ def plan(self, desired, processors=[]):
 
         # allow the provider to filter out false positives
         before = len(changes)
-        changes = [c for c in changes if self._include_change(c)]
+        changes = [c for c in changes if self._process_change(c)]
```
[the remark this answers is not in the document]
ross: Should be really easy to test too :grin:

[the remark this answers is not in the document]
ross: Payed around with a `process_desired_zone` locally in a branch and I think it will be a good place for providers validate their ability to support a `desired` state, i.e. warn as this branch will do and eventually support throwing an exception when they can't do what they're being asked to do. Will push up an example of that to a branch when I have something to look at.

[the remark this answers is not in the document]
ross: Mentioned POC around processed_desired_zone being used to warn/except when a provider can't do what it's asked to do in https://github.com/octodns/octodns/pull/757.  I think that change is related to what's being done here in terms of the path taken, but it doesn't have to block this moving forward if we don't want it to.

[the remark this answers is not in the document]
viranch: Agreed I did not like mixing `_process_change` with `_include_change`. I like your idea. The only concern I have is providers that re-implement `process_desired_zone` forgetting to put `super` call for all the PTR stuff we put in BaseProvider. We can may be implement 2 such methods, one for overriding and another for BaseProvider only. The latter can call the former. What do you think?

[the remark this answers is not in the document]
viranch: I understand that providers can override any method, but I'd like it to be as less error prone as possible.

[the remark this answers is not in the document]
ross: I have https://github.com/octodns/octodns/pull/757/files#diff-59941e8a52c6a7c89c54df502c7eee2c4ffae8bd79ab84b4a314f94acbd9520dR35-R47 which impelements the idea to see how it felt/works. Part of that documents clearly that `super` needs to be called. I think I'm OK with trusting that people will do the right thing after it's clearly doc'd and if they do it's on them...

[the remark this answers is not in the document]
viranch: Alright makes sense, I'm going to update this PR.

ross (approved):
👍

PR #755, opened 2021-08-13
adds support for CA provinces

meghashyamps:
This PR adds support for the usage of Canadian provinces as geos for NS1. 

```
www2:
  dynamic:
    pools:
      apac:
        fallback: na
        values:
        - value: abc.com.
      na:
        values:
        - value: def.com.
      rest:
        values:
          - value: ghi.com.
    rules:
    - geos:
      - NA-US-CO
      - NA-CA-NL             <---- This
      pool: apac
    - geos:
      - NA
      pool: na
    - pool: rest
  octodns:
    healthcheck:
      host: 
      path: /pop/admin
      port: 443
      protocol: HTTPS
  ttl: 300
  type: CNAME
  value: jkl.com.
```

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -622,10 +622,13 @@ class Ns1Provider(BaseProvider):
                     for c in countries:
                         geos.add('{}-{}'.format(continent, c))
 
-            # States are easy too, just assume NA-US (CA providences aren't
-            # supported by octoDNS currently)
+            # States and provinces are easy too,
+            # just assume NA-US or NA-CA
             for state in meta.get('us_state', []):
                 geos.add('NA-US-{}'.format(state))
+
+            for province in meta.get('ca_province', []):
+                geos.add('NA-CA-{}'.format(province))
 
             if geos:
                 # There are geos, combine them with any existing geos for this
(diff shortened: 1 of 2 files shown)


ross (commented):
What happens if you try and push records with `NA-CA-*` to a provider that doesn't support them, e.g. Route53?

meghashyamps:
> What happens if you try and push records with `NA-CA-*` to a provider that doesn't support them, e.g. Route53?

I did not consider that since I only tested on NS1 and AzureDNS for this. 

A quick check on route53 tells me that it fails:
```
botocore.errorfactory.InvalidChangeBatch: An error occurred (InvalidChangeBatch) when calling the ChangeResourceRecordSets operation: [Cannot find location: 'country = CA subdivision = NL'.  No changes made.]
```

What do you propose we do here?

ross:
> What do you propose we do here? I'm thinking something on similar lines as [this](https://github.com/octodns/octodns/pull/754#pullrequestreview-730035739)

Yeah. That's roughly what I'm thinking as well, but I would like to take a day or two to dig into it and try and come up with a solution that is somewhat future proof in terms of if/when we need other sorts of "supports" checking for dynamic records. I may end up throwing up a stand-alone PR that implements the checking on top of which this PR w/CA support can be added.

ross (approved):
After digging in, I've realized that CA provinces are already supported in general, just not by route53 and ns1 (until this PR.) 

Good to get NS1 supporting it since the underlying provider does. I'm going to look at adding some warnings to Route53 when it sees CA provinces in things and have it ignore them. That'd be more in line with out other things that aren't supported work, e.g. record types. 

Wouldn't be surprised if there's other details of support with providers, but we'll deal with them as they come up.

PR #756, opened 2021-08-16
Route53Provider needs to skip and warn when it sees NA-CA-* geos

ross:
Having it ignore and warn is more in line with how octoDNS handles the lack of support for other things, e.g. record types. I have and always have had mixed feelings about that behavior since if people aren't paying attention to the warnings they could miss why something their expecting to happen isn't. I may revisit that as an option at some point, but for now this turns a partial apply failure into a more consistent behavior.

/cc https://github.com/octodns/octodns/pull/755#pullrequestreview-730311028 which lead to discovering this issue

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -31,6 +31,7 @@ def _octal_replace(s):
 
 
 class _Route53Record(EqualityTupleMixin):
+    log = logging.getLogger('Route53Record')
 
     @classmethod
     def _new_dynamic(cls, provider, record, hosted_zone_id, creating):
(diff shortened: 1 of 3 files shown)


octodns/provider/route53.py
```diff
@@ -90,7 +91,21 @@ def _new_dynamic(cls, provider, record, hosted_zone_id, creating):
...
+                                 if not g.startswith('NA-CA-')]
+                if not filtered_geos:
+                    # We've removed all geos, we'll have to skip this rule
+                    cls.log.warning('NA-CA-* not supported resulting in '
+                                    'empty geo target, skipping rule %d', i)
+                    continue
+                elif geos != filtered_geos:
+                    # We've removed some geos
+                    cls.log.warning('NA-CA-* not supported, using remaining '
+                                    'geos resulting in rule %d', i)
```
viranch, line 105: Looks like this warning will be shown only apply only? If so, it may be better to show such things early on in the plan phase so the user knows in advance what they're getting into: potentially different geo-fencing between providers.

ross: Yeah. That unfortunaly is the way it currently goes. I have some thoughts/sketches on how to move these sorts of checks and warnings into the plan phase and hopefully support making them errors and not just warnings, but it'll be a fairly sizeable undertaking. It's on my todo list, but not sure when I'll get to digging into it. For now this is begrudgingly OK.)

viranch: You could move this NA-CA-* check in `_extra_changes` no? Did you mean that is not trivial?

ross: I was meaning a more general solution to warning about these sorts of problems during planning and the option to make them exceptions. Something that runs around the same point as extra changes may make sense, but I'd rather not overload what `_extra_changes` is responsible for. I don't think the place to do checks would be a bunch of work, but I suspect going through and getting all of the checks sorted will be a bit involved. At least based on the experimentation I did when thinking about this.

ross: POC of allowing providers to indicate they can't support things during the planning phase with warnings or exceptions in https://github.com/octodns/octodns/pull/757. 

If it continues to pan out that may be a better route to take than this PR since it would cover this and warn before apply. Going to hold off merging this to see how/where that goes.

ross:
Replaced by https://github.com/octodns/octodns/pull/757

PR #757, opened 2021-08-18
Provider.process_desired_zone with warning & exception support

ross:
Liking the way this shaped up. This implements a place for providers to hook into the plan process in order to modify the desired state based on what they're able to support and documents what they can do and how they need to go about it. 

As a working/testing example `Route53Provider` implements filtering of `NA-CA-*` geos adhering to the process guidlines. This is a likely replacement for https://github.com/octodns/octodns/pull/756

It's similar to discussion in https://github.com/octodns/octodns/pull/754 and providers a potential platform on which to implement multi-value PTR support on top of cleanly. 

/cc @viranch lmk what you think

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -12,6 +12,10 @@ from ..zone import Zone
 from .plan import Plan
 
 
+class ProviderException(Exception):
+    pass
+
+
 class BaseProvider(BaseSource):
 
     def __init__(self, id, apply_disabled=False,
@@ -43,6 +47,12 @@ class BaseProvider(BaseSource):
         base NS records.
         '''
         return []
+
+    def supports_warn_or_except(self, msg):
+        # TODO: base class param to control warn vs except
+        if False:
+            raise ProviderException(msg)
+        self.log.warning(msg)
 
     def process_desired_zone(self, desired):
         return desired
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -52,10 +52,11 @@ class BaseProvider(BaseSource):
         for record in desired.records:
             if record._type == 'PTR' and len(record.values) > 1:
                 # replace with a single-value copy
-                self.supports_warn_or_except('does not support multi-value '
-                                             'PTR records; will use only {} '
-                                             'for {}'.format(record.value,
-                                                             record.fqdn))
+                msg = 'multi-value PTR records not supported for {}' \
+                    .format(record.fqdn)
+                fallback = 'falling back to single value, {}' \
+                    .format(record.value)
+                self.supports_warn_or_except(msg, fallback)
                 record = record.copy()
                 record.values = [record.value]
 
@@ -78,10 +79,10 @@ class BaseProvider(BaseSource):
         '''
         return []
 
-    def supports_warn_or_except(self, msg):
+    def supports_warn_or_except(self, msg, fallback):
         if self.strict_supports:
             raise ProviderException('{}: {}'.format(self.id, msg))
-        self.log.warning(msg)
+        self.log.warning('{}; {}'.format(msg, fallback))
 
     def plan(self, desired, processors=[]):
         self.log.info('plan: desired=%s', desired.name)
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -44,8 +44,13 @@ class BaseProvider(BaseSource):
         '''
         return []
 
+    def process_desired_zone(self, desired):
+        return desired
+
     def plan(self, desired, processors=[]):
         self.log.info('plan: desired=%s', desired.name)
+
+        desired = self.process_desired_zone(desired)
 
         existing = Zone(desired.name, desired.sub_zones)
         exists = self.populate(existing, target=True, lenient=True)
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -10,17 +10,15 @@ from six import text_type
 from ..source.base import BaseSource
 from ..zone import Zone
 from .plan import Plan
-
-
-class ProviderException(Exception):
-    pass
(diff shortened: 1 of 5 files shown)


octodns/provider/base.py
```diff
@@ -57,19 +52,43 @@ def _process_desired_zone(self, desired):
         for record in desired.records:
             if record._type == 'PTR' and len(record.values) > 1:
                 # replace with a single-value copy
-                self.log.warn('does not support multi-value PTR records; '
-                              'will use only %s for %s', record.value,
-                              record.fqdn)
+                self.supports_warn_or_except('does not support multi-value '
+                                             'PTR records; will use only {} '
+                                             'for {}'.format(record.value,
+                                                             record.fqdn))
```
viranch, line 58: NBD, but this message can be confusing to show in the exception. Same for Route53's NA-CA-* message.

octodns/provider/base.py
```diff
@@ -28,26 +30,19 @@ def __init__(self, id, apply_disabled=False,
...
     def _process_desired_zone(self, desired):
         '''
-        Providers can use this method to make any custom changes to the
-        desired zone.
+        An opportunity for providers to modify that desired zone records before
+        planning.
+
+        - Must do their work and then call super with the results of that work
+        - Must not modify the `desired` parameter or its records and should
+          make a copy of anything it's modifying
```
viranch, line 42: Should we send a copy of `desired` to begin with?

octodns/provider/base.py
```diff
@@ -28,26 +30,19 @@ def __init__(self, id, apply_disabled=False,
...
-        return True
-
-    def _extra_changes(self, existing, desired, changes):
-        '''
-        An opportunity for providers to add extra changes to the plan that are
-        necessary to update ancillary record data or configure the zone. E.g.
-        base NS records.
-        '''
-        return []
+        self.strict_supports = strict_supports
```
viranch, line 33: Sounds like it implies that we should check for all `SUPPORTS_*` features in `_process_desired_zone` and not just PTR one? May be good to consolidate all of them in one method.

octodns/provider/base.py
ross: Went back and forth on that. We'd have to copy the zone and all the records to make it safe. Decided that was a bunch of overhead for things that 90-whatever percent of the time won't actually make copies, but I could be convinced.

viranch: We're slowly increasing programability available to the providers. That also starts increasing the chances of providers doing unwanted things and potentially 🦶 🔫 . I guess it's a paradigm choice to make on our part. safety vs simplicity/performance (I am guessing large zones will have non trivial overhead of copying per provider). I don't have a strong opinion either way.

A good balance could be to create a copy only if a provider re-implements the function. My Python OOP is not good enough to recall if this is possible.

octodns/provider/base.py
ross: Yeah. That's what I was meaning when I said the larger work here would be sizeable. There'd likely even be some simplification/refactoring if we remove `dynamic` (and `geo`) from the desired during this point of things. 

That's probably too big of a change for a single PR so my thinking would be to ship this once we're happy with it and iterate towards other things happening here that are already happening elsewhere over time/PRs.

octodns/provider/base.py
viranch: I'm okay keeping it as-is too and letting providers re-implement it properly.

ross: Been thinking about this a bit more. I think we could potentially come up with a setup that does ~copy-on-write. 

I.e. some sort of `Zone.shallow_copy` that could be passed in at fairly low cost, e.g. the `Zone` object and it's set would be duplicated, but all the `Record` objects would be shared. Could then add `Zone.replace_record` that would take the modified record and replace it in the shallow copied zone. There may be other options for replacing with less work on the part of the provider.

That said it would still have the overhead of duplicating and coping everything into the `Zone` for 99% of cases where nothing was actually going to change... For smaller/common zones it's probably NBD, but I'm fairly sure there are people managing zones with many 1000's of records and the solution doesn't seem in balance with the (lack of) problem so as interesting as that'd be I'd probably still land with things behaving as they currently do. 

It might be worth exploring the copy-on-write support in general for the providers to use when they do modify things. It might make it easier for them to do that work. Will see how that goes as we move more things into this paradigm. Definitely not a change for this PR.

octodns/provider/base.py
```diff
@@ -57,19 +52,44 @@ def _process_desired_zone(self, desired):
...
+
+    def _extra_changes(self, existing, desired, changes):
+        '''
+        An opportunity for providers to add extra changes to the plan that are
+        necessary to update ancillary record data or configure the zone. E.g.
+        base NS records.
+        '''
+        return []
+
+    def supports_warn_or_except(self, msg, fallback):
```
ross, line 82: I've split the message param into two parts. The first always applies for both exceptions and log warnings as it's the "problem". The `fallback` bit only gets used in log warnings when we've worked around the lack of support. I think that will make things less confusing.

viranch: This is exactly what I was thinking, but thought it might be an unnecessary complexity. But looks great now 👍

ross:
```
2021-08-21T02:51:46  [4565028352] WARNING Route53Provider[route53] NA-CA-* not supported for exxampled.com.; filtering rule 0 from (EU, NA-CA-BC) to (EU)
2021-08-21T02:51:46  [4565028352] WARNING Route53Provider[route53] NA-CA-* not supported for exxampled.com.; skipping rule 1
```

With `strict_supports` on the provider:

```
octodns.provider.ProviderException: route53: NA-CA-* not supported for exxampled.com.
```

This appears to work as expected.

PR #758, opened 2021-08-19
compare ipv6 addresses properly

meghashyamps:
Octo thinks there is a diff when it sees the same IPv6 address in short vs long form:

**sham-test1.com.yaml**
```
---
'':
  - type: AAAA
    ttl: 86400
    values:
    - 2620:109:c002::6cae:a7e

```

**NS1:**
image

**Octo dry run**
```
* sham-test1.com.
* ns1 (Ns1Provider)
*   Update
*     <AaaaRecord AAAA 86400, sham-test1.com., ['2620:0109:c002:0000:0000:0000:6cae:0a7e']> ->
*     <AaaaRecord AAAA 86400, sham-test1.com., ['2620:109:c002::6cae:a7e']> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=2
```

The PR compares IPv6 addresses as IPv6address objects instead of string compare.

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -304,7 +304,14 @@ class _ValuesMixin(object):
         self.values = sorted(self._value_type.process(values))
 
     def changes(self, other, target):
-        if self.values != other.values:
+        self_values = self.values
+        other_values = other.values
+        # compare IPv6 addresses separately since they could be
+        # in short/long format
+        if self._type == 'AAAA':
+            self_values = [IPv6Address(x) for x in self_values]
+            other_values = [IPv6Address(x) for x in other_values]
+        if self_values != other_values:
             return Update(self, other)
         return super(_ValuesMixin, self).changes(other, target)
 


octodns/record/__init__.py
```diff
@@ -304,7 +304,14 @@ def __init__(self, zone, name, data, source=None):
...
     def changes(self, other, target):
-        if self.values != other.values:
+        self_values = self.values
+        other_values = other.values
+        # compare IPv6 addresses separately since they could be
+        # in short/long format
+        if self._type == 'AAAA':
+            self_values = [IPv6Address(x) for x in self_values]
+            other_values = [IPv6Address(x) for x in other_values]
+        if self_values != other_values:
```
ross, line 314: Good catch!

Looks like we already convert things to `IPv6Address`/`IPv4Address` as part of validation. Shame we didn't go ahead and make the values IP types. 

This solution will work, but I'll probably poke around a bit and see if I can come up with a way that avoids having `_ValuesMixin` know about the details of some of it's potential types.

ross: After beating my head on this for a while I realized the most fullproof solution was to normalize the representation using IPv[46]Address during the record create. 

Looking back I wish I'd have the values be the actual types and not just strings, but that'd be too big of a shift now...

Normalization PR is https://github.com/octodns/octodns/pull/759. Thanks for reporting this one.

ross:
Thanks for the PR. Digging into this lead to https://github.com/octodns/octodns/pull/759 which should have addressed the underlying issue and clean up IP address handling across a handful of providers. Lmk if you run into any further issues.

PR #759, opened 2021-08-20
Normalize IP addresses

ross:
This prevents issues with providers which normalize IP addresses resulting in false changes. All records from all sources/providers will come through this so it won't matter what's in yaml config files or what comes back from provider apis, they'll all be normalized by the same code. 

By normalize I mean `2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b` -> `2001:db8:3c4d:15::1a2f:1a2b`, zero removal and `::` where appropriate. 

/cc Replaces https://github.com/octodns/octodns/pull/758 
/cc @meghashyamps

### DIFF
--- a/octodns/provider/constellix.py
+++ b/octodns/provider/constellix.py
@@ -8,7 +8,6 @@ from __future__ import absolute_import, division, print_function, \
 from collections import defaultdict
 from requests import Session
 from base64 import b64encode
-from ipaddress import ip_address
 from six import string_types
 import hashlib
 import hmac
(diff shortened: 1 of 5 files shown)


yzguy:
Think this will let me remove this bit: https://github.com/octodns/octodns/blob/master/octodns/provider/constellix.py#L141-L144

ross:
> Think this will let me remove this bit: https://github.com/octodns/octodns/blob/master/octodns/provider/constellix.py#L141-L144

👍 feel free to push that to this PR if so.

yzguy (approved):

ross:
Made a pass through the other uses of `ipaddress` bits to make sure they were still necessary. The `Route53Provider` one was (improved the comment there) due to their non-standard normalization. `UltraProvider`'s use is no longer needed. This should be gtg.

PR #760, opened 2021-08-20
Cache NS1 zones and records for faster re-retrival

viranch:
`NS1Provider` retrieves every zone and dynamic record more than once. This makes octoDNS slow down for zones with large number of dynamic records.

Adding zone and record caches to `NS1Client` resolves this slow down. API calls will speed up further when https://github.com/ns1/ns1-python/pull/77 gets merged and we bump ns1-python version.

Will add tests if implementation looks good.

### DIFF
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -240,7 +240,6 @@ class Ns1Client(object):
     def zones_retrieve(self, name):
         if name not in self._zones_cache:
             self._zones_cache[name] = self._try(self._zones.retrieve, name)
-            print(f'insert {name} to cache with val {self._zones_cache[name]}')
         return self._zones_cache[name]
 
     def _try(self, method, *args, **kwargs):
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -96,6 +96,7 @@ class Ns1Client(object):
                 # remove record's zone from cache
                 del self._zones_cache[zone]
 
+            # write to (or delete) record cache
             cached = self._records_cache.setdefault(zone, {}) \
                 .setdefault(domain, {})
             new_record = func(self, zone, domain, _type, **params)
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -205,10 +205,11 @@ class Ns1Client(object):
 
     def records_delete(self, zone, domain, _type):
         try:
-            # remove record from cache
-            del self._records_cache[zone][domain][_type]
             # remove record's zone from cache
             del self._zones_cache[zone]
+            # remove record from cache, after zone since we may not have
+            # fetched the record details
+            del self._records_cache[zone][domain][_type]
         except KeyError:
             # never mind if record is not found in cache
             pass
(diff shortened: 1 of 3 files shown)


octodns/provider/ns1.py
```diff
@@ -193,22 +197,50 @@ def notifylists_list(self):
...
+        cached[_type] = self._try(self._records.create, zone, domain, _type,
+                                  **params)
+        return cached[_type]
 
     def records_delete(self, zone, domain, _type):
+        try:
+            # remove record from cache
+            del self._records_cache[zone][domain][_type]
+            # remove record's zone from cache
+            del self._zones_cache[zone]
```
ross, line 211: @viranch any specific reasons behind clearing the zone when records are deleted or updated?

ross (commented):
Went through and added coverage as well as caching on creates to this. Question to get discussed inline before 🚢

octodns/provider/ns1.py
ross: Ah. I'd forgotten that basic record info was in the zone and `records_retrieve` is only called when the record is `tier > 1`. 

I think that means we need to make sure to "clear" the zone when deleting it even when there's nothing in `records_cache` since most things won't be cached there (only `dynamic` stuff.)

viranch: Yeah zones contain the records also. Good catch clearing zone cache first.

octodns/provider/ns1.py
```diff
@@ -193,22 +197,56 @@ def notifylists_list(self):
...
+            .setdefault(domain, {})
+        try:
+            # remove record's zone from cache
+            del self._zones_cache[zone]
+            # remove record from cache, after zone since we may not have
+            # fetched the record details
+            del cached[_type]
+        except KeyError:
+            # never mind if record is not found in cache
+            pass
```
viranch, line 237: I feel like it might better to wrap each `del` in its own try-except in all methods. Current block assumes that if zone cache doesn't exist, record cache won't exist either. But `records_update` clears zone cache then re-populates record cache which invalidates the assumption. It is practically unlikely we'll run into this now, but I'd rather make it harder to make mistakes in cache management for future code changes. It'll probably be useful to have a clear_cache helper method. Let me take a stab at it.

viranch: Decided to go a little fancy with decorators for a more centralized cache management.

ross: Came to the same realization this morning. Subsequent deletes of records in the same zone wouldn't work as the zone wouldn't exist and it wouldn't make it to the record. So either one first doesn't work. So definitely have to delete both always.

octodns/provider/ns1.py
```diff
@@ -80,11 +80,48 @@ def __init__(self, api_key, parallelism=None, retry_count=4,
...
+    def read_or_set_record_cache(func):
+        def call(self, zone, domain, _type):
+            cached = self._records_cache.setdefault(zone, {}) \
+                .setdefault(domain, {})
+            if _type not in cached:
+                cached[_type] = func(self, zone, domain, _type)
+
+            return cached[_type]
+
+        return call
```
ross, line 124: 👍 cleans up the actual methods.

ross (approved):

viranch:
Reporting some data, on a test zone with 35 dynamic records:

```
 $ grep -c dynamic src/dev/test/exampled.com.yaml
35

$ # before
$ time octodns-sync --config-file=./config/test.yaml exampled.com. --target ns1
...
3.21s user 0.15s system 11% cpu 28.937 total

$ # after
$ time octodns-sync --config-file=./config/test.yaml exampled.com. --target ns1
...
3.02s user 0.12s system 27% cpu 11.383 total
```

Down from ~30s to ~12s 💪

PR #761, opened 2021-08-21
Zone shallow copy support and usage

ross:
This PR implements `Zone.copy` which does "shallow" copies of `Zone`s to allow situations that are modifying zone contents to more cleanly/safely do so. This is needed in situations with VERY large zones and infrequent/unlikely modifications.

Things that utilize the functionality:
* `Processor.*` - by their nature they (potentially) modify the makeup of a zone - considered alpha so the "breaking" changes should be fair game
* `Provider._process_desired_zone` - infrequently makes changes to zones based on the providers ability to support what it's being asked to do - as of yet unreleased so fair game

/cc https://github.com/octodns/octodns/pull/757#discussion_r691676030 for discussion that lead to this path/solution
/cc @viranch @yzguy fyi

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -10,7 +10,7 @@ from six import text_type
 from ..source.base import BaseSource
 from ..zone import Zone
 from .plan import Plan
-from . import ProviderException
+from . import SupportsException
 
 
 class BaseProvider(BaseSource):
@@ -85,7 +85,7 @@ class BaseProvider(BaseSource):
 
     def supports_warn_or_except(self, msg, fallback):
         if self.strict_supports:
-            raise ProviderException('{}: {}'.format(self.id, msg))
+            raise SupportsException('{}: {}'.format(self.id, msg))
         self.log.warning('{}; {}'.format(msg, fallback))
 
     def plan(self, desired, processors=[]):
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -34,21 +34,29 @@ class BaseProvider(BaseSource):
 
     def _process_desired_zone(self, desired):
         '''
-        An opportunity for providers to modify that desired zone records before
-        planning.
+        An opportunity for providers to modify the desired zone records before
+        planning. `desired` is a "shallow" copy, see `Zone.copy` for more
+        information
 
-        - Must do their work and then call super with the results of that work
-        - Must not modify the `desired` parameter or its records and should
-          make a copy of anything it's modifying
+        - Must do their work and then call `super` with the results of that
+          work, returning the result of the `super` call.
+        - Must not modify `desired` directly, should call `desired.copy` and
+          modify the shallow copy returned from that.
+        - Must not modify records directly, `record.copy` should be called,
+          the results of which can be modified, and then `Zone.add_record` may
+          be used with `replace=True`
+        - Must call `Zone.remove_record` to remove records from the copy of
+          `desired`
         - Must call supports_warn_or_except with information about any changes
           that are made to have them logged or throw errors depending on the
-          configuration
+          provider configuration
         '''
         if self.SUPPORTS_MUTLIVALUE_PTR:
             # nothing do here
             return desired
 
-        new_desired = Zone(desired.name, desired.sub_zones)
+        # Shallow copy
+        new_desired = desired.copy()
         for record in desired.records:
             if record._type == 'PTR' and len(record.values) > 1:
                 # replace with a single-value copy
@@ -59,8 +67,7 @@ class BaseProvider(BaseSource):
                 self.supports_warn_or_except(msg, fallback)
                 record = record.copy()
                 record.values = [record.value]
-
-            new_desired.add_record(record)
+                new_desired.add_record(record, replace=True)
 
         return new_desired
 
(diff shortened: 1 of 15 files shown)


octodns/provider/base.py
```diff
@@ -34,21 +34,29 @@ def __init__(self, id, apply_disabled=False,
...
-        - Must not modify the `desired` parameter or its records and should
-          make a copy of anything it's modifying
+        An opportunity for providers to modify the desired zone records before
+        planning. `desired` is a "shallow" copy, see `Zone.copy` for more
+        information
+
+        - Must do their work and then call `super` with the results of that
+          work, returning the result of the `super` call.
+        - Must not modify `desired` directly, should call `desired.copy` and
+          modify the shallow copy returned from that.
```
ross, line 44: This is the main bit I've gone back and forth on. Part of me wants to call `_process_desired_zone` with a copy in the first place so that callers don't have to worry about it. The slightly funky part of that is things are then modifying a method parameter and we would probably want to get rid of the return value since it'd just be what was passed in. 

The pro would be that implementations don't have to make the copy and when the method is overridden there won't be multiple shallow copies made. The cons are that it would always be making a single copy and the aforementioned modifying a parameter thing.

Same things would be true for processors, and we should probably do the same there as here, but in that case there could be a lot of (shallow) copies made, one for each processor.

ross: Thought more about this and decided that it is best to create a single copy early on and pass it through everything after that allowing them to modify. As implmented in what's being pushed now that applies to both `Provider._process_source_zone` and the processors. Also added a bunch of documentation to all of these methods about the dos and don'ts.

ross:
Just played around with this a bit more locally and implemented `SUPPORTS` and `SUPPORTS_DYNAMIC` checking/handling and I'm really liking how this works. It's good to have explicit/consistent warns when things aren't supported and the option of making them errors so that you don't accidentally miss something you expect to be there. 

Assuming everyone is on board with this approach. We can work through this PR and then I'll submit a couple/few others soon after to implement other supports bits.

I'm actually leaning towards making `strict_supports=True` by default as it's the safer option and easy enough for anyone who prefers warnings to get them. It's probably best to get everything implemented as warnings before making that change. It's definitely something that should be front and center in the CHANGELOG if it happens.

octodns/provider/base.py
```diff
@@ -34,21 +34,26 @@ def __init__(self, id, apply_disabled=False,
...
-          make a copy of anything it's modifying
+        An opportunity for providers to modify the desired zone records before
+        planning. `desired` is a "shallow" copy, see `Zone.copy` for more
+        information
+
+        - Must call `super` at an appropriate point for their work, generally
+          that means as the final step of the method, returning the result of
+          the `super` call.
+        - May modify `desired` directly.
+        - Must not modify records directly, `record.copy` should be called,
```
viranch, line 45: Do you think it might be worth it to make record objects "immutable" during this phase so that trying to set an attribute would result in an exception?

octodns/provider/route53.py
```diff
@@ -957,11 +954,12 @@ def _process_desired_zone(self, desired):
                         rule.data['geos'] = filtered_geos
                     rules.append(rule)
 
-                dynamic.rules = rules
+                if rules != dynamic.rules:
+                    record = record.copy()
+                    record.dynamic.rules = rules
+                    desired.add_record(record, replace=True)
```
viranch, line 960: It might nice to provide a helper for this. Something like:
```
desired.replace_record(record, rules=new_rules)
```
which will internally make a copy of the passed record with overridden attributes.

Or probably provide a more convenient `record.copy`:
```
record = record.copy(rules=new_rules)
```
Something to think about for another PR.

viranch:
Really interesting copy-on-write implementation with the `hydrate` stuff!

re strict_supports, I'm wondering if users may get easily confused between the `lenient` and the `strict_supports` flags and the fine differences between them. Just something to keep in mind for a future PR around strict_supports=True.

octodns/provider/base.py
ross: That could be interesting. Will have to think about it/look into it.

octodns/provider/route53.py
ross: I think in order to be generic and allow it to work with arbitrary records it'd look something like:

```python
record = record.copy(data={
    'dynamic': {
        'rules': new_rules,
    }
})
```

Basically the implementation (pseudo code) of it would look something like:

```
def copy(self, data):
    merged = self.data.deep_merging_update(data)
    return Record.new(self.zone, self.name, merged)
```

octodns/provider/base.py
ross: Changes required to do this were fairly involved. Bascially everything has to be accessed through `@property` w/o setters. Still like the idea of it, but definitely out of scope for now.

octodns/provider/route53.py
ross: Going to keep this as a todo. I think it'd be handy.

ross:
> re strict_supports, I'm wondering if users may get easily confused between the `lenient` and the `strict_supports` flags and the fine differences between them. Just something to keep in mind for a future PR around strict_supports=True.

Tried to document them both in the README to avoid some confusion, for those who look/read anyway.

PR #767, opened 2021-09-01
Bump ns1-python

viranch:
The new ns1-python release includes optimizations with connection pooling so that every API call re-uses the TCP connection from previous calls if any.

Some timings from my very unscientific local testing on a zone with ~35 dynamic records (which comprise of the majority of API calls):

Before:
```
$ time octodns-sync --config-file=./config/test.yaml exampled.com. --target ns1
...
2.88s user 0.12s system 16% cpu 18.421 total
```
After:
```
$ time octodns-sync --config-file=./config/test.yaml exampled.com. --target ns1
...
2.62s user 0.09s system 34% cpu 7.800 total
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ ipaddress==1.0.23; python_version < '3.3'
 jmespath==0.10.0
 msrestazure==0.6.4
 natsort==6.2.1
-ns1-python==0.16.0
+ns1-python==0.16.1
 ovh==0.5.0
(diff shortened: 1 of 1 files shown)


yzguy:
Relevant change in [ns1-python](https://github.com/ns1/ns1-python/commit/189b2a0c50b5b1e51b728d8a05fbaae217a9d19f). Nothing crazy, just using (request's session object](https://docs.python-requests.org/en/master/user/advanced/).

Issue #711, opened 2021-05-13
Move to python's f-strings


viranch:
With this project fully on python3 now, are there any plans to replace all usages of `str.format` to f-strings? It's much easier to use among numerous other advantages listed in https://www.python.org/dev/peps/pep-0498/.

Let me know your thoughts and I will be happy to contribute.

viranch:
I noticed that CI runs only 3.x checks, so I assumed that octo officially dropped support for python 2. Did I read #677 correctly that 2.x support will completely end on 2021-08-16?

ross:
> that 2.x support will completely end on 2021-08-16?

Yeah. I'd actually forgot that I picked a date :grin: 

Until then it's on a YMMV/no promises support to give a couple bit stragglers time to update their setup...

ross:
Fixed by https://github.com/octodns/octodns/pull/768

PR #768, opened 2021-09-02
Python 3: use f-strings

ross:
* [x] Convert all the sources & provider to use f-strings, they're usually a bit shorter/cleaner and are definitely more readable in terms of what goes where in the strings.
* [x] Convert the Manager & Zone
* Not touching Constellix or Transip as they both have large active PRs that will either make the changes or they'll be made afterwards 
* Leaving tests for later as I've had enough f-string conversion for now

/cc Related https://github.com/octodns/octodns/issues/711 @viranch


ross:
All the providers have been updated. Will leave this around for a week or for anyone who wonders by to :eye:, ideally provider authors would look at their own, but I don't plan to CC/assign everyone who's worked on a provider or anything. I've been careful when making the changes and a 2nd pass of being careful when `git add -p`ing them. I won't say there are no problems or anything, but it should be pretty solid.

viranch:
Eyeballed manager.py, azuredns.py (thanks for fixing log.debugs!) and ns1.py. LGTM.

Thank you for the rather laborious work!

ross:
See https://github.com/octodns/octodns/pull/777 for why CI is currently failing.

ross:
> * Not touching Constellix or Transip as they both have large active PRs that will either make the changes or they'll be made afterwards
> * Leaving tests for later as I've had enough f-string conversion for now

Constellix is done, transip is temporarily removed. Tests have been f-stringed.

PR #769, opened 2021-09-02
Test improvements: no network access allowed & speed ups

ross:
…overage

Was looking into some slow tests and discovered that they're making network calls. Looked around for the best way to disable network access completely during tests and found a couple old broken nose plugins that do so in non-trivial ways. I created https://github.com/ross/nose-no-network which is dirt simple and just breaks `socket.socket`. That plugin is now installed and will be used by `./script/test` & `./script/coverage`.

* [x] Fix existing tests that are making network calls
* [x] Use `./script/test --with-timer` to see if there are any remaining slow tests, 1s+, and look at speeding them up if possible

### DIFF
--- a/octodns/provider/easydns.py
+++ b/octodns/provider/easydns.py
@@ -48,19 +48,19 @@ class EasyDNSClient(object):
     # Domain Portfolio
     domain_portfolio = 'myport'
 
-    def __init__(self, token, api_key, currency, portfolio, sandbox):
+    def __init__(self, token, api_key, currency, portfolio, sandbox,
+                 domain_create_sleep):
         self.log = logging.getLogger('EasyDNSProvider[{}]'.format(id))
-        self.token = token
-        self.api_key = api_key
         self.default_currency = currency
         self.domain_portfolio = portfolio
-        self.apienv = 'sandbox' if sandbox else 'live'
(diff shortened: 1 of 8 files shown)


ross:
> * Fix existing tests that are making network calls

* transip's `DomainService` (client) was making calls to get the service wsdl in the constructore before it was otherwise mocked for testing, reworked things to avoid creating the real object during tests
* Route53's credentials stuff apparently makes calls to the AWS metadata api `http://169.254.169.254/latest/meta-data/` even with stubbed, manually patched things to avoid that




> * Use `./script/test --with-timer` to see if there are any remaining slow tests, 1s+, and look at speeding them up if possible

`EasyDnsClient` has a sleep in `domain_create` which kind of stinks in general, but if it's required... Added a parameter `domain_create_sleep` which can be set in the provider config in case anyone has problems with 1s not being long enough. 
 More relevant to this PR it can be set during tests to avoid sleeping for now reason when everything is mocked.

This should be good to go and we should no longer make network calls (cuts off north of 5s) nor sleeping (cuts off 2s). Not a big deal unless you're constantly running the tests like i'm doing for https://github.com/octodns/octodns/pull/768. Things definitely feel speedier now.

PR #770, opened 2021-09-07
ProviderBase._process_desired_zone `strict_supports` checking of types and dynamic

ross:
Fairly simple PR that implements `_type` checking and `dynamic` supports in `ProviderBase._process_desired_zone` so that we get the new `supports_warn_or_except` behavior when a provider is asked to create records it doesn't support.

This is the big one in terms of getting the most common case for all providers to warn or error when they can't create records they're being asked to.

There will likely be some more detail PRs, but this gets the 90% case.

/cc @viranch @yzguy


viranch (approved):
Love it! Really elegant way to implement a common SUPPORTS logic.

PR #771, opened 2021-09-08
Record.copy with data param using deep_merge

ross:
This allows copies of records with modifications in a single call. The old way of mucking with them after the fact would still work, but this way seems cleaner. Went ahead and converted existing `Record.copy`s to use it.

/cc https://github.com/octodns/octodns/pull/761#discussion_r694323825 for discussion that lead to this implementation

### DIFF
--- a/octodns/processor/acme.py
+++ b/octodns/processor/acme.py
@@ -38,9 +38,9 @@ class AcmeMangingProcessor(BaseProcessor):
                record.name.startswith('_acme-challenge'):
                 # We have a managed acme challenge record (owned by octoDNS) so
                 # we should mark it as such
-                record = record.copy()
-                record.values.append('*octoDNS*')
-                record.values.sort()
+                record = record.copy(data={
+                    'values': sorted(record.values + ['*octoDNS*']),
+                })
                 # This assumes we'll see things as sources before targets,
                 # which is the case...
                 self._owned.add(record)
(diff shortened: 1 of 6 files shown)


ross:
This is also not working super great when you want to do something like change a single key of a value type, e.g. I want to modify the `flags` field on the first value of this CAA record and leave everything else alone. 

```python
        a_values = [{
            'flags': 0,
            'tag': 'issue',
            'value': 'ca.example.net',
        }, {
            'flags': 128,
            'tag': 'iodef',
            'value': 'mailto:security@example.com',
        }]
```

There's no real way to do that. The following kind of "feels" right, but it would replace two values with one and the other fields are overwritten. 

```python
        other = a.copy(data={
            'values': [{
                'flags': 128,
            }],
        })
```

I'm finding this stuff while looking at making the record objects immutable. That isn't as bad as I thought it'd be, but hitting these copy w/data hiccups make this feel ugly.

ross:
Haven't really found clean ways of going about this. I like the idea of immutable record objects, but without less messy way to make small changes I don't think it's a viable option 😦 

Going to abandon this as a result. See https://github.com/octodns/octodns/pull/774 for work towards making records immutable. That was going well, but the lack of a clean way to make modifications doesn't make it viable. 

I could imagine a different tact where records were marked as immutable, but otherwise editable, but that'd look very different and be a little ugly to implement. Not going to dive directly into that, but maybe I'll poke at it someday.

Closing this.

PR #779, opened 2021-09-17
Remove explicit use of six

ross:
We're python3 only now and I believe `six` is the last thing on to the TODO list of moving away from things that are explicitly there to support python2.

/cc Follow on from https://github.com/octodns/octodns/issues/711 which removed f-strings

### DIFF
--- a/octodns/cmds/report.py
+++ b/octodns/cmds/report.py
@@ -13,8 +13,6 @@ from logging import getLogger
 from sys import stdout
 import re
 
-from six import text_type
-
 from octodns.cmds.args import ArgumentParser
 from octodns.manager import Manager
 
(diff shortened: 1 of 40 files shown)


viranch (approved):
Straightforward enough, ship it.

yzguy (approved):

yzguy:
Agreed, straight forward. Also gives me a solution to some 2.7 code I'm slowly working on moving to Python 3 and dealing with the issue this was implemented for.

ross:
> Agreed, straight forward. Also gives me a solution to some 2.7 code I'm slowly working on moving to Python 3 and dealing with the issue this was implemented for.

The other things to watch out for with 2 to 3 conversion is that `filter`, `map`, etc. no longer return lists and now return generators. There were a number of hidenbugs in octoDNS or related stuff I was porting around the same time due to that.

Issue #422, opened 2019-11-15
Option to disable auto-created health checks for dynamic records


danhanks:
I have a use case where I'd like to use dynamic records, but for which I don't want the healthchecks automatically created. Would it be possible to add an option to the config to disable the auto-creation of the health checks?

ross:
> I have a use case where I'd like to use dynamic records, but for which I don't want the healthchecks automatically created.

Can you elaborate on your use case just to get it recorded down so that if/when someone goes to look at this they have an example to keep in mind?

It should be possible to do. There are some existing bits to customize health checks documented in https://github.com/github/octodns/blob/master/docs/dynamic_records.md#health-checks. Adding something like `disabled: true` should be doable, though off the top of my head I don't know how much that would change/impact the dynamic providers. 

This is a big enough task and not in our current usage of things that I probably won't get to it soon. Would welcome PRs.

bazd:
I would also like to disable the health checks for dynamic records, as I have dns records that resolve to private addresses and these are not supported for health checks in Route 53

ross:
> There are cases when I simply want to CNAME to another dynamic record which does its own health-checking and failover and I wouldn't want to health-check that dynamic record itself. I may also want to CNAME to a 3rd party load balancer (for eg [Citrix ITM](https://www.citrix.com/products/citrix-intelligent-traffic-management/)) which also has its own mechanism for health-checking and failover.

I've done this pretty frequently, but in those cases it's a simple CNAME pointed at a dynamic A. I wouldn't want 2 levels of targeting involved as it'd seem to get really complicated where traffic was directed. What's the setup leading to wanting 2 levels of targeting?

> (PS: This [current issue](https://github.com/octodns/octodns/issues/422) is for disabling health-check record wide, my use-case is per-pool, commenting here since it is a very similar request)

In general not opposed to optional health checking in general at the pool level, thought it's not something I've needed and thus haven't implemented. It'd be a bit of work to figure out the schema on the records/pools themselves and would require some research for each of the various dynamic providers to figure out the implementation details.

meghashyamps:
Below are 2 illustration of why we need the option to have a monitor as "always up". 

1. (This is the nesting or 2 levels of targeting) The answer/answers are themselves dynamic records which are self managed. `as-rr` in the below example:
```
sample from test.com.yaml zone file
---
main:
  dynamic:
    pools:
      as-rr:
        values:
        - value: as-rr.test.com
      eu-rr:
        values:
[...]

PR #780, opened 2021-09-17
Force enable/disable pool values

viranch:
Wanted to take a stab at this for #422. Will add validation & tests if implementation looks good.

/cc @danhanks @bazd @meghashyamps

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -914,8 +914,8 @@ class AzureProvider(BaseProvider):
                 for value in pool.data['values']:
                     if value['up']:
                         # Azure only supports up=None & up=False, not up=True
-                        msg = 'up=True is not supported for "{}" pool in {}' \
-                            .format(name, record.fqdn)
+                        msg = f'up=True is not supported for "{name}" pool ' \
+                            f'in {record.fqdn}'
                         fallback = 'ignoring it'
                         self.supports_warn_or_except(msg, fallback)
 
@@ -1064,7 +1064,7 @@ class AzureProvider(BaseProvider):
                 # mark default
                 ep_name += '--default--'
                 default_seen = True
-            ep_status = 'Disabled' if val['up'] == False else \
+            ep_status = 'Disabled' if val['up'] is False else \
                 'Enabled'
             endpoints.append(Endpoint(
                 name=ep_name,
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -917,7 +917,8 @@ class AzureProvider(BaseProvider):
                         # Azure only supports up=None & up=False, not up=True
                         msg = f'up=True is not supported for "{name}" pool ' \
                             f'in {record.fqdn}'
-                        fallback = 'ignoring it'
+                        fallback = \
+                            'will ignore the flag and respect healthcheck'
                         self.supports_warn_or_except(msg, fallback)
 
         return super()._process_desired_zone(desired)
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -457,7 +457,7 @@ class AzureProvider(BaseProvider):
     '''
     SUPPORTS_GEO = False
     SUPPORTS_DYNAMIC = True
-    SUPPORTS_POOL_VALUE_UP = True
+    SUPPORTS_POOL_VALUE_STATUS = True
     SUPPORTS_MULTIVALUE_PTR = True
     SUPPORTS = set(('A', 'AAAA', 'CAA', 'CNAME', 'MX', 'NS', 'PTR', 'SRV',
                     'TXT'))
@@ -808,14 +808,14 @@ class AzureProvider(BaseProvider):
                 defaults.add(val)
                 ep_name = ep_name[:-len('--default--')]
 
-            up = None
+            status = 'obey'
             if pool_ep.endpoint_status == 'Disabled':
-                up = False
+                status = 'down'
 
             values.append({
                 'value': val,
                 'weight': pool_ep.weight or 1,
-                'up': up,
+                'status': status,
             })
 
         return values
@@ -905,7 +905,7 @@ class AzureProvider(BaseProvider):
         return data
 
     def _process_desired_zone(self, desired):
-        # check for support of force up/down values
+        # check for status=up values
         for record in desired.records:
             if not getattr(record, 'dynamic', False):
                 continue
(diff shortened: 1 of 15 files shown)


octodns/provider/azuredns.py
```diff
@@ -898,6 +905,22 @@ def _data_for_dynamic(self, azrecord):
...
+            if not getattr(record, 'dynamic', False):
+                continue
+            for name, pool in record.dynamic.pools.items():
+                for value in pool.data['values']:
+                    if value['up']:
+                        # Azure only supports up=None & up=False, not up=True
+                        msg = f'up=True is not supported for "{name}" pool ' \
+                            f'in {record.fqdn}'
+                        fallback = 'ignoring it'
+                        self.supports_warn_or_except(msg, fallback)
```
ross, line 920: Will the `ignoring it` here result in persistent changes or will it just be set and ignored by azure when created and not seen as a diff when planning changes? If it will show as a diff we'll probably want to do a modification here like with `Route53Provider` even if it results in a setup that doesn't make sense.

octodns/record/__init__.py
```diff
@@ -418,6 +418,7 @@ def __init__(self, _id, data):
             {
                 'value': d['value'],
                 'weight': d.get('weight', 1),
+                'up': d.get('up'),  # TODO: add validation on this field
```
ross, line 421: The meaning of `up` probably needs to be documented in https://github.com/octodns/octodns/blob/master/docs/dynamic_records.md. 

It seems like the concept exists in NS1 (`up`) & Azure (`endpoint_status`), not 100% sure `up` is the best name to use for it. Maybe `force_status` or something a bit more explicit? Open to suggestions, mostly don't feel like `up` describes the behavior sufficiently.

ross (commented):
Overall seems reasonable. Few comments inline regarding lack of support/falling back and other misc bits.

Also note that as of https://github.com/octodns/octodns/pull/616 `ConstellixProvider` now supports dynamic. /cc @mintopia & @gartline re that. If it has a concept of always `up` then it can hopefully be accommodated. If not it'll need to implement a `_process_desired_zone` that warns and removes the up from the record.

viranch:
> note that as of #616 ConstellixProvider now supports dynamic.

Oh yes I missed that. I think it may be better to add a `SUPPORTS_` flag instead of adding to every dynamic-capable provider's `_process_desired_zone`. The code block in route53 can just be moved to the base provider that way.

octodns/provider/azuredns.py
viranch: Although no supported, the provider is marking in the endpoint name that the pool value is set to up=True so that we can parse the flag correctly in the plan phase and not show unnecessary changes. I'd rather no drop the up flag here since Azure does support up=False.

octodns/record/__init__.py
viranch: Totally agreed that just `up` is not very good. I scratched my head for quite a bit on this.

We essentially want 3 states: force up, force down, and obey monitor. One was enums, but it can become unintuitive to use and complex to validate.

I thought of `force_status`, but `force_status=False` can mean "don't force, obey monitor" instead of "don't serve". Which then led me to also considering separate flags `force_up` and `force_down`, but we then end up with 4 possible value combinations of those flags which again makes it complex to build exhaustive logic as well as validate all combinations.

I really don't have any better ideas here for the flag name.

octodns/provider/route53.py
```diff
@@ -941,6 +941,16 @@ def _process_desired_zone(self, desired):
...
 
+                # check for use of 'up' flag and warn/except if used
+                for name, pool in dynamic.pools.items():
+                    for value in pool.data['values']:
+                        if value['up'] is not None:
+                            msg = f'"up" flag is not supported for "{name}"' \
+                                f' pool in {record.fqdn}'
+                            fallback = 'ignoring it, octodns-sync command ' \
+                                'will keep showing changes'
+                            self.supports_warn_or_except(msg, fallback)
```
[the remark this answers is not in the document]
viranch: I like dropping the `up` flag much better than showing changes.

I took a quick peek at the Route53 portal and it seems like the functionality is possible, but I'm not familiar with Route53 APIs and the provider here, so I decided to let it be unsupported until someone can add it.

viranch:
I've implemented dropping the `up` flag for unsupported providers in BaseProvider. Let me know if it looks good now.

octodns/record/__init__.py
viranch: I did like Route53's notion of `healthcheck=False` forcing up and `weight=0` for forcing down. But we again end up with 2 flags.

ross: I guess my first inclination would lean towards the enum route:

`status': (up|down|obey)` with obey being the implicit value/default.

That seems a little more explicit than `'up': (true|false)` and having that key missing meaning obey. 

Thoughts?

ross (commented):
Base handling of the value looks good.

octodns/provider/azuredns.py
viranch: I now got what you meant, and it makes sense to me to reset the flag when there's an unsupported case (`up=True`). I've added that change to the PR.

octodns/record/__init__.py
viranch: Enum sounds intuitive for the user. Let me give it a shot.

octodns/provider/base.py
```diff
@@ -57,14 +57,28 @@ def _process_desired_zone(self, desired):
...
-                record.dynamic = None
-                desired.add_record(record, replace=True)
+            elif getattr(record, 'dynamic', False):
+                if self.SUPPORTS_DYNAMIC:
+                    pools = record.dynamic.pools.values()
+                    values = [p.data['values'] for p in pools]
+                    if any(isinstance(v['up'], bool) for v in values) and not \
+                       self.SUPPORTS_POOL_VALUE_UP:
+                        msg = f'"up" flag used in {record.fqdn} is not ' \
+                            f'supported'
```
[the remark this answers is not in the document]
viranch: Done

octodns/record/__init__.py
viranch: updated to an enum flag.

octodns/source/base.py
```diff
@@ -9,6 +9,7 @@
 class BaseSource(object):
 
     SUPPORTS_MULTIVALUE_PTR = False
+    SUPPORTS_POOL_VALUE_STATUS = False
```
viranch, line 12: I can't say if `BaseSource` is the right place for the flags or `BaseProvider`. I don't exactly see the implications of one vs the other.

ross: Iirc the thinking was that you could implement a source (not provider) that supported any of these advanced things  and thus the flags belong on the source. I don't think there's anything like that in practice, but as a thought expiriment imagine `YamlSource` rather than `YamlProvider` where it doesn't support being a target, but does support configuring values with whatever flags these are. 

Wouldn't be surprised in the slightest if there were shortcomings/issues if someone tried to do something of that sort since nothing has as of yet, but I believe that was the thinking.

viranch:
I had missed adding the flag to `dynamic_records.md`. With that done, this PR is ready for final review.

ross (approved):

PR #782, opened 2021-09-21
Retain octodns settings on record copy

viranch:
I realized copying dynamic records loses the healthcheck settings. Here's a fix to retain the `octodns` block on copy.


ross (approved):
👍 good catch. Thanks for adding a test for the behavior while you were in there.

Issue #776, opened 2021-09-16
Use of Route53 list_hosted_zones causes mass rate limiting in our scale tests


slandry90:
During scale testing of octodns we've hit an issue with the route53 provider.

To briefly describe our setup, we have ~12000 zones and are using a notifier to sync individual zones when changes are made to that zone. This prevents us from needing to iterate through thousands of zones and records that we know have not changed since the last execution.

https://github.com/octodns/octodns/blob/4c9f0137a1f1a378e190670a5279363111c87d2b/octodns/provider/route53.py#L658

The r53_zones function is currently used to retrieve the hosted_zone_id of a particular zone, in order to do that it runs a list_hosted_zone call. The list_hosted_zone call has a hard pagination limit of 100. In our case where we have about 12000 zones in our route53 account, this means that each time a new sync job runs it has to execute at least 120 list_hosted_zones calls in order to retrieve all the zones just to get the singular hosted_zone_id which could be retrieved via a list_hosted_zones_by_name call.

I know different people use octodns differently and swapping the functionality from list_hosted_zone to list_hosted_zones_by_name could have some adverse effects for those that are running 100s of zones within the same sync job (would have the same problem but in a different direction) so would like some thoughts on a route forward here.

I'd like to initially propose adding a flag in route53 to toggle the behavior to list_hosted_zones_by_name, leaving the current behavior as the default but allowing users like us to be able to avoid the mass number of api calls. I'd rather not fork if we don't need to but we can do that as well if this isn't a use case the group here is game for supporting.

Would also be happy to help with testing or submitting a PR if that helps as well.

Thanks

ross:
> 'd like to initially propose adding a flag in route53 to toggle the behavior to list_hosted_zones_by_name, leaving the current behavior as the default but allowing users like us to be able to avoid the mass number of api calls. I'd rather not fork if we don't need to but we can do that as well if this isn't a use case the group here is game for supporting.

👍 to that approach. I'll think about it and see if I can come up with anything auto-magic/smarter, but nothing immediately comes to mind. 

> Would also be happy to help with testing or submitting a PR if that helps as well.

Definitely. Will be happy to 👁️ one.

slandry90:
@ross proposed the PR here, https://github.com/octodns/octodns/pull/783. Let me know if I need to make changes or improvements!

ross:
Fixed in https://github.com/octodns/octodns/pull/783

slandry90:
[...]

PR #783, opened 2021-09-22
Add ability to optionally retrieve route53 hosted zone ids by name

slandry90:
Issue described in https://github.com/octodns/octodns/issues/776


octodns/provider/route53.py
```diff
@@ -659,7 +661,19 @@ def r53_zones(self):
...
+            response = self._conn.list_hosted_zones_by_name(
+                DNSName=name, MaxItems="1"
+            )
+            if len(response['HostedZones']) != 0:
+                # if there is a response that starts with the name
+                if response['HostedZones'][0]['Name'].startswith(name):
+                    id = response['HostedZones'][0]['Id']
+                    self.log.debug('get_zones_by_name:   id=%s', id)
+                    return id
+        elif name in self.r53_zones:
```
ross, line 676: Looks like `_get_zone_id` is called a few times during the course of planning. I believe this will result in calling `list_hosted_zones_by_name` a few times for each zone, quick scan looks like 3. If this is working for you that's probably ok, but it'll be a bit slower and more likely to hit rate limits. 

If this is working well enough I'm ok with it as-is. If you'd like to look into solving that I think the best option may be to remove `@property` from `r53_zone` and add a `name` param. That should allow the cache to still work and decide whether to get all or fetch just what it's asked for, something like:

```python
    def r53_zone(self, name):
        if self._r53_zones is None:
            if self. get_zones_by_name:
                self._r53_zones = {}
            else:
                ... existing implementation ...
        zone_id = self._r53_zones.get(name, None)
        if not zone_id and self.get_zones_by_name:
            ... new lookup by name ...
            self._r53_zones[name] = zone_id
        return zone_id
```

I think w/something like that `def _get_zone_id` could remain as it was. Lmk which way you'd like to go.

slandry90: This is a good point, I was originally torn on how I wanted to handle this and was avoiding removing this property on the r53_zone function. 

On the latest commit, I rewrote the r53_zone function to instead be an update function which updates the _r53_zones dictionary and handled the different cases where the dictionary would need to be updated. It's now functioning as 1 check per zone if using get_zones_by_name.

ross (approved):

ross:
Fixes https://github.com/octodns/octodns/pull/783

PR #787, opened 2021-09-29
properly return and cache none replies for get_zone_by_name in route53

slandry90:
Followup to an earlier PR, during testing additional testing it was found that when a zone id did not exist it was returning a None back when attempting to create a zone.

Adding additional tests to catch this condition as well.


slandry90:
@ross followup from my previous MR, we did more extensive real world testing and found that there was an issue during apply with the caching of zones that couldn't be found causing the create to return None instead of the correct id. Added additional test cases to cover this as well. Verified this works in our end to end testing as well.

ross (approved):

PR #794, opened 2021-10-14
Overwrite invalid NS1 dynamic records

viranch:
Prevent octoDNS from aborting plan phase when it runs into an invalid/unparsable record, and instead treat it as an invalid record to let octoDNS overwrite it with proper filter configs and answer+region notes. This can happen if the record was created outside of octoDNS, by hand or with independent scripts.


viranch:
> example record config that is causing problems

Think any advanced record created on NS1 manually by hand outside of octoDNS. That implies possibly different set/order of filter chains and no notes whatsoever. Unexpected filter chains makes octoDNS give up with an exception, and missing notes just crashes it.

That said, I'm trying a different approach as this PR would introduce garbage invalid values when `octodns-dump` is used.

PR #798, opened 2021-10-20
Improvements in NS1 filter chain management

viranch:
1. Instead of giving up when an invalid filter chain is seen, mark the record for update to fix it.

2. Better handling of mixed disabled: Instead of throwing exception when there's mixed disabled flags, assume `disabled=False` wherever absent and proceed `with_disabled=True` everywhere; essentially removing `with_disabled` switch altogether and filters becoming almost hard-coded. I wanted to have the inverse - not have disabled flag unless it comes in set to `True` - but that would cause a lot of record updates on octoDNS version downgrades, since `with_disabled` defaults to `True` in `_params_for_dynamic` today.

3. Cache filter chains while honoring record type - current implementation in `_get_ns1_filters` doesn't remember record's type so A/AAAA dynamic records will conflict if they were to have different filter chains. Additional caching avoids more API calls, even though we're already caching the records so this shouldn't make practical difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+
+## v0.9.15 - 202?-??-?? - ??
+
+#### Noteworthy changes
+
+* NS1 provider has received improvements to the dynamic record implementation.
+  As a result, if octoDNS is downgraded from this version, any dynamic records
+  created or updated using this version will show an update.
+
 ## v0.9.14 - 2021-10-10 - A new supports system
 
 #### Noteworthy changes


ross:
> Just an upgrade won't trigger any record updates, it will silently ignore all the `disabled=False` flags and pass the filter chain equality check. Only a downgrade would trigger updates, and only for those records that were created/updated using the newer version.

Ah. I misunderstood. That's probably acceptable so long as it's clearly called out in the CHANGELOG as a noteworthy change.

PR #809, opened 2021-11-15
Skip explicit countries when converting continent to countries in NS1

viranch:
`NS1Provider` converts `NA` and `OC` to individual list of countries. This creates a problem when a subset of those countries are already defined in another rule pointing to a different pool. This may lead to an undefined behavior as those countries point to two different DNS responses in NS1 record. 

This PR skips over the explicitly listed countries when expanding continents to countries; and uses region notes to parse the incomplete list of continent countries back to the continent in octoDNS record.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 * NS1 provider has received improvements to the dynamic record implementation.
   As a result, if octoDNS is downgraded from this version, any dynamic records
   created or updated using this version will show an update.
+* An edge-case bug related to geo rules involving continents in NS1 provider
+  has been fixed in this version. However, it will not show/fix the records that
+  match this edge-case. See https://github.com/octodns/octodns/pull/809 for
+  more information. If octoDNS is downgraded from this version, any dynamic
+  records created or updated using this version and matching the said edge-case
+  will not be read/parsed correctly by the older version and will show a diff.
 
 ## v0.9.14 - 2021-10-10 - A new supports system
 
--- a/octodns/provider/ns1.py
+++ b/octodns/provider/ns1.py
@@ -613,7 +613,7 @@ class Ns1Provider(BaseProvider):
 
         return default, pools
 
-    def _parse_rule_geos(self, meta):
+    def _parse_rule_geos(self, meta, notes):
         geos = set()
 
         for georegion in meta.get('georegion', []):
@@ -628,6 +628,11 @@ class Ns1Provider(BaseProvider):
         # regions will be present in meta['country']. If all the countries
         # in _CONTINENT_TO_LIST_OF_COUNTRIES[<region>] list are found,
         # set the continent as the region and remove individual countries
+
+        # continents that don't have all countries here because a subset of
+        # them were used in another rule, but we still need this rule to use
+        # continent instead of the remaining subset of its countries
+        continents_from_notes = set(notes.get('continents', '').split(','))
 
         special_continents = dict()
         for country in meta.get('country', []):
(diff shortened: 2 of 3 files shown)


ross (commented):
Wish I/we'd thought to do this originally. 

To check my read (it's been a long time since I was in this bit of things) this doesn't change any of the existing behavior around continents and their mapping to a set of countries when required. 

It just adds an additional tracking of when a continent was included (via notes) and makes sure to include individual countries that are explicit on top of when they are implicitly included b/c of continent mapping?

viranch:
Mostly correct, yes. Continent->country mapping is unaffected when none of those mapped countries are used elsewhere explicitly.

The explicit countries, when used, are skipped from the list of implicit continent mapping, rather than put on top of them. Since we can no longer map the subset back to its continent, we make use of notes to record that we want the continent not the individual countries.

ross:
> The explicit countries, when used, are skipped from the list of implicit continent mapping, rather than put on top of them. Since we can no longer map the subset back to its continent, we make use of notes to record that we want the continent not the individual countries.

If all of the countries are still there it will get marked as a continent even though it won't be in the notes as it always previously was?

Looks like this will set the notes continent on the next update, but am I correct in my read that it won't show that as a diff?

viranch:
The continent will stay in the notes whether or not all of it's countries are present.

ross:
I'm mostly wondering/concerned about the impact on existing setups coming form older versions or downgrading. Cases I can think of:

* coming from old version with continent that requires country list, no explicit countries
   * will populate still convert the set of countries into the continent? 
   * will it show a diff b/c the change in notes? (ideally won't b/c that'd 😨) don't think it would have in the past and not sure it would now
   * when the record otherwise changes i assume the continent will be added to the notes (seems fine)
* coming from old version with continent that requires country list, some explicit countries
   * as i understand it this wasn't possible before this PR, if so then 👍 
* push a record with the new version, continent marked in the notes, no explicit countries, downgrade to old version
   * will everything just work the way it has in the past by turning the set into the continent? seems likely.
* push a record with the new version, continent marked in the notes, has explicit countries too, downgrade to old version
   * what happens here? my guess would be it translates the countries into continent and loses the explicit countries so shows as a diff.

viranch:
Numbered answers to the questions:

1a. yes it will correctly map full country set to their continent in absence of the notes
1b. no we don't take notes into account when planning changes
1c. correct, the continent will be added to notes on any record update

2a. it was possible to do this but the resultant behavior in DNS responses would've been undefined (due to explicit countries being mapped to multiple pools). new version will parse such records correctly but won't show a diff for fixing the buggy setup - this is probably ok but not ideal. In order to show a diff, we'd need some transitional code in `_extra_changes` to catch such records and mark them for update, not sure if we have a standardized framework/process for handling this kind of version changes in code.

3a. correct, the older version will turn the set of countries to their continent because all countries are present

4a. yes this scenario will show a diff as it doesn't know that the subset of implicit countries actually map to their continent.  I believe this is expected if you go towards the version with the bug.

ross:
> In order to show a diff, we'd need some transitional code in `_extra_changes` to catch such records and mark them for update, not sure if we have a standardized framework/process for handling this kind of version changes in code.

IF we wanted to `_extra_changes` would be the way to go. The fix/change should at least be included in a noteworthy changes section of the changelog. I think the use case is probably edge-case-y enough that we don't have to handle it if you don't want to beother.

Overall 👍. Will look things over once more now...

ross (approved):
Please ad a line to noteworthy changes in the Changelog and then feel free to merge.

PR #811, opened 2021-11-18
Fix geo comparison in Azure DNS traffic manager profiles

viranch:


octodns/provider/azuredns.py
```diff
@@ -913,6 +923,7 @@ def _process_desired_zone(self, desired):
                     if value['status'] == 'up':
                         # Azure only supports obey and down, not up
                         up_pools.append(name)
+                        break
```
viranch, line 926: also sneaking in a small fix for logging only unique pool names for unsupported `status=up` flag value.

ross (approved):

Issue #622, opened 2020-11-03
Future of octoDNS - Road to 1.0


ross:
Using this issue to write out some thoughts/plans for octoDNS's path forward. There's no specific timeline on any of this, some of it will happen almost immediately other bits will come over time.

### Separate Org - github.com/octodns

The biggest change is that I plan to move things from https://github.com/github/octodns/ to https://github.com/octodns/. That will put this repo at https://github.com/octodns/octodns/. The good news is that everything should pretty much redirect and work normally if you hit the old URL. There's a few reasons behind this change, the primary one being so that we can break up the core of octodns from the providers. This give us a better setup on a few fronts.

#### Maintainers, Access, &amp; Testing

Having separate repos for each provider will allow outside contributors to be granted further access and own/drive specific providers. There are many providers where I don't have access to run or test with making it difficult for me to verify changes and making me/GitHub a bottleneck. Having things in separate repos that don't live under the /github org will allow us to grant control to the authors of the providers.

#### Release Schedules

Having repos like https://github.com/octodns/route53, https://github.com/octodns/azure, https://github.com/octodns/ns1, ... This will allow each provider to release on its own schedule as makes sense for that provider and structure things more cleanly in terms of issues and PRs. 

#### Requirements &amp; Dependancies

This will also result in a much better setup around requirements. With everything in a single repo we've chosen to only include the core requirements for octoDNS to avoid installing 100's of modules, most of which wouldn't be used in any given install. With separate modules each provider can spell out their exact requirements in their setup.py/requirements.txt and things will work as people expect. 

### The path from here to there

Again no specific timeline, the first few of these will likely happen over the next few weeks. The "api", testing, and breakup of providers will likely happen over time.

* [x] Create the https://github.com/octodns/ org
* [x] Grant access (owners/admins) to roughly match those on https://github.com/github/octodns
* [x] Move https://github.com/github/octodns/ to https://github.com/octodns/octodns/
* [x] Update docs, readmes, pypi etc to point to the new location
* [x] Codify the existing Source & Provider structure as a formal API & write extensive behavior tests for that api to help ensure that changes won't break providers
* [x] Create a new "template" provider to establish the structure and behavior of sources/provider sub-modules and to serve as a starting point for pulling existing providers out and for developing new providers
* [x] Look into running tests for all the submodules against the octoDNS core on https://github.com/octodns/octodns/ PRs.
* [x] Split out each provider into a separate repo & module.
* ...

yeled:
[...]

PR #814, opened 2021-11-25
Adds awsacm processor that completely ignores AWS ACM validation CNAME records at source and destination

nikolay-te:
Adds awsacm processor that completely ignores AWS ACM validation CNAME records at source and destination

### DIFF
--- a/octodns/processor/awsacm.py
+++ b/octodns/processor/awsacm.py
@@ -1,0 +1,46 @@
+#
+# Ignores AWS ACM validation CNAME records.
+#
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+from logging import getLogger
+
+from .base import BaseProcessor
+
+
+class AwsAcmMangingProcessor(BaseProcessor):
+    log = getLogger('AwsAcmMangingProcessor')
+
+    def __init__(self, name):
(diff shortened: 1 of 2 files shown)


gaston-te (approved):

octodns/processor/awsacm.py
```diff
@@ -0,0 +1,46 @@
...
+        ...
+
+        zones:
+          something.com.:
+          ...
+          processors:
+            - awsacm
+          ...
+        '''
+        super(AwsAcmMangingProcessor, self).__init__(name)
```
ross, line 31: Looks like `__init__` is overridden just to add the doc string? If so it's probably fine to put it on the class (above `log` to avoid having to override. We haven't done the greatest job with documentation (consistency) so it should be fine to live there until if/when we go through and improve things.

ross (commented):
Looks reasonable. After the docstring tweak we can get it 🚢 ed

Eventually this would likely get pulled out into an AWS provider module when things are broken up someday, cc https://github.com/octodns/octodns/issues/622

ross (approved):

ross:
Thanks.

nikolay-te:
Thanks, @ross!

PR #816, opened 2021-12-03
Route53 weight support

tkwy-rick:
Tried to build in some basic support for weighted records in Route53. Reason behind this is because dynamic-records is not what we are searching for and is somewhat depending on health checks. 

Python Tests etc still need to be added. but it works like follows:

```yaml
test:
  weighted:
    test-1:
      value: test1.example.com.
      weight: 200
    test-2:
      value: test2.example.com.
      weight: 100
    test-3:
      value: test3.example.com.
      weight: 100
  ttl: 60
  type: CNAME  # or A
  # fallback with weight: 0
  # identifier: _octodns_default
  value: test0.example.com.
  ```

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -45,6 +45,13 @@ class _AggregateTarget(object):
                 return False
         return True
 
+    @property
+    def SUPPORTS_WEIGHTED(self):
+        for target in self.targets:
+            if not target.SUPPORTS_WEIGHTED:
+                return False
+        return True
+
 
 class MakeThreadFuture(object):
 
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -470,6 +470,7 @@ class AzureProvider(BaseProvider):
     '''
     SUPPORTS_GEO = False
     SUPPORTS_DYNAMIC = True
+    SUPPORTS_WEIGHTED = False
     SUPPORTS_POOL_VALUE_STATUS = True
     SUPPORTS_MULTIVALUE_PTR = True
     SUPPORTS = set(('A', 'AAAA', 'CAA', 'CNAME', 'MX', 'NS', 'PTR', 'SRV',
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -76,10 +76,6 @@ class BaseProvider(BaseSource):
                     for pool in record.dynamic.pools.values():
                         for value in pool.data['values']:
                             value['status'] = 'obey'
-
-                    print("_1_____JJJOOWOOWOWOWOWOOWOWOWOWO")
-                    print(record)
-                    print("_1_____JJJOOWOOWOWOWOWOOWOWOWOWO")
                     desired.add_record(record, replace=True)
                 else:
                     msg = f'dynamic records not supported for {record.fqdn}'
@@ -87,15 +83,13 @@ class BaseProvider(BaseSource):
                     self.supports_warn_or_except(msg, fallback)
                     record = record.copy()
                     record.dynamic = None
-                    print("_2_____JJJOOWOOWOWOWOWOOWOWOWOWO")
-                    print(record)
-                    print("_2_____JJJOOWOOWOWOWOWOOWOWOWOWO")
                     desired.add_record(record, replace=True)
             # elif getattr(record, 'weighted', False):
             #     if self.SUPPORTS_WEIGHTED:
             #         if record._type in self.SUPPORTS_WEIGHTED_TYPES:
             #             print("# BASEPROVIDER")
             #             print(record._type)
+            #             print(record.weighted)
             #             print("# BASEPROVIDER")
             #     else:
             #         msg = f'weighted records not supported for {record.fqdn}'
@@ -161,14 +155,10 @@ class BaseProvider(BaseSource):
         before = len(changes)
         changes = [c for c in changes if self._include_change(c)]
         after = len(changes)
+
         if before != after:
             self.log.info('plan:   filtered out %s changes', before - after)
 
-        # allow the provider to add extra changes it needs
-        for c in changes:
-            print(c)
-            print(dir(c))
-            print()
         extra = self._extra_changes(existing=existing, desired=desired,
                                     changes=changes)
         if extra:
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -76,6 +76,10 @@ class BaseProvider(BaseSource):
                     for pool in record.dynamic.pools.values():
                         for value in pool.data['values']:
                             value['status'] = 'obey'
+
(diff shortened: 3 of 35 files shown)

### CODE REVIEW

**Comment by ross** [comment]
Did you take a look at the `status` flag to healthchecks? That's likely the way to get the behavior you're looking for. I'm not sure whether Route53 currently supports it, but it should be possible to do so if not.

**Comment by viranch** [comment]
@tkwy-rick you can achieve the same behavior with a dynamic record:

```yaml
test:
  dynamic:
    pools:
      weighted:
        values:
        - value: test1.example.com.
          weight: 200
          status: up
        - value: test2.example.com.
          weight: 100
          status: up
        - value: test3.example.com.
          weight: 100
          status: up
    rules:
    - pool: weighted
  ttl: 60
  type: CNAME  # or A
  # fallback with weight: 0
  # identifier: _octodns_default
  value: test0.example.com.
```

`status: up` skips health-checking and always serves the value (see https://github.com/octodns/octodns/blob/master/docs/dynamic_records.md#health-checks). But this is not implemented in the Route53 provider, so the right way would be add support for that flag.

side note: you'll also want to normalize the weights as octodns caps the weight to 100.

tkwy-rick:
@viranch Didn't thought of implementing that indeed. might work as well indeed. Will see what i can figure out here. what do you suggest when status = down. set weight to 0? As i don't wan't to remove the records.

ross:
> @viranch Didn't thought of implementing that indeed. might work as well indeed. Will see what i can figure out here. what do you suggest when status = down. set weight to 0? As i don't wan't to remove the records.

Looks like health checks support a disabled settling that claims it assumes healthy. That sounds equivalent to `up`

Screen Shot 2021-12-06 at 6 08 58 AM

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.create_health_check `disabled`. 

`down` my be supportable with `disabled` and `invert`, but dunno.

tkwy-rick:
That might sound like a pretty good option. Just tried it manually and it could work! If i find the time, i will see if i can implement it.

tkwy-rick:
Tried to implement it. but this will not work for private ranges. the IP in dynamic must be public!

```yaml
test:
  octodns:
    healthcheck:
      path: /test-dns
    route53:
      healthcheck:
        failure_threshold: 2
  dynamic:
    pools:
      weighted:
        values:
          - value: 1.1.1.1
            # weight: 1
            status: up
          - value: 2.2.2.2
            # weight: 2
            status: up
          - value: 3.3.3.5
            weight: 3
            status: down
    rules:
      - pool: weighted
  ttl: 60
  type: A
  value: 1.1.1.4
  ```

ross:
> Tried to implement it. but this will not work for private ranges. the IP in dynamic must be public!

Not following, the config you have there isn't using private IPs? You saying Route53 won't let you create health checks with private IP addresses? 

If so the solution may be to skip creating the health check when `status: up` rather than having one that is disabled.

tkwy-rick:
@ross What i did was use indeed the healthchecks but control it using the Disabled & Inverted option. https://github.com/tkwy-rick/octodns/pull/2

ross:
> and set weight on 0 when status = down.

I don't think that would work correctly in other cases where there are fallbacks etc. Since the healthcheck could still be healthy.

tkwy-rick:
Maybe something like this. It's covers the case i need it for: https://github.com/tkwy-rick/octodns/pull/3/files

ross:
I have the provider implementation for Route53 `status` support pushed up to https://github.com/octodns/octodns/pull/819. I'm still working on the tests and coverage and have a few more things to verify.

I think things were getting more complicated that absolutely required by combining `status` and RFC1918 addresses into the same thing/problem. In reality they're unrelated. That PR will add support for `status`. A separate PR will add support for using RFC1918 values in Route53 dynamic records. I suspect other providers may not allow them to be health checked either so that will likely turn into a PR that touches more than just Route53.

tkwy-rick:
@ross Great work you did on #819. About the RFC1918 i was thinking. What if when IPaddress [is_private](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_private) we override the value that is being put in the healthcheck. but only when status = up or down (so healthcheck is disabled) 

We could set value to something like `http://_octodns_disabled_.healtcheck`

Screenshot from 2021-12-10 10-01-41

or do something similar like i did in https://github.com/tkwy-rick/octodns/pull/3 but use the status option for this, and add an extra status option: `disabled` Might be more work. but it might be interesting, because route53 for example has a default quota of 200 active health checks. this allows you to control the traffic only with weight. (status=disabled, weight=0 is allowed)

ross:
> @ross Great work you did on #819. About the RFC1918 i was thinking. ...

Yeah. That's one of the two options I'm looking into currently. 

I'm actually starting with a validation that will require `status: up` on any non-routable IP addresses. That'll make it explicit that no health checking will be done on the values and it would also give people a way to override it with `lenient` if for some reason/with some provider it is allowed. 

Once that's in place it'll be up to each provider to figure out how to support handle the private IPs appropriately. For Route53 and NS1 (which I verified errors on non-routable IPs) my current plan is to have no health check on the value. If for some reason that doesn't work out the backup plan is to have a dummy value as mentioned above. 

I haven't had a chance to look at Azure and don't have access to Constellix but I'd be surprised if either allow non-routable IPs.

viranch:
Are we sure we want to add special handling for private IPs? Doesn't sound like the job of octoDNS to handle that. What about CNAMEs that point to private IPs?

ross:
> Are we sure we want to add special handling for private IPs? Doesn't sound like the job of octoDNS to handle that. What about CNAMEs that point to private IPs?

I was hoping that it'd be simple to detect non-routable IPs and flag that `up` is required on them. That said after that each provider would still have to do something special to avoid creating health checks for them. CNAMEs are a hiccup that I ran into when looking at the validation (a blank function that needed an implementation.) I ended up deciding that it's not really possible to detect whether they're routable or not so yeah. This is looking pretty ugly/hairy overall. 

I plan to keep thinking about it and see if a clean/full solution comes to mind, but as of right now it's not looking great.

tkwy-rick:
The reason it's only Route53 Specific was the reason why i implemented it with some route53 specific value under `octodns` in https://github.com/tkwy-rick/octodns/pull/3 this will indeed remove health-checks entirely. For me at least this will. I think the combo with `status` gives an option to manage traffic for public ips/Cnames. and a route53 HC disable flag can work for internal ips. a little documention around it for this specific setting and we're good. There are already some custom settings: https://github.com/octodns/octodns/blob/master/docs/dynamic_records.md#route53-health-check-options

ross:
This should now be doable as of https://github.com/octodns/octodns/pull/819, `status: up` will avoid adding a health check.

PR #819, opened 2021-12-09
Add value status support to Route53 dynamic

ross:
Adds dynamic record value status support to Route53. This will allow `status: (up|down|obey)`.

TODO:
- [x] Verify that `disabled=True` `inverted=True` gets the `down` behavior we're after
- [x] Look into upgrade/downgrade behavior
- [x] Test fixes and additions

/cc @viranch for early 👀 on the implementation
/cc @tkwy-rick & https://github.com/octodns/octodns/pull/816 as an fyi

### DIFF
--- a/octodns/provider/route53.py
+++ b/octodns/provider/route53.py
@@ -80,9 +80,11 @@ class _Route53Record(EqualityTupleMixin):
             # be served out according to their weights
             for i, value in enumerate(pool.data['values']):
                 weight = value['weight']
+                status = value['status']
                 value = value['value']
                 ret.add(_Route53DynamicValue(provider, record, pool_name,
-                                             value, weight, i, creating))
+                                             value, weight, status, i,
+                                             creating))
 
         # Rules
         for i, rule in enumerate(record.dynamic.rules):
@@ -345,19 +347,21 @@ class _Route53DynamicRule(_Route53Record):
 
 class _Route53DynamicValue(_Route53Record):
 
-    def __init__(self, provider, record, pool_name, value, weight, index,
-                 creating):
+    def __init__(self, provider, record, pool_name, value, weight, status,
+                 index, creating):
         fqdn_override = f'_octodns-{pool_name}-value.{record.fqdn}'
         super(_Route53DynamicValue, self).__init__(provider, record, creating,
                                                    fqdn_override=fqdn_override)
 
         self.pool_name = pool_name
+        self.status = status
         self.index = index
         value_convert = getattr(self, f'_value_convert_{record._type}')
         self.value = value_convert(value, record)
         self.weight = weight
 
         self.health_check_id = provider.get_health_check_id(record, self.value,
+                                                            self.status,
                                                             creating)
 
     @property
@@ -433,7 +437,7 @@ class _Route53GeoRecord(_Route53Record):
 
         value = geo.values[0]
         self.health_check_id = provider.get_health_check_id(record, value,
-                                                            creating)
+                                                            'obey', creating)
 
     def mod(self, action, existing_rrsets):
         geo = self.geo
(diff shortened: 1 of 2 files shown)


ross:
> * [x]  Verify that `disabled=True` `inverted=True` gets the `down` behavior we're after

Good news here. This seems to work:

```yaml
---
? ''
: dynamic:
    pools:
      main:
        values:
          - status: up
            value: 1.2.3.4
            weight: 10
          - status: down
            value: 192.30.255.112
            weight: 10
          - value: 3.4.5.6
            weight: 12
          - value: 192.30.255.113
            weight: 12
    rules:
      - pool: main
  octodns:
    healthcheck:
      host: github.com
  type: A
  values:
    - 1.2.3.4
    - 2.3.4.5
```

```
$ for i in `seq 100`; do dig +short exxampled.com @xxx.yyy.zzz.; done | sort | uniq -c
  47 1.2.3.4
  53 192.30.255.113
```

Screen Shot 2021-12-09 at 10 51 03 AM

`1.2.3.4` is forced up even though it would never pass health checks. `192.30.255.112` would pass health checks, but is forced down by `disabled` and `inverted`. The last two are using obey and work correctly.

viranch:
simple enough, looks promising. Not a R53 expert, but I don't see any regressions from the code. An upgrade will start honoring the `status` flag, and a downgrade will just overlook `disabled`/`inverted` flags so values will default to `obey` and match the desired zone?

viranch:
> if you upgraded, and started using status and then tried to downgrade. You'd get errors

with `strict_supports` defaulting to `False` (I think?), it will just convert to status=obey with a warning but not actually change the values to 'obey' as old version is oblivious of `Disabled` and `Inverted`. Even better if `strict_supports` is set to `True` in which case we will just refuse to proceed. 

> It's a pretty edge-casey thing and I would be surprised if anyone hits it and worse case it's just not going to make changes and leave things as they are so

Agreed.

viranch (approved):

ross:
> with `strict_supports` defaulting to `False` (I think?), it will just convert to status=obey with a warning but not actually change the values to 'obey' as old version is oblivious of `Disabled` and `Inverted`. Even better if `strict_supports` is set to `True` in which case we will just refuse to proceed.

Ah. Forgot about the supports stuff mapping it. :+1:

ross:
Haven't yet shipped this as after looking at NS1 more closely I found that it doesn't create/install health checks when `status: up` and I'd like to match that here. I think that'll allow people to work around the private IP problem and avoid potential health check related costs.

ross:
Base of the `up` means no check is now in place. There's some hiccups with the way the existing extra changes method works, assuming no health check means it needs an update. Will look into that in a bit.

ross:
OK. This should be good (again) now and I've checked the behavior during upgrades/downgrades.

viranch:
I was wondering about the no-healthcheck approach during first review but assumed it's not possible with R53. Glad that's not the case and we're ultimately saving cost!

viranch:
Looking at the diff (sorry I'm late here and I know it's already merged), can't we achieve no-healthcheck for `status: down` also with `weight=0`?

One glitch I can think of is if weight is set to a value with `status: down` then we can't set/read that weight value; we could just ignore/reset it in `process_desired_zone`.

ross:
> One glitch I can think of is if weight is set to a value with `status: down` then we can't set/read that weight value; we could just ignore/reset it in `process_desired_zone`.

I think I said something about that elsewhere, but I'm not finding it atm. I suspect that `status: down` as `weight: 0` wouldn't get the correct behavior in terms of fallbacks, e.g. say a pool with 2 answers one, `down`, the other `obey` if the `obey` one started failing health checks and we implement `down` by `weight: 0` would the provider know to fallback? Same would be the case for both being `down`. In my mind `weight: 0` isn't the same thing as unhealthy and thus wouldn't result in a fallback, but that's not something I've explicitly tested.

viranch:
Fair enough. I'm curious to test it out, hopefully soon.

PR #821, opened 2021-12-20
process_desired_zone check for non routable IPs

ross:
This PR adds `process_desired_zone` supports checking for non-routable IPs in dynamic records. When found `strict: true` will error out with a message about the problem. When `false` the record(s) will have `status: up` set which avoids using health checks.

I think this is a fairly clean way to coach the needed setup for anyone who tries to use private IPs in dynamic records, e.g. for a multi-value weight setup like in https://github.com/octodns/octodns/pull/816. 

@viranch thoughts?

/cc https://github.com/octodns/octodns/pull/816 for prior discussion
/cc https://github.com/octodns/octodns/pull/819 for related work

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -5,6 +5,8 @@
 from __future__ import absolute_import, division, print_function, \
     unicode_literals
 
+from ipaddress import ip_address
+
 from ..source.base import BaseSource
 from ..zone import Zone
 from .plan import Plan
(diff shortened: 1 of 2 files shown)


octodns/provider/base.py
```diff
@@ -30,6 +32,24 @@ def __init__(self, id, apply_disabled=False,
...
 
+    def _check_dynamic_for_private_ips(self, desired, record):
+        if record._type not in ('A', 'AAAA'):
+            # won't necessarily contain IPs
+            return
+        for name, pool in record.dynamic.pools.items():
+            for i, value in enumerate(pool.data['values']):
+                ip = ip_address(value['value'])
+                status = value['status']
+                if not ip.is_global and status != 'up':
```
viranch, line 43: I think we should allow private IPs with `status: down` instead of forcing them `up`. Health-check is only present if `status: obey` so only `obey`s should be forced `up`.

octodns/provider/base.py
```diff
@@ -30,6 +32,24 @@ def __init__(self, id, apply_disabled=False,
...
+                    msg = f'non-routable dynamic record {record.fqdn} value ' \
+                        f'{ip} requires status `up` and has `{status}`'
+                    fallback = 'setting status to `up`'
+                    self.supports_warn_or_except(msg, fallback)
+                    record = record.copy()
+                    pool = record.dynamic.pools[name]
+                    pool.data['values'][i]['status'] = 'up'
```
viranch, line 50: Azure DNS does not support `status: up` (supports `obey` and `down`), so this will mess up the apply phase. Since the base `process_desired_zone` is called after the provider's override, AzureProvider won't be able to fix it.

ross:
I don't feel strongly about it, I'd just wanted to find a clean way to go about detecting the situation (for A/AAAA.) Fine with leaving it as-is now that it's at least possible.

PR #822, opened 2021-12-21
Extract PowerDNSProvider and implement shim

ross:
This is the prototype PR for extracting providers into individual repos. The reasons for this are outlined in the [road to 1.0 issue](https://github.com/octodns/octodns/issues/622), but the gist is that it will allow things to be owned by people with expertise/access to specific providers and they can release on their own schedule and not be tied to larger octoDNS (core) releases. 

The provider has been extracted into https://github.com/octodns/octodns-powerdns without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-powerdns/.

Rough plan is that this PR will become a template for the process, potentially even with a template repo, then all of the providers with the exception of Yaml would be removed and shimmed in the next release (or two/few) and things would live that way for a while until the rest of the roadmap issue is done. Then a 1.0 release would be cut that removes the shims.

/cc @viranch @yzguy for general thoughts & feedback


viranch:
Nice! Do we need to specify an additional pip install for octodns-powerdns in the table or requirements.txt?

ross:
> Nice! Do we need to specify an additional pip install for octodns-powerdns in the table or requirements.txt?

I set the name up to link to the repo and the repo has the details on what's needed for pip so I was planning to leave that in the provider specific repos. 

I'm also torn about whether or not the provider support details belong in this table or in the repos. It seems more logical to have it in the repos since it's tied closely with the code there, but at the same time it's probably nice to have it in a central place if you want to look at what supports what at a high level.

yzguy:
This is great, I have never really broken anything out into it's own like this so this will be a good reference so I can help.

I think it makes sense to list the provider in the OctoDNS repo, linking to the provider repo, but then in the provider repo is where the support details go. It's going to get quickly out of sync if something support-wise changes in one of the provider repos, then someone needs to go change it here in this one.

Unless there is a easy or clever way to dynamically show it in this repo, pulling from the other repos for the up to date information.

ross:
> I think it makes sense to list the provider in the OctoDNS repo, linking to the provider repo, but then in the provider repo is where the support details go. It's going to get quickly out of sync if something support-wise changes in one of the provider repos, then someone needs to go change it here in this one.

👍 I eventually realized what (I think) @viranch was meaning, that we should have a column for the module name, i.e. `octodns-powerdns`. 

Will do a PR to add that column and start to remove the other info for things that have been extracted.

> Unless there is a easy or clever way to dynamically show it in this repo, pulling from the other repos for the up to date information.

It'd be possible, but would probably require writing code to dynamically generate the table by looking at all the provider repos/classes. Doable, but probably not worth it. We'll see if i get bored enough at some point :grin:

viranch:
> 👍 I eventually realized what (I think) @viranch was meaning, that we should have a column for the module name, i.e. octodns-powerdns.

Yep that's what I meant.

Another long term complexity I see is having to define a version support matrix between octoDNS and provider package versions, since now they all can be individually upgraded/downgraded. This might also make every PR that much more complex when we try to simulate multiple combinations of upgrade/downgrade paths.

ross:
> Another long term complexity I see is having to define a version support matrix between octoDNS and provider package versions, since now they all can be individually upgraded/downgraded. This might also make every PR that much more complex when we try to simulate multiple combinations of upgrade/downgrade paths.

Yeah. More than a text matrix I'm hoping that we'll be able to keep the requirements in sync via the `install_requires` versions in `setup.py`.  It's not very interesting yet, but https://github.com/octodns/octodns-powerdns/blob/400c6208ab635a1a51fb65f370706a48e7022d75/setup.py#L30

I've always tried to aim for backwards compatibility in the core octoDNS api and this will be a forcing function on that. I hope to add some specific api tests etc to cover the expected/required behaviors to the degree reasonable. 

Worse case I think there'll be some required lock-step upgrades where both octoDNS and the modules need to be upgraded together. If version requirements aren't enough we could add some sort of "api" versioning to the providers/manager that looks at whether things worth together and warns, but I hope that isn't necessary.

ross:
Reworked the "supports" table to talk about extraction and show what has been extracted, linking to it. https://github.com/octodns/octodns-powerdns/pull/4 added support information for PowerDNS to the module repo.

I believe this is ready for a 👍 / 👎

README.md
```diff
@@ -218,6 +254,7 @@ The above command pulled the existing data out of Route53 and placed the results
 | [ZoneFileSource](/octodns/source/axfr.py) | | A, AAAA, CAA, CNAME, MX, NS, PTR, SPF, SRV, TXT | No | read-only |
 | [TinyDnsFileSource](/octodns/source/tinydns.py) | | A, CNAME, MX, NS, PTR | No | read-only |
 | [YamlProvider](/octodns/provider/yaml.py) | | All | Yes | config |
+>>>>>>> origin/master
```
yzguy, line 257: Looks like some merge conflict stuff hanging around here

ross: Doh, been so long since i resolved conflicts that I forgot how to do it :grin:

yzguy (approved):

README.md
```diff
@@ -186,38 +186,40 @@ $ octodns-dump --config-file=config/production.yaml --output-dir=tmp/ example.co
...
+| [EtcHostsProvider](/octodns/provider/etc_hosts.py) | | | A, AAAA, ALIAS, CNAME | No | |
+| [EnvVarSource](/octodns/source/envvar.py) | | | TXT | No | read-only environment variable injection |
+| [GandiProvider](/octodns/provider/gandi.py) | | | A, AAAA, ALIAS, CAA, CNAME, DNAME, MX, NS, PTR, SPF, SRV, SSHFP, TXT | No | |
+| [GCoreProvider](/octodns/provider/gcore.py) | | | A, AAAA, NS, MX, TXT, SRV, CNAME, PTR | Dynamic | |
+| [GoogleCloudProvider](/octodns/provider/googlecloud.py) | | google-cloud-dns | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, TXT  | No | |
+| [HetznerProvider](/octodns/provider/hetzner.py) | | | A, AAAA, CAA, CNAME, MX, NS, SRV, TXT | No | |
+| [MythicBeastsProvider](/octodns/provider/mythicbeasts.py) | | Mythic Beasts | A, AAAA, ALIAS, CNAME, MX, NS, SRV, SSHFP, CAA, TXT | No | |
+| [Ns1Provider](/octodns/provider/ns1.py) | | ns1-python | All | Yes | |
+| [OVH](/octodns/provider/ovh.py) | | ovh | A, AAAA, CAA, CNAME, MX, NAPTR, NS, PTR, SPF, SRV, SSHFP, TXT, DKIM | No | |
+| [PowerDnsProvider](https://github.com/octodns/octodns-powerdns/) | [octodns_powerdns](https://github.com/octodns/octodns-powerdns/) | | | | |
```
viranch, line 213: No big deal, but any reason for not choosing `octodns-powerdns` over the underscored version for the python package name? I think dashed name is more common?

ross: Dunno really. The actual module name can't have a `-` in it so I guess I just typed what that would be. Good news is that pypi finds it either way https://pypi.org/search/?q=octodns-powerdns & https://pypi.org/search/?q=octodns_powerdns and it actually looks like it changes it to a `-` in some of the spots. 

The repos actually use the `-`. I think it's probably something we can change if we want to. For now I'll keep doing the `_` and if we want we can go through and change things en mass at some point.

viranch (approved):
LGTM!

PR #828, opened 2021-12-31
Extract Dnsimple

ross:
This PR extracts DnsimpleProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-dnsimple. Most of the action is over in that PR. This one just stubs out the original core provider.

The provider has been extracted into https://github.com/octodns/octodns-powerdns without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-powerdns/.

/cc https://github.com/octodns/octodns-dnsimple


ross:
Start to finish this took probably 30m or so while watching YouTube videos :grin:. The https://github.com/octodns/octodns-template stuff works fairly well.

yzguy (approved):

PR #829, opened 2022-01-02
Update README to reflect CF URLFWD record support

SmithPeder:
In [this PR](https://github.com/octodns/octodns/pull/722) URLFWD record support was added to the NS1, YAML, and Cloudflare providers. All the correct `SUPPORTS` sets were updated, and [docs/records.md](https://github.com/octodns/octodns/blob/master/docs/records.md) added the `URLFWD` in its list of supported tags. 

For **NS1** and **YAML** the `README.md` table already has **Record Support = all**  so no changes were needed here, but for Cloudflare, the list was hardcoded, and URLFWD should have been added here. It's a very minor thing, but without looking at the code, it looks like Cloudflare URLFWDs are not supported.


yzguy:
Great catch! Thanks for the quick update.

brianeclow:
😓 Missed that.

PR #833, opened 2022-01-04
Extract & shim AzureProvider

ross:
This PR extracts AzureProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-azure. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-azure without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-azure/.

/cc https://github.com/octodns/octodns-azure


viranch (approved):
Sorry for the delay, finally had some time to test this in real world. Works like a charm!

ross:
> Sorry for the delay, finally had some time to test this in real world. Works like a charm!

I assumed as much. Thanks!

PR #834, opened 2022-01-04
Extract & Shim ConstellixProvider

ross:
This PR extracts ConstellixProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-constellix. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-constellix without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-constellix/.

/cc https://github.com/octodns/octodns-constellix
/cc @mintopia & @gartline as authors of the provider/dynamic support


yzguy (approved):
Looks good from basic record support from when I created it. Haven't worked with Dynamic records yet so would be good to get a +1 from at least one of the people you mentioned.

octodns/provider/constellix.py
```diff
@@ -5,1113 +5,19 @@
...
+from logging import getLogger
+
+logger = getLogger('Constellix')
+try:
+    logger.warn('octodns_constellix shimmed. Update your provider class to '
+                'octodns_constellix.ConstellixProvider. '
+                'Shim will be removed in 1.0')
+    from octodns_constellix import ConstellixProvider, ConstellixBaseProvider
+    ConstellixProvider  # pragma: no cover
+    ConstellixBaseProvider  # pragma: no cover
```
istr, line 17: @ross This class does not exist (yet?) in https://github.com/octodns/octodns-constellix/blob/main/octodns_constellix/__init__.py
I would tend to drop the reference here.

ross: Doh. Copy-n-paste error. Will get a PR going shortly.

ross: https://github.com/octodns/octodns/pull/843

PR #835, opened 2022-01-05
Extract and shim CloudflareProvider

ross:
This PR extracts CloudflareProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-cloudflare. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-cloudflare without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-cloudflare/.

/cc https://github.com/octodns/octodns-cloudflare
/cc @parkr as someone who uses the provider and has made some tweaks/improvements to it in the past


parkr:
@ross 👍 This looks good. From my previous experience splitting a big project up into multiple projects, it can be a bit tedious to manage many repos, but it looks like you have all the helpful scripts you need in `script/`, and you have GitHub Actions setup too.

I maintain [an octodns Docker image](https://github.com/parkr/dockerfiles/pkgs/container/octodns). I think my upgrade path here would be to specify versions of all the dependencies I want to include somewhere then start making a YYYY.MM.PATCH release for each dependency bump. Additionally, users will need to know to specify `class: octodns_yourprovider.YourProviderProvider` in their configs, right?

ross:
> @ross 👍 This looks good. From my previous experience splitting a big project up into multiple projects, it can be a bit tedious to manage many repos, but it looks like you have all the helpful scripts you need in `script/`, and you have GitHub Actions setup too.

Yeah. That's the hope, but I definitely expect it to be a bit more effort to manage everything. The main thing we'll gain is that each module will have it's own dependencies specified in setup.py so if you install `octodns-cloudflare` you'll be good to go and won't need to know what other stuff needs to get installed, `requests` in its case. That's more meaningful for things like `octodns-azure` where there's 3 azure modules that need to be installed, 2 of which are non-obvious identity and trafficmanager. 

Yeah, the hope is that `script/` and actions will make things a bit easier to work with since everything will be so similar. I also created https://github.com/octodns/octodns-template which is what all the extraction modules are based off of and it should also serve as a starting point for creating new providers.

> I maintain [an octodns Docker image](https://github.com/parkr/dockerfiles/pkgs/container/octodns). I think my upgrade path here would be to specify versions of all the dependencies I want to include somewhere then start making a YYYY.MM.PATCH release for each dependency bump. 

Yeah that seems logical. That does unfortunately leave you with a bunch of things to watch. There may be a way to use `pip` and some tooling to watch/test.

Looks like there's a handful of people doing docker containers with it :-/ https://hub.docker.com/search?q=octodns&type=image

Looks like yours is https://hub.docker.com/r/parkr/octodns. There a repo that's driven off of? Dunno if you're interested, but we could potentially house it in https://github.com/octodns/ and call it the official one if you're interested and think that'd make sense. Nbd either way.

> Additionally, users will need to know to specify `class: octodns_yourprovider.YourProviderProvider` in their configs, right?

Yeah. The shims that are in place for a while will spit out a message about the move so long as the old path is being used and try and shim the new one into place if it happens to be installed as well. If it's not installed the message is printed and things blow up, which is probably preferable since it'll push the change.

https://github.com/octodns/octodns/blob/355995daddb37e1286d88f5a92bb5be4736fa120/octodns/provider/cloudflare.py#L10-L22

Fwiw, `YamlProvider` will stay with octoDNS core, but atm everything else will be pulled out.

parkr:
> Looks like there's a handful of people doing docker containers with it :-/ https://hub.docker.com/search?q=octodns&type=image

A major bump with an obvious callout in the release notes should do the trick. 

> Looks like yours is https://hub.docker.com/r/parkr/octodns. There a repo that's driven off of? Dunno if you're interested, but we could potentially house it in https://github.com/octodns/ and call it the official one if you're interested and think that'd make sense. Nbd either way.

It's generated from https://GitHub.com/parkr/dockerfiles. Happy to contribute to an official docker image. I'd make octodns/octodns:latest include all common modules then make slimmer versions for some common module groupings.

> The shims that are in place for a while will spit out a message about the move so long as the old path is being used and try and shim the new one into place if it happens to be installed as well. If it's not installed the message is printed and things blow up, which is probably preferable since it'll push the change.

That sounds good 👍 The one thing I'd suggest is adding a URL to explain the migration path in a bit more detail to the error message. The message you linked to is a little vague. Since it's coming from a shim strictly for migration purposes, you could add an anchor link to this repo's readme explaining all the pieces for the transition.

parkr:
Shim works like a charm: https://github.com/parkr/dns/pull/23#issuecomment-1007008191

:shipit:

ross:
> It's generated from https://GitHub.com/parkr/dockerfiles. Happy to contribute to an official docker image. I'd make octodns/octodns:latest include all common modules then make slimmer versions for some common module groupings.

👍. Fwiw I don't imagine the size difference between everything and specific things would be all that meaningful. 

I've always avoided including everything in octoDNS's setup.py mostly b/c it'd make tracking dependency updates annoying/super frequent when you'd get a lot of things you didn't care about. I guess that might apply to docker, but since people wouldn't have to worry about it directly it's probably not as bad.

> That sounds good 👍 The one thing I'd suggest is adding a URL to explain the migration path in a bit more detail to the error message. The message you linked to is a little vague. Since it's coming from a shim strictly for migration purposes, you could add an anchor link to this repo's readme explaining all the pieces for the transition.

Makes sense. I'll put something like that on the TODO list.

PR #838, opened 2022-01-06
Extract & shim DynProvider

ross:
This PR extracts DynProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-dyn. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-dyn without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-dyn/.

/cc https://github.com/octodns/octodns-dyn


ross:
This provider is deprecated anyway, so turbo-merging it.

PR #850, opened 2022-01-13
Quell all python warnings, mostly *Equals -> Equal and warn -> warning

ross:
This is a step in preparation for moving to pytest which captures and emits warnings. We had TONS of them and this will get rid of them before we go there. A good thing to do in general, but pretty much required to be able to cope with pytest (without disabling them)


viranch (approved):
Simple enough, good to go after resolving the conflicts.

PR #852, opened 2022-01-13
Add python 3.10 to actions matrix now that we're off nose

ross:


ross:
Oh, that's interesting. OVH's vendored requests isn't python 3.10 compatible. It hasn't seen a release since 2018 😞. This will soon be fixed when that provider is extracted so Will just hold off on this PR until after that. That repo can then only support < 3.10, which is part of the point of breaking things up.

viranch (approved):

PR #855, opened 2022-01-14
Extract RackspaceProvider from octoDNS core

ross:
This PR extracts RackspaceProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-rackspace. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-rackspace without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-rackspace/.

/cc https://github.com/octodns/octodns-rackspace


octodns/provider/ovh.py
```diff
@@ -9,9 +9,9 @@
 
 logger = getLogger('Ovh')
 try:
-    logger.warn('octodns_ovh shimmed. Update your provider class to '
-                'octodns_ovh.OvhProvider. '
-                'Shim will be removed in 1.0')
+    logger.warning('octodns_ovh shimmed. Update your provider class to '
+                   'octodns_ovh.OvhProvider. '
+                   'Shim will be removed in 1.0')
```
viranch, line 14: no big deal, but was this intentional for this PR?

tests/test_octodns_record.py
```diff
@@ -1512,6 +1512,9 @@ def test_mx_value(self):
         self.assertTrue(c >= c)
         self.assertTrue(c <= c)
 
+        self.assertEqual(a.__hash__(), a.__hash__())
+        self.assertNotEqual(a.__hash__(), b.__hash__())
+
```
viranch, line 1517: what's up here? 🤔

viranch (approved):
Overall good to go.

octodns/provider/ovh.py
ross: Tbh, not sure how it made it through the ovh PR without getting noticed. Maybe warnings are errors. Will look into that. It was just in here to avoid a dedicated PR since I've been doing so many of them already

tests/test_octodns_record.py
ross: This is getting some reasonable coverage for `MxValue.__hash__()`. Apparently RackspaceProvider was somehow hitting those code paths and nothing else was. I have a note to look into what's up there, but whatever it is it seemed out of scope for removing the provider.

octodns/provider/ovh.py
viranch: Yeah those are warnings not errors.

tests/test_octodns_record.py
istr: Just a suspicion:
```
    def test_apply_create_MX(self):
        class TestData(object):
            OtherRecords = [
                {
                    "subdomain": '',
                    "data": {
                        'type': 'MX',
                        'ttl': 300,
                        'value': {
                            'value': 'mail1.example.com.',
                            'priority': 1,
                        }
                    }
                },
```
(in `test_octodns_provider_rackspace.py`, l. 342 ff.) hits

```
        # UNTIL 1.0 remove value fallback
        try:
            exchange = value['exchange']
        except KeyError:
            exchange = value['value']
        self.exchange = exchange.lower()
```
(in `record/__init__.py`, l. 1111 ff.).

istr:
@ross btw kudos for the mass of PRs you lift as a result of the module split...

PR #857, opened 2022-01-14
Use proper provider names and product links in README

viranch:
Nice to have actual provider names with product links. Module links still exist in the table.

Preview: https://github.com/octodns/octodns/tree/readme-provider-links#providers


yzguy:
Yeah this is a nice UX change, good call.

PR #859, opened 2022-01-17
Extract ultra

ross:
This PR extracts UltraProvider from octoDNS core into a dedicated repo/module, https://github.com/octodns/octodns-ultra. Most of the action is over in that repo. This one just stubs out the original core provider & processor.

The provider has been extracted into https://github.com/octodns/octodns-ultra without any meaningful changes other than packaging. There's an initial release pushed up to pypi https://pypi.org/project/octodns-ultra/.

/cc https://github.com/octodns/octodns-ultra


tests/test_octodns_provider_base.py
```diff
@@ -167,6 +167,29 @@ def test_plan(self):
...
+            'ttl': 30,
+            'type': 'AAAA',
+            'value': '2601:644:500:e210:62f8:1dff:feb8:947a',
+        })
+        zone.add_record(not_supported)
+        provider = HelperProvider()
+        plan = provider.plan(zone)
+        self.assertTrue(plan)
+        self.assertEqual(1, len(plan.changes))
+        self.assertEqual(supported, plan.changes[0].new)
```
ross, line 191: Case that was previously tested by the actual providers and no longer covered now the last one is gone so explcit test added.

viranch (approved):

viranch:
This seems to be the last provider? Great work on these PRs! I like the segregation.

ross:
> This seems to be the last provider? Great work on these PRs! I like the segregation.

Yup. Dropped an update in https://github.com/octodns/octodns/issues/622#issuecomment-1014914555 to that effect.

PR #863, opened 2022-01-26
Link all of the extration help messages to the README section

ross:
Some of the later stuff already linked to this. This goes through and updates the messages for all the earlier ones per @parkr's suggestion.

### DIFF
--- a/octodns/provider/azuredns.py
+++ b/octodns/provider/azuredns.py
@@ -17,5 +17,7 @@ try:
 except ModuleNotFoundError:
     logger.exception('AzureProvider has been moved into a seperate module, '
                      'octodns_azure is now required. Provider class should '
-                     'be updated to octodns_azure.AzureProvider')
+                     'be updated to octodns_azure.AzureProvider. See '
+                     'https://github.com/octodns/octodns/README.md#updating-'
+                     'to-use-extracted-providers for more information.')
     raise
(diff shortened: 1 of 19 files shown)


octodns/provider/azuredns.py
```diff
@@ -17,5 +17,7 @@
 except ModuleNotFoundError:
     logger.exception('AzureProvider has been moved into a seperate module, '
                      'octodns_azure is now required. Provider class should '
-                     'be updated to octodns_azure.AzureProvider')
+                     'be updated to octodns_azure.AzureProvider. See '
+                     'https://github.com/octodns/octodns/README.md#updating-'
```
parkr, line 21:
```suggestion
                     'https://github.com/octodns/octodns#updating-'
```

`/README.md` isn't a valid GitHub file path. It has to be `/blob/<ref>/README.md` or since the root renders the readme, it can be just at the root of the repo:

https://github.com/octodns/octodns/#updating-to-use-extracted-providers

parkr:
@ross I believe every link you added here needs to be updated. I only submitted my comment for 1 link but the others need to be updated too.

ross:
> @ross I believe every link you added here needs to be updated. I only submitted my comment for 1 link but the others need to be updated too.

Realized that right after I clicked the button. Follow on PR to fix everything coming shortly.

parkr:
😄

PR #866, opened 2022-01-27
Initial pass at script/update-requirements, results of run

ross:
This seems to work well enough with the way octodns's setup.py's work. Assuming everyone is 👍 with this approach I plan to add the script to octodns-template and all of the providers along with similar updates for them to get everything on the newest stuff.


ross:
I think I've found a better way to go about this letting setup.py do most/all of the work itself. Going to go back to draft while I play around with that approach.

ross:
I like this better. No more parsing of setup.py. It does require using `extras_require`, but it seems cleaner to do that and include the maintainer/dev bits in there than in the `update-requirements` script.

setup.py
```diff
@@ -55,33 +55,43 @@ def long_description():
...
+            'twine>=3.4.2',
+        ),
+    },
+    install_requires=(
         'PyYaml>=4.2b1',
         'dnspython>=1.15.0',
         'fqdn>=1.5.0',
         'natsort>=5.5.0',
         'pycountry>=19.8.18',
         'pycountry-convert>=0.7.2',
```
viranch, line 85: From what I can tell, these are not runtime requirements right? They're only used in `script/generate-geo-data` which I think is not used within an installed package?

viranch (approved):
Looks fine other than the minor comment/question.

setup.py
ross: Will double check they're not needed by anything else and move to the dev side of things if not.

ross: Wow. good call on that one. pycountry-convert has a dumb requirements setup that's including a lot of crap that are really test-time requirements etc. https://gist.github.com/ross/61121ab47414fc2334fa1ed480fc3e93 (the pypi linked repo is no more/has moved so gistified it)

Moving it brings requirements.txt down to 7 items https://github.com/octodns/octodns/pull/866/commits/49535fa1045ae18f141d52109998706b690a4a0f

PR #868, opened 2022-01-27
Fully vet the setup.py requirements and installed code during cibuild

ross:
This builds off of https://github.com/octodns/octodns/pull/866 as it needs the `extras_require.dev` stuff in there to know what to install during tests. 

This runs `setup.py install` and then `octodns-sync --help` to generally vet things and ensure the commands are installed.

It then runs `pytest --no-network` against the tmp venv which should run the tests against the installed code (no `PYTHONPATH=.`

If this works out something similar will be created for octodns-template and all the provider repos.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -37,5 +37,5 @@ python setup.py install
 octodns-sync --help
 echo "## validate tests can run against installed code ###############################"
 pip install pytest pytest-network
-pytest
+pytest --disable-network
 echo "## complete ####################################################################"
--- a/script/cibuild
+++ b/script/cibuild
@@ -37,6 +37,5 @@ python setup.py install
 octodns-sync --help
 echo "## validate tests can run against installed code ###############################"
 pip install .[dev]
-cd $TMP_DIR
 pytest --disable-network
 echo "## complete ####################################################################"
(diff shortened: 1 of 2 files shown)


ross:
So this works now. Will cause things to take quite a bit longer to run, but that probably doesn't matter too much.

script/cibuild
```diff
@@ -27,4 +28,14 @@ echo "## tests/coverage ########################################################
...
+echo "## validate setup.py install ###################################################"
+deactivate
+TMP_DIR=$(mktemp -d -t ci-XXXXXXXXXX)
+python3 -m venv $TMP_DIR
+. "$TMP_DIR/bin/activate"
+python setup.py install
+octodns-sync --help
+echo "## validate tests can run against installed code ###############################"
+pip install pytest pytest-network
+pytest
```
viranch, line 40: Did you want this to be `pytest --no-network` per the PR description?

Also looks like we're running coverage test as well as a separate pytest. Any way we could run the full coverage test suite just once on the installed code to speed things up? May be even put the install+test approach in `script/coverage` itself? That way the pre commit hook will also work more similarly as the CI build.

ross: Yup. I originally was trying other things to install the test requirements and had to remove the no-network temporarily and forgot to put it back. Will do.

viranch (approved):

ross:
This adds about 15s give or take a few seconds to the existing CI time. Doesn't seem like the end of the world and it fully vets that the installed package is functional so that seems like a win.

PR #869, opened 2022-02-01
pycountry is only a dev requirement in core now

ross:
Some providers will require it, but the core library does not at runtime. PRs for those providers will be created shortly. 

/cc https://github.com/octodns/octodns-azure/pull/18#pullrequestreview-868736395


viranch (approved):

ross:
Going to cut new releases of the 3 modules that directly need pycountry-convert before I merge this.

Issue #600, opened 2020-08-31
Dry run didn't catch improperly formatted MX record, broke with --doit


zachlatta:
Hi there, we use OctoDNS to manage [Hack Club](https://hackclub.com)'s DNS records and it's amazing! It means we can open source all of our DNS records at [`hackclub/dns`](https://github.com/hackclub/dns). I love the idea of "infrastructure as code" and OctoDNS helps us a lot.

We recently ran into an issue where someone added an MX record with the `exchange` value `100 mx.sendgrid.net.`. We have a test suite that does a dry run, which didn't surface any issues with this value, but when it was ran with `--doit`, we got an ambiguous error.

I was able to figure out that we had to remove `100 ` from the record value, which fixed `--doit`. Setting `exchange` to just `mx.sendgrid.net.` worked fine, but as a user I would expect OctoDNS to surface this error during the dry run.

Here's the error log:

```
...(many lines above this)...
*   Create <MxRecord MX 1, front-mail.hackclub.com., [''1 100 mx.sendgrid.net.'']> (config)
*   Create <TxtRecord TXT 1, front-mail.hackclub.com., ['v=spf1 include:sendgrid.net ~all']> (config)
*   Summary: Creates=3, Updates=1, Deletes=0, Existing Records=184
********************************************************************************


2020-08-30T23:28:15  [140229093717632] INFO  DnsimpleProvider[dnsimple] apply: making changes
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/2.7.18/x64/bin/octodns-sync", line 11, in <module>
    load_entry_point('octodns==0.9.10', 'console_scripts', 'octodns-sync')()
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/octodns/cmds/sync.py", line 39, in main
    dry_run=not args.doit, force=args.force)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/octodns/manager.py", line 344, in sync
    total_changes += target.apply(plan)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/octodns/provider/base.py", line 95, in apply
    self._apply(plan)
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/octodns/provider/dnsimple.py", line 416, in _apply
    getattr(self, '_apply_{}'.format(class_name))(change)
[...]

PR #873, opened 2022-02-04
Validate record fields that should hold FQDNs w/tests

ross:
Add FQDN validation to ALIAS/CNAME value, MX exchange, SRV target and tests of the functionality.

/cc Fixes https://github.com/octodns/octodns/issues/600#issuecomment-683837402 @zachlatta

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -1091,7 +1091,10 @@ class MxValue(EqualityTupleMixin):
             exchange = None
             try:
                 exchange = value.get('exchange', None) or value['value']
-                if not exchange.endswith('.'):
+                if not FQDN(str(exchange), allow_underscores=True).is_valid:
+                    reasons.append(f'Invalid MX exchange "{exchange}" is not '
+                                   'a valid FQDN.')
+                elif not exchange.endswith('.'):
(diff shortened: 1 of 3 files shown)


octodns/record/__init__.py
```diff
@@ -1091,7 +1091,10 @@ def validate(cls, data, _type):
             exchange = None
             try:
                 exchange = value.get('exchange', None) or value['value']
```
viranch, line 1093: Do we also want to do type checking or casting to str here? If there's a non-str, the else condition will throw an exception.

ross: Good question. I just copy-n-pasted the existing `_TargetValue` code block which does the `str`. 

Coming from Yaml I think the only other types that could happen are: int, float, list, or dict. All of those would stringify cleanly and fail the `FQDN` check so I think this might be OK as-is, though maybe not an ideal error message. 

You see any other cases?

There's probably a TON of places we could do better error messages for unexpected types. Not opposed to it here if the existing invalid FQDN message isn't sufficient.

viranch: Good point, I think the FQDN error will suffice.

viranch (approved):

Issue #38, opened 2017-05-24
Zone NS records


vanbroup:
After dumping a zone from Cloudflare or Route53 (did not test any other providers) the YAML zone doesn't contain any zone NS records, where you would want to keep the main NS records in sync. I suppose these zone NS records are currently ignored and need to be configured manually at each provider but should this not be something that OctoDNS keeps in sync?

```
; <<>> DiG 9.10.3-P4-Ubuntu <<>> github.com NS
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44915
;; flags: qr rd ra; QUERY: 1, ANSWER: 8, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;github.com.                    IN      NS

;; ANSWER SECTION:
github.com.             28      IN      NS      ns-1283.awsdns-32.org.
github.com.             28      IN      NS      ns-1707.awsdns-21.co.uk.
github.com.             28      IN      NS      ns-421.awsdns-52.com.
github.com.             28      IN      NS      ns-520.awsdns-01.net.
github.com.             28      IN      NS      ns1.p16.dynect.net.
github.com.             28      IN      NS      ns2.p16.dynect.net.
github.com.             28      IN      NS      ns3.p16.dynect.net.
github.com.             28      IN      NS      ns4.p16.dynect.net.

;; Query time: 4 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Wed May 24 10:26:42 UTC 2017
;; MSG SIZE  rcvd: 259
```

ross:
> I suppose these zone NS records are currently ignored 

Correct they are ignored for all of the current providers with the exception of `PowerDnsProvider` which has the optional ability to manage them based with [provider config supplied values](https://github.com/github/octodns/blob/2ff34997a5df83ec8aba4edd7a6d4bd33e553a68/octodns/provider/powerdns.py#L351-L355).

> need to be configured manually at each provider but should this not be something that OctoDNS keeps in sync?

The reason behind the current state is they're a bit more complicated to reason about. As mentioned [in the blog post about octodns](https://githubengineering.com/enabling-split-authority-dns-with-octodns/) we use the above two providers actively, but push our data into further providers, both external and internal. 

[...]

PR #875, opened 2022-02-08
Rework _AggregateTarget to dynamically handle SUPPORTS

ross:
Was exploring our options for adding root `NS` record support and as part of expirimenting with that I ran into the fact that `_AggregateTarget` didn't support a bunch of the newer `SUPPORTS_*` properties we've added. This reworks that class to dynamically support them so that all current and future such properties will be supported. It also adds some more testing around the behavior of that class.

/cc https://github.com/octodns/octodns/issues/38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
-## v0.9.15 - 202?-??-?? - ??
+## v0.9.16 - 2022-??-?? - ???
+
+#### Stuff
+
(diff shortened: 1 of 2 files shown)


viranch (approved):
This is a new territory for me, PR looks ok after getting a vague sense of `_AggregateTarget`, which seems to be used only for `octodns-compare`.

Issue #879, opened 2022-02-15
Null MX records throw ValidationError


viranch:
FQDN validation for MX records introduced in #873 has started causing `ValidationError`s on [null MX records](https://datatracker.ietf.org/doc/html/rfc7505).

```
octodns.record.ValidationError: Invalid record test.com.
  - Invalid MX exchange "." is not a valid FQDN.
```

Example null MX record:

```
'':
  - type: MX
    ttl: 86400
    values:
    - exchange: .
      preference: 0
```

RFC also notes that `.` is not a valid FQDN and should be considered a special case.

PR #880, opened 2022-02-15
Fix handling of null MX records and add test for behavior

ross:
/cc Fixes https://github.com/octodns/octodns/issues/879 @viranch


viranch (approved):
Good to go. At some point, we can consider adding more validations on null mx, like preference should be zero and no other mx should exist.

PR #889, opened 2022-03-16
Dynamic Record type registration

ross:
Support for dynamically registering new record types in octoDNS.  This should allow providers to add special record types that only they support without polluting the core of octoDNS.

* [x] Full testing of registration including usage
* [x] Document requirements/behavior, specifically naming conventions
   * ~~ROUTE53-ALIAS?~~
   * ~~SYMLINK (better name for Route53's ALIAS, claiming top-level names/free for all)~~
   * Route53Provider/ALIAS

/cc https://github.com/octodns/octodns-route53/issues/24 for a potential use

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -90,7 +90,7 @@ class Record(EqualityTupleMixin):
     def register_type(cls, _class, _type=None):
         if _type is None:
             _type = _class._type
-        existing = cls._CLASSES.get(_type, None)
+        existing = cls._CLASSES.get(_type)
         if existing:
             module = existing.__module__
             name = existing.__name__
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -87,7 +87,9 @@ class Record(EqualityTupleMixin):
     _CLASSES = {}
 
     @classmethod
-    def register_type(cls, _type, _class):
+    def register_type(cls, _class, _type=None):
+        if _type is None:
+            _type = _class._type
         existing = cls._CLASSES.get(_type, None)
         if existing:
             module = existing.__module__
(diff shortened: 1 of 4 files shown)


ross:
So this is dumb, ran all the module jobs against this branch over in https://github.com/octodns/octodns/actions/runs/2012856277 but they didn't report their status here. Will poke around a bit more, but maybe this approach isn't going to work (if there's no way to manually run them.) I guess worse case we can just start running them all the time along side the regular CI ☹️

ross:
This is ready for review and generally should be independent of https://github.com/octodns/octodns-route53/pull/25 as I think the dynamic registration works fine for the built-in stuff even if we don't use it for anything (yet.)

octodns/record/__init__.py
```diff
@@ -80,6 +84,20 @@ def __init__(self, fqdn, reasons):
 class Record(EqualityTupleMixin):
     log = getLogger('Record')
 
+    _CLASSES = {}
+
+    @classmethod
+    def register_type(cls, _class, _type=None):
+        if _type is None:
+            _type = _class._type
+        existing = cls._CLASSES.get(_type, None)
```
viranch, line 93: Minor cosmetics: any value in explicitly setting default fallback to `None`?

viranch (approved):
Pretty reasonable, ship!

octodns/record/__init__.py
ross: If i follow the question, the `None` default prevents this line from throwing a `KeyError` exception when the class isn't already registered. This could have been done with a try/except if that's preferable, e.g.

```python
        try:
            existing = cls._CLASSES[_type]
            module = existing.__module__
            name = existing.__name__
            msg = f'Type "{_type}" already registered by {module}.{name}'
            raise RecordException(msg)
        except KeyError:
            cls._CLASSES[_type] = _class
```

viranch: Oh no I meant `.get(_type, None)` vs just `.get(_type)`

ross: Ah, maybe I started out with `False` in there or something. I wasn't even seeing that bit.

viranch: Like I said, minor cosmetic, no big deal

PR #890, opened 2022-03-21
Run the modules ci jobs when reviews are submitted

ross:
Looking for a better workflow here. Will likely review the required bit from these jobs so they don't show up/get in the way until we're close to the end (review submitted.)


ross (commented):
Testing.

ross:
Works as hoped/expected in conjunction with https://github.com/octodns/terraformed/pull/9

PR #891, opened 2022-03-26
Log the octoDNS version as part of Manager.__init__ info line

ross:
Useful for reporting problems and just generally being able to quickly see what's being used.

/cc https://github.com/octodns/octodns-docker/issues/27#issuecomment-1079544376 which lead me to make this addition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
   to do so, but weren't. There was an ordering before, but it was essentially
   arbitrarily picked.
 
+#### Stuff
+
+* Manager includes the octoDNS version in its init log line
+
 ## v0.9.16 - 2022-03-04 - Manage the root of the problem
 
 #### Noteworthy changes
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -119,9 +119,8 @@ class Manager(object):
             kwargs = self._build_kwargs(provider_config)
             try:
                 self.providers[provider_name] = _class(provider_name, **kwargs)
-                if not module.startswith('octodns.'):
-                    self.log.info('__init__: provider=%s (%s %s)',
-                                  provider_name, module, version)
+                self.log.info('__init__: provider=%s (%s %s)', provider_name,
+                              module, version)
             except TypeError:
                 self.log.exception('Invalid provider config')
                 raise ManagerException('Incorrect provider config for ' +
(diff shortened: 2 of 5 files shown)


ross:
This issue will be linked to by PRs to all the modules that print their version.

ross:
Reworked this to have Manager try and find and print versions in a standard way for modules so that each and every module doesn't have to add code to do so.

```
2022-03-25T19:47:38  [4501360128] INFO  Manager __init__: provider=ns1 (octodns_ns1 0.0.2)
2022-03-25T19:47:38  [4501360128] INFO  Manager __init__: provider=route53 (octodns_route53 0.0.4)
```

ross:
Reworked to walk the hierarchy looking for __VERSION__. This should work better for things that aren't in the top-level modules (common for processors etc.) 

```
2022-03-25T20:20:06  [4648537600] INFO  Manager __init__: provider=config (octodns.provider.yaml 0.9.16)
2022-03-25T20:20:06  [4648537600] INFO  Manager __init__: provider=ns1 (octodns_ns1 0.0.2)
2022-03-25T20:20:07  [4648537600] INFO  Manager __init__: provider=route53 (octodns_route53 0.0.4)
2022-03-25T20:20:07  [4648537600] INFO  Manager __init__: processor=acme (octodns.processor.acme 0.9.16)
```

I think this is good to go.

viranch:
If possible, it would be nice to know if a package was installed using the commit SHA instead of a released version.

ross:
> If possible, it would be nice to know if a package was installed using the commit SHA instead of a released version.

Yeah. I was thinking about that, but didn't have any great options off the top of my head. I'll poke around and see if I can find some standard/clean solution to that.

ross:
As of https://github.com/octodns/octodns/pull/892 we now have 8 chars of sha included when modules are installed from git added to the end of the version string, e.g. `x.y.z+abcdef12`

ross:
> As of #892 we now have 8 chars of sha included when modules are installed from git added to the end of the version string, e.g. `x.y.z+abcdef12`

Doh. So didn't really occur to me that the `<module>.__VERSION__` that this PR is looking at isn't going to be changed by the install. There is an importlib helper that gets the actually installed version:

```python
from importlib.metadata import version
version('octodns_route53')
```

Going to poke around and see what that looks like.

ross:
Got something workable, though `importlib.metadata` isn't available in py3.7 so there's a noop fallback in place for that version for now. I think this is gtg.

yzguy (approved):

yzguy:
Yeah this will definitely be helpful when issues come up, perhaps a Issue template that requests the OctoDNS version, python version, etc when created.

ross:
> Yeah this will definitely be helpful when issues come up, perhaps a Issue template that requests the OctoDNS version, python version, etc when created.

Makes sense. I also want to look at a `--version` flag or something similar to just print out this info

PR #892, opened 2022-03-27
Implement pep440 style public and local version numbers

ross:
This implements [pep440](https://peps.python.org/pep-0440/) style version numbers. 

True/public release will continue to have the `<x>.<y>.<z>` style numbering, controlled by an environmental variable `OCTODNS_RELEASE` being set when doing the install/builds. 

`0.9.16`

Local releases, installing from a git checkout etc., will use the same version number appending the first 8 characters of the git sha. 

`0.9.16.b4d66c3f`

This should differentiate publicly released builds from those done from a specific SHA, either locally or with managers such as pip. 

If we're going to go this route we'll want to add this same logic to all of the provider modules and the template. 

/cc https://github.com/octodns/octodns/pull/891#issuecomment-1079611056 for the suggestion @viranch

### DIFF
--- a/script/release
+++ b/script/release
@@ -16,6 +16,9 @@ if [ ! -f "$ACTIVATE" ]; then
 fi
 . "$ACTIVATE"
 
+# Set so that setup.py will create a public release style version number
+export OCTODNS_RELEASE=1
+
 VERSION="$(grep __VERSION__ "$ROOT/octodns/__init__.py" | sed -e "s/.* = '//" -e "s/'$//")"
 
 git tag -s "v$VERSION" -m "Release $VERSION"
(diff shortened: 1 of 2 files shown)


ross:
```
...
creating 'dist/octodns-0.9.16+cce5f7ac-py3.10.egg' and adding 'build/bdist.linux-x86_64/egg' to it
removing 'build/bdist.linux-x86_64/egg' (and everything under it)
Processing octodns-0.9.16+cce5f7ac-py3.10.egg
Copying octodns-0.9.16+cce5f7ac-py3.10.egg to /tmp/ci-6QNIOWzx9C/lib/python3.10/site-packages
Adding octodns 0.9.16+cce5f7ac to easy-install.pth file
...
```

viranch (approved):
Nice trick! Ship it

Issue #897, opened 2022-04-11
octodns-dump does not respect YamlProvider configuration


viranch:
I prefer to not skip defaults (like TTL) in the yamls when dumping. So as a work-around, I set it to an odd unlikely number in the config, but I noticed octodns-dump command does not really respect the provider config:

```yaml
providers:
  cfg:
    class: octodns.provider.yaml.YamlProvider
    directory: ./src/dev/test
    default_ttl: 9
    enforce_order: False
```

I think this line here doesn't take any provider configs into account https://github.com/octodns/octodns/blob/master/octodns/manager.py#L570.

May be we add the ability to set these provider configs as CLI switches since there can be multiple YamlProvider configs in the config file and dump would not know which one to pick. Or provide one switch to disable omitting default values.

ross:
Yeah, issue here is that it creates a `YamlProvider` on the fly to do the dump and there's no way to configure the details of it.

Quick work-around would be to do

```yaml
zones:
  whatever.com.:
      sources:
      - something
      targets:
      - config
```

Broader solution will probably involve improving dump so that it can use a configured provider, quick thoughts on that:
- it could allow you to specify a provider id/name and then verify that it's a YamlProvider and use it
- `--output-dir` could still specify where the dump goes to, but we'd probably need to duplicate the provider before changing its `output_dir`
- it could potentially search for a YamProvider to use, but that seems a bit dangerous in terms of accidentially overwriting files without any directive to do so

viranch:
Doing a sync with `source` as provider and `target` as config results in exception if there are non-parsable dynamic records, unlike the dump command:

```
2022-04-13T12:56:14  [139698595239680] WARNING Ns1Provider[ns1_test] Cannot parse www.example.com dynamic record due to missing pool name in first answer note, treating it as an empty record
Traceback (most recent call last):
  File "/home/viranch/code/playground/dynamic-dns/venv/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/home/viranch/code/playground/dynamic-dns/venv/lib/python3.8/site-packages/octodns/cmds/sync.py", line 38, in main
    manager.sync(eligible_zones=args.zone, eligible_sources=args.source,
  File "/home/viranch/code/playground/dynamic-dns/venv/lib/python3.8/site-packages/octodns/manager.py", line 413, in sync
    ps, d = future.result()
  File "/export/apps/python/3.8/lib/python3.8/concurrent/futures/_base.py", line 444, in result
[...]

PR #898, opened 2022-04-14
Implement a sketch of --output-provider support for dump

ross:
Allow the dump provider to be specified on the command line.

Requires that the configured provider support `copy()` and `directory`. This requirement is to ensure that we don't modify a provider that's being used as a source, e.g. if someone is trying to dump from yaml to yaml (with different provider settings.) 

/cc https://github.com/octodns/octodns/issues/897 @viranch


octodns/manager.py
```diff
@@ -549,25 +549,46 @@ def compare(self, a, b, zone):
...
+            if not hasattr(target, 'directory'):
+                msg = f'output_provider={output_provider}, does not support ' \
+                    'directory property'
+                raise ManagerException(msg)
+            if target.directory != output_dir:
+                if not hasattr(target, 'copy'):
+                    msg = f'output_provider={output_provider}, does not ' \
+                        'support copy method'
+                    raise ManagerException(msg)
+                target = target.copy()
```
viranch, line 582: It will be helpful to have a comment here (or another better place) that explains the need for this `copy` functionality.

octodns/manager.py
```diff
@@ -549,25 +549,46 @@ def compare(self, a, b, zone):
...
+                self.log.info('dump: setting directory of output_provider '
+                              'copy to %s', output_dir)
+                target.directory = output_dir
+        else:
+            self.log.info('dump: using custom YamlProvider')
+            clz = YamlProvider
+            if split:
+                clz = SplitYamlProvider
```
viranch, line 590: I'm not sure what this splitting does, but do we need to support when an output provider is specified?

ross: Strictly speaking no people could now configure a splitting provider to use in dumps, but it's probably best to leave it in place in case anyone is using it. We could deprecate it, I suppose.

viranch:
Looks like this is good to go other than the merge conflict?

PR #901, opened 2022-05-10
Feature: TLSA record

Aquifoliales:
Hi,
I tried to add TLSA record to the supported record types (https://datatracker.ietf.org/doc/html/rfc6698). It is my first time writing test cases, so I hope they suit well.  I do not get a coverage of 100% since I am not aware how to test 
`if not isinstance(data, (list, tuple)):`
line 1524, in octodns/record/\_\_init\_\_.py.
If you could point me to the right direction I will get it fixed.

Thanks for your work!

### DIFF
--- a/docs/records.md
+++ b/docs/records.md
@@ -18,6 +18,7 @@ OctoDNS supports the following record types:
 * `SPF`
 * `SRV`
 * `SSHFP`
+* `TLSA`
 * `TXT`
 * `URLFWD`
 
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -1517,6 +1517,71 @@ class SrvRecord(ValuesMixin, Record):
 Record.register_type(SrvRecord)
 
 
+class TlsaValue(EqualityTupleMixin):
+
+    @classmethod
+    def validate(cls, data, _type):
+        if not isinstance(data, (list, tuple)):
+            data = (data,)
+        reasons = []
+        for value in data:
+            try:
+                certificate_usage = int(value.get('certificate_usage', 0))
+                if certificate_usage < 0 or certificate_usage > 3:
+                    reasons.append(f'invalid certificate_usage "{certificate_usage}"')
+            except ValueError:
+                reasons.append(f'invalid certificate_usage "{value["certificate_usage"]}"')
+
+            try:
+                selector = int(value.get('selector', 0))
+                if selector < 0 or selector > 1:
+                    reasons.append(f'invalid selector "{selector}"')
+            except ValueError:
+                reasons.append(f'invalid selector "{value["selector"]}"')
+
+            try:
+                matching_type = int(value.get('matching_type', 0))
+                if matching_type < 0 or matching_type > 2:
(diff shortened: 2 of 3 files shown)


ross:
Looking at the coverage details it looks like all the `data` being passed to `validate` are not lists or tuples so that means a test case somewhere needs to pass in multiple values to validate. 

Screen Shot 2022-05-10 at 9 13 41 AM

Looking in https://github.com/octodns/octodns/pull/901/files#diff-034a576720d85caf6c8d90b613e218b24f59cd773cd0401992cc1dc6470340c5R3273, which is the section that's testing out the validate function in detail when called by `Record.new`, everything there is using `value`. It should be easy enough to add another test case or modify an existing one to use `values` (plural) and use and array of dictionaries. 

```python
                'values': [{
                    # first value is valid and will effectively be ignored
                    'certificate_usage': 0,
                    'selector': 0,
                    'matching_type': 0,
                    'certificate_association_data': 'AAAAAAAAAAAAA'
                }, {
                    # second is missing certificate_association_data
                    'certificate_usage': 0,
                    'selector': 0,
                    'matching_type': 0
                }]
```

ross (approved):
Looks good. Thanks for sorting the lint & test coverage. 

Note that this will add support for the type to the core of octoDNS so that other providers can no start to use the type so PRs to those providers will be required to manage the type. PowerDNS is likely trivial assuming it supports the type and might be a decent place to start, but feel free to start with the provider(s) you plan to use if you prefer.

Aquifoliales:
I added support for TLSA records to [ZoneFileProvider](https://github.com/kompetenzbolzen/octodns-custom-provider), and it works like a charm.

One further question:
Sometimes within `octodns/record/__init__.py` values are checked against predefined tuples, e.g.
```
class SshfpValue(EqualityTupleMixin):
    VALID_ALGORITHMS = (1, 2, 3, 4)
    VALID_FINGERPRINT_TYPES = (1, 2)
[...]
                if fingerprint_type not in cls.VALID_FINGERPRINT_TYPES:
```
but sometimes, like I have done it, it is checked like this:
`if certificate_usage < 0 or certificate_usage > 3:`

Is one of the solutions better than the other, and why though?

ross:
> I added support for TLSA records to [ZoneFileProvider](https://github.com/kompetenzbolzen/octodns-custom-provider), and it works like a charm.

Cool. Hadn't seen that/those providers.

> Is one of the solutions better than the other, and why though?

The `not in` stuff would be better/required if the values aren't contiguous. E.g. if they were `(1, 2, 5, 42)` or something of that sort, since the >/</=  clauses for that would be ugly and only get worse over time if new values were added. 

The if </> are likely more performant if they were used in a situation where that mattered (it generally does not with octoDNS.) Main downside of this route is that if the possible values are updated the checks have to be modified to match.

In that code it's probably just whatever the person who implemented the thing thought to do when they were writing it out. Ideally it'd be consistent, but it's not the end of the world. I'd lean a little bit towards the `not in` route now that I'm sitting and thinking about it since it's a little more future proof, but I don't think it's critical regardless.

Issue #199, opened 2018-02-18
Support for IDN


ad-m:
Hello,

I am owner of ```zajęzyk.pl``` (punycode: ```xn--zajzyk-y4a.pl```) hosted on CloudFlare.

I got following errors:
```
$ octodns-dump --config-file=env/production.yaml --output-dir=zones zajęzyk.pl. cloudflare_shared
2018-02-18T12:58:33  [140367053055744] INFO  Manager __init__: config_file=env/production.yaml
2018-02-18T12:58:33  [140367053055744] INFO  Manager __init__:   max_workers=1
2018-02-18T12:58:33  [140367053055744] INFO  Manager __init__:   max_workers=False
Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 3: ordinal not in range(128)
Logged from file manager.py, line 347
Traceback (most recent call last):
  File "/home/adas/.virtualenvs/infra/bin/octodns-dump", line 11, in <module>
    sys.exit(main())
  File "/home/adas/.virtualenvs/infra/local/lib/python2.7/site-packages/octodns/cmds/dump.py", line 31, in main
    manager.dump(args.zone, args.output_dir, args.lenient, *args.source)
  File "/home/adas/.virtualenvs/infra/local/lib/python2.7/site-packages/octodns/manager.py", line 360, in dump
    zone = Zone(zone, self.configured_sub_zones(zone))
  File "/home/adas/.virtualenvs/infra/local/lib/python2.7/site-packages/octodns/manager.py", line 159, in configured_sub_zones
    pieces = zone_name[:-1].split('.')[::-1]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 3: ordinal not in range(128)
$ octodns-dump --config-file=env/production.yaml --output-dir=zones xn--zajzyk-y4a.pl. cloudflare_shared
2018-02-18T12:59:28  [139790821807872] INFO  Manager __init__: config_file=env/production.yaml
[...]

PR #903, opened 2022-05-20
Add idna encode/decode helpers

ross:
/cc https://github.com/octodns/octodns/issues/199
/cc https://datatracker.ietf.org/doc/html/rfc3490


octodns/idna.py
```diff
@@ -0,0 +1,32 @@
...
+        if name.startswith('*'):
+            # idna.encode doesn't like the *
+            name = _encode(name[2:]).decode('utf-8')
+            return f'*.{name}'
+        return _encode(name).decode('utf-8')
+
+
+def idna_decode(name):
+    pieces = name.lower().split('.')
+    if any([p.startswith('xn--') for p in pieces]):
```
viranch, line 25: Slightly better?

```suggestion
    if any(p.startswith('xn--') for p in pieces):
```

ross: :cool: guess I didn't know the list comprehension would work w/o the `[]`

viranch:
> Each provider will need updates to pull in and use these methods appropriately at the last minute.

Some documentation/hints around how and when to use the helpers would be very helpful for the providers.

octodns/idna.py
viranch: I think in my version, `any` will get an iterator which will stop evaluating subsequent elements after it finds a `True`.

ross:
> Some documentation/hints around how and when to use the helpers would be very helpful for the providers.

To some degree that's still TBD, but I can probably put something together based on what NS1 looks like https://github.com/octodns/octodns-ns1/pull/20.

viranch:
> I can probably put something together based on what NS1 looks like

Even a PR would be a good starting point for reference. Doesn't have to be in this PR.

viranch (approved):
Simple enough from code perspective, I don't have use-cases irl to test it. Ship it!

ross:
After getting distracted and working on non-octoDNS stuff for a bit I'm getting back to this.

Issue #378, opened 2019-07-17
Cannot manage subzones that are more than one level under parent


nakato:
When attempting to add NS records for a subzone that is more than one level under the parent zone OctoDNS refuses to add the record.

I would expect to be able to configure the zone in the parent zone add a zone between the parent and child.

domain.tld.yaml
```
---
zone.stg:
  ttl: 600
  type: NS
  values:
    - ns1.domain.tld
    - ns2.domain.tld
```

zone.stg.domain.tld.yaml
```
---
record:
  ttl: 60
  type: A
  values:
    - 1.2.3.4
```

`stg.domain.tld.yaml` does not exist, nor is `stg.domain.tld.` configured in OctoDNS config, nor does `stg.domain.tld.` exist in any DNS server.

Results in:
```
Traceback (most recent call last):
  File "/usr/local/bin/octodns-sync", line 11, in <module>
    load_entry_point('octodns', 'console_scripts', 'octodns-sync')()
  File "/octodns/octodns/cmds/sync.py", line 39, in main
    dry_run=not args.doit, force=args.force)
  File "/octodns/octodns/manager.py", line 302, in sync
    plans = [p for f in futures for p in f.result()]
  File "/octodns/octodns/manager.py", line 56, in result
    return self.func(*self.args, **self.kwargs)
  File "/octodns/octodns/manager.py", line 224, in _populate_and_plan
    source.populate(zone)
  File "/octodns/octodns/provider/yaml.py", line 77, in populate
    self._populate_from_file(filename, zone, lenient)
  File "/octodns/octodns/provider/yaml.py", line 62, in _populate_from_file
    zone.add_record(record, lenient=lenient)
  File "/octodns/octodns/zone.py", line 68, in add_record
    .format(record.fqdn))
octodns.zone.SubzoneRecordException: Record zone.stg.domain.tld. is under a managed subzone
```

ross:
The `NS` record in the top-level zone should be defined in the sub-zone since that zone owns that part of things. It'd have a record name of `''` there.

nakato:
OctoDNS doesn't really support adding NS records to subzones, except in the case of PowerDNS, where it is defined in the provider configuration.  Issue #38 seems to contain the details on this, and a NS record on `''` would be silently dropped anyways in [zone.py](https://github.com/github/octodns/blob/master/octodns/zone.py#L27)

[...]

PR #908, opened 2022-06-23
Support dotted subdomains for subzones

nakato:
Currently if there are two zones configured;
- example.com.
- delegated.subdomain.example.com.

When an NS record is created in example.com.yaml as such:
delegated.subdomain:
  type: NS
  values:
  - ns1.example.org.

The NS record for delegated.subdomain.example.com cannot be created as it
throws an exception:

octodns.zone.SubzoneRecordException: Record delegated.subdomain.example.com is under a managed subzone

Additionally, all records other than NS are rejected for
subdomain.example.com..

This is caused by zone_tree being the result of all zones split on '.'
and being added to the tree, even if a zone does not exist at that
point.

To support records where a subzone is dotted, the the map is built such
that each node represents the subdomain of the closest subzone.

Before:
{"com", {"example": {"subdomain": {"delegated": {}}}}}

After:
{"com.example": {"subdomain.delegated": {}}}

Fixes: https://github.com/octodns/octodns/issues/378


nakato:
Force push between 9b926e6 and 5592f5d:
Changed tree from having the subzone name be the subdomain in reverse domain name notation to having the subdomain in the order a user would expect.
The difference between these two means that using the details contained Zone.sub_zones is natural and not error prone, not needing to be re-reversed to be useful to something such as a Processor, which might insert NS records for subzones automatically.

ross:
Will be a week or so before I'll get a chance to dig into this. Thanks for the PR. Will update once I get to 👀

ross:
So the currently working/supported path to do this is to split out `subdomain.example.com.` into its own zone as well and then place the NS record for `delegated.subdomain.example.com.` in that zone. Skipping levels when breaking out zone files isn't currently supported. 

Trying to decide whether it should be or not and looking at the PR itself.

octodns/manager.py
```diff
@@ -162,23 +162,18 @@ def __init__(self, config_file, max_workers=None, include_meta=False):
...
+        # zone exists it will be seen after the subzone, thus we can easily
+        # reparent children to their parent zone from the tree root.
+        for name in sorted(self.config['zones'].keys(),
+                           key=lambda s: 0 - s.count('.')):
+            name = name[:-1]
+            this = {}
+            for sz in filter(
+                lambda k: k.endswith(name), set(zone_tree.keys())
+            ):
+                this[sz[:-(len(name) + 1)]] = zone_tree.pop(sz)
```
ross, line 175: this section

octodns/manager.py
```diff
@@ -274,21 +269,14 @@ def _build_kwargs(self, source):
...
-                # Point where at the value of our current piece
-                where = where[pieces.pop(0)]
-        except KeyError:
-            self.log.debug('configured_sub_zones: unknown zone, %s, no subs',
-                           zone_name)
-            return set()
-        # We're not pointed at the dict for our name, the keys of which will be
-        # any subzones
+        while True:
+            parent = next(filter(lambda k: name.endswith(k), where), None)
```
ross, line 275: these lines

tests/config/unit.tests.yaml
```diff
@@ -162,6 +162,11 @@ sub:
   values:
   - 6.2.3.4.
   - 7.2.3.4.
```
ross, line 164: We may want to add a `NS` for `subzone.unit.tests.` here to make sure that a child works in addition to grandchild. Probably should have been here in the first place, but since we're in the neighborhood it's a good time to add it.

ross (commented):
Overall looks good. Couple non-trivial places where I'd prefer more comments, akin to what was there before. And a comment about a hopefully simple to add additional test.

octodns/manager.py
```diff
@@ -274,21 +269,14 @@ def _build_kwargs(self, source):
...
+            if not parent:
+                break
+            where = where[parent]
```
ross, line 278: Fwiw, using list comprehensions over next & filter is preferred in py3, though both are a little cumbersome here.

If my read/understanding is correct something like this may work:

```suggestion
            parents = [k for k in where if name.endswith(k)]
            if not parents:
                break
            where = where[parent[0]]
```

octodns/zone.py
```diff
@@ -68,10 +68,10 @@ def add_record(self, record, replace=False, lenient=False):
             self.hydrate()
 
         name = record.name
-        last = name.split('.')[-1]
 
-        if not lenient and last in self.sub_zones:
-            if name != last:
+        if not lenient and any(map(lambda sz: name.endswith(sz),
+                                   self.sub_zones)):
```
ross, line 73: Again if my understanding is correct something like may work:

```suggestion
        if not lenient and any([name.endswith(sz) for sz in self.sub_zones]):
```

octodns/manager.py
nakato: There will only ever be one parent, thus my goal was to have the search be lazy and to stop attempting string comparisons once the parent was found.

If you're okay with keeping the `next()`?
If so, I could use a list comprehension as follows, that will keep that short-circuit, which I think is cleaner as it doesn't require indexed access to the value on line 278 and 279.
```
parent = next((k for k in where if name.endswith(k)), None)
```

nakato: I've used this in the follow commit adding comments and switching to list comprehensions.

nakato:
> So the currently working/supported path to do this is to split out subdomain.example.com. into its own zone as well and then place the NS record for delegated.subdomain.example.com. in that zone. Skipping levels when breaking out zone files isn't currently supported.
>
> Trying to decide whether it should be or not and looking at the PR itself.

I'll strongly advocate for supporting dotted subdomain zones.  It is valid per the DNS RFC's, and each additional zone is a maintenance burden on operators, especially as deletions are a manual effort.

From a user viewpoint I think it would be unnecessarily rigid to require a subzone for every dot between the parent zone and the subzone; this is mentioned in [RFC1034, Section 4.2 Paragraph 5](https://datatracker.ietf.org/doc/html/rfc1034#section-4.2) as follows.

> It would be possible, though not particularly useful, to partition the
> name space so that each domain name was in a separate zone or so that
> all nodes were in a single zone.

With a subzone for every dot in a subdomain's name, this would preclude the user's ability to use CNAME records for those names as well as signal to users they should create new zones for everything.  I advocate to users to limit the creation of subzones to only when it is required to denote some form of ownership.

As such I want users to do the following:

config.yaml
```
zones:
  example.com:
    ...
```

example.com.yaml
```
---
blue.prod:
  ...
green.prod:
  ...
```

`prod` would be valid to be a cname in this case.

---

I don't want users to do this:

config.yaml
```
zones:
  example.com:
    ...
  prod.example.com:
    ...
```

example.com.yaml
```
---
   prod:
    type: NS
    ...
```

prod.example.com.yaml
```
---
blue:
  ...
green:
  ...
```

`prod.example.com` could not be a cname in this layout.

---

However, if the user had a valid reason to have `cluster.prod.example.com.` as a subzone, I'd need them to move to this second layout.  It would become even more confusing for the user if they wanted to delegate `cluster.blue.prod.example.com.`.

octodns/manager.py
ross: That seems good. More inline with what py3 prefers.

ross (approved):
Looks good. Thanks for getting the comments added and switching to comprehensions.

PR #913, opened 2022-07-22
AxfrSource and ZoneFileSource: support multi-value PTR records

### LINKED ISSUE #912: ZoneFileSource provider to support multi-valued PTR records?
I'm going through the process of importing our zone file into yaml.  some of our PTR records have multiple values, however these are not handled by the `ZoneFileSource` provider. any chance that this could be supported?

eg:

in `10.in-addr.arpa.dns`
```txt
1.2.3  PTR name1.domain1.abc.
          PTR name2.domain2.xyz.

```

currently run this through the `ZoneFileSource` provider will only yield the following

```yaml
1.2.3:
  ttl: 18000
  type: PTR
  value: name1.domain1.abc.
```

any chance of having the support to capture both PTR values from the original zone file? esp now that octoDNS does support multi-valued PTR records? 

something like this

```yaml
1.2.3:
  ttl: 18000
  type: PTR
  values:
    - name1.domain1.abc.
    - name2.domain2.xyz.

```

or there's another method that i m not aware of?

yzguy:
Fixes #912 

PTR records can be multi-valued, the initial source was written with an mistaken assumption that they were only single value.

/cc @harpreetnav

### DIFF
--- a/octodns/source/axfr.py
+++ b/octodns/source/axfr.py
@@ -59,6 +59,7 @@ class AxfrBaseSource(BaseSource):
     _data_for_A = _data_for_multiple
     _data_for_AAAA = _data_for_multiple
     _data_for_NS = _data_for_multiple
+    _data_for_PTR = _data_for_multiple
 
     def _data_for_CAA(self, _type, records):
         values = []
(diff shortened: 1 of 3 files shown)


ross (approved):
:shipit:

PR #914, opened 2022-08-09
Rel 0.9.18 version bump & CHANGELOG update

ross:
## v0.9.18 - 2022-09-09 - Internationalization

* Added octodns.idna idna_encode/idna_decode helpers, providers will need to
  individually add support via these helpers though :-/


ross:
Release was yanked (only available with exact pin) due to https://github.com/octodns/octodns/issues/916. Fixed in 0.9.19, #920

PR #915, opened 2022-08-09
Fqdn check w idna encode

ross:
Allow UTF-8 values in record answers. We were previously using a `FQDN` function to check values with a `str` to fix issues with passing unicode strings in 2.7. 2.7 is no more (for octoDNS) so we don't need the `str`, but the `FQDN` checks still don't allow utf-8 so we need to encode before sending things off to it.

These are only validations and if you run into them before this is included in a release you can either pin a SHA will this change or temporarily add `lenient` to the records to ignore the validation issues.

/cc https://github.com/octodns/octodns/issues/199 
/cc https://github.com/octodns/octodns-ns1/pull/20#issuecomment-1209560348 @viranch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v0.9.19 - 2022-??-?? -
+
+* Allow utf-8 (IDNA) values in record types that support fqdn for
+  values/targets
+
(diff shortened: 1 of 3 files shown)


ross:
Subsumed by https://github.com/octodns/octodns/pull/935

Issue #916, opened 2022-08-10
octodns.zone.SubzoneRecordException Error


brianeclow:
The `0.9.18` version causes an `octodns.zone.SubzoneRecordException` error. I reverted back to `0.9.17` confirm this was something new. Were there some required changes we need to implement on our local files?

On one of our bigger zones, source octodns.provider.yaml (via split), targets NS1 and Route53:

```bash
Traceback (most recent call last):
  File "/dns/env/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/dns/env/lib/python3.9/site-packages/octodns/cmds/sync.py", line 64, in main
    manager.sync(
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 551, in sync
    ps, d = future.result()
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 77, in result
    return self.func(*self.args, **self.kwargs)
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 359, in _populate_and_plan
    source.populate(zone, lenient=lenient)
  File "/dns/env/lib/python3.9/site-packages/octodns/provider/yaml.py", line 329, in populate
    self._populate_from_file(yaml_filename, zone, lenient)
  File "/dns/env/lib/python3.9/site-packages/octodns/provider/yaml.py", line 187, in _populate_from_file
    zone.add_record(
  File "/dns/env/lib/python3.9/site-packages/octodns/zone.py", line 79, in add_record
    raise SubzoneRecordException(
octodns.zone.SubzoneRecordException: Record jigglypuff.example.com. is under a managed subzone
```

On one of our parked/minimal zones, source octodns.provider.yaml, target NS1:

```bash
Traceback (most recent call last):
  File "/dns/env/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
  File "/dns/env/lib/python3.9/site-packages/octodns/cmds/sync.py", line 64, in main
    manager.sync(
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 551, in sync
    ps, d = future.result()
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 77, in result
    return self.func(*self.args, **self.kwargs)
  File "/dns/env/lib/python3.9/site-packages/octodns/manager.py", line 359, in _populate_and_plan
    source.populate(zone, lenient=lenient)
[...]

PR #917, opened 2022-08-11
Fix issues with sub-zone handling

ross:
Refactor `Manager.zone_tree` for easier testing and add some thorough tests of it as well as `Manager.configured_sub_zones`.

/cc https://github.com/octodns/octodns/issues/916 which reported issues with sub-zone handling as of 0.9.18
/cc https://github.com/octodns/octodns/pull/908 which was a change in that release and likely related. 
/cc @brianeclow @nakato

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -315,12 +315,13 @@ class Manager(object):
             while zones:
                 # Grab the one we'lre going to work on now
                 zone = zones.pop()
-                trimmer = len(zone) + 1
+                dotted = f'.{zone}'
+                trimmer = len(dotted)
                 subs = set()
                 # look at all the zone names that come after it
                 for candidate in zones:
-                    # If they end with this zone's name them they're a sub
-                    if candidate.endswith(zone):
+                    # If they end with this zone's dotted name, it's a sub
+                    if candidate.endswith(dotted):
                         # We want subs to exclude the zone portion
                         subs.add(candidate[:-trimmer])
 
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -185,25 +185,6 @@ class Manager(object):
                     'Incorrect processor config for ' + processor_name
                 )
 
-        zone_tree = {}
-        # Sort so we iterate on the deepest nodes first, ensuring if a parent
-        # zone exists it will be seen after the subzone, thus we can easily
-        # reparent children to their parent zone from the tree root.
-        for name in sorted(
-            self.config['zones'].keys(), key=lambda s: 0 - s.count('.')
-        ):
-            # Trim the trailing dot from FQDN
-            name = name[:-1]
-            this = {}
-            for sz in [k for k in zone_tree.keys() if k.endswith(name)]:
-                # Found a zone in tree root that is our child, slice the
-                # name and move its tree under ours.
-                this[sz[: -(len(name) + 1)]] = zone_tree.pop(sz)
-            # Add to tree root where it will be reparented as we iterate up
-            # the tree.
-            zone_tree[name] = this
-        self.zone_tree = zone_tree
-
         self.plan_outputs = {}
         plan_outputs = manager_config.get(
             'plan_outputs',
@@ -243,6 +224,32 @@ class Manager(object):
                 raise ManagerException(
                     'Incorrect plan_output config for ' + plan_output_name
                 )
+
+        self._zone_tree = None
+
+    @property
+    def zone_tree(self):
+        if self._zone_tree is None:
+            zone_tree = {}
+            # Sort so we iterate on the deepest nodes first, ensuring if a parent
+            # zone exists it will be seen after the subzone, thus we can easily
+            # reparent children to their parent zone from the tree root.
+            for name in sorted(
+                self.config['zones'].keys(), key=lambda s: 0 - s.count('.')
+            ):
+                # Trim the trailing dot from FQDN
+                name = name[:-1]
+                this = {}
+                for sz in [k for k in zone_tree.keys() if k.endswith(name)]:
+                    # Found a zone in tree root that is our child, slice the
+                    # name and move its tree under ours.
+                    this[sz[: -(len(name) + 1)]] = zone_tree.pop(sz)
+                # Add to tree root where it will be reparented as we iterate up
+                # the tree.
+                zone_tree[name] = this
+            self._zone_tree = zone_tree
+
+        return self._zone_tree
 
     def _try_version(self, module_name, module=None, version=None):
         try:
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -228,37 +228,7 @@ class Manager(object):
                     'Incorrect plan_output config for ' + plan_output_name
                 )
 
-        self._zone_tree = None
-
-    @property
-    def zone_tree(self):
-        if self._zone_tree is None:
-            zone_tree = {}
-
-            # Get a list of all of our zone names. Sort them from shortest to
-            # longest so that parents will always come before their subzones
-            zones = sorted(
-                self.config['zones'].keys(), key=lambda z: len(z), reverse=True
-            )
-            zones = deque(zones)
-            # Until we're done processing zones
-            while zones:
-                # Grab the one we'lre going to work on now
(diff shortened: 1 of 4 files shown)


ross:
My current thinking is that `another.sub.unit.tests.` should not have any configured sub-zones:

```
=================================== FAILURES ===================================
______________________ TestManager.test_subzone_handling _______________________

self = <test_octodns_manager.TestManager testMethod=test_subzone_handling>

    def test_subzone_handling(self):
        manager = Manager(get_config_filename('simple.yaml'))
    
        manager.config['zones'] = {
            'unit.tests.': {},
            'sub.unit.tests.': {},
            'another.sub.unit.tests.': {},
            'skipped.alevel.unit.tests.': {},
        }
    
        self.assertEqual(
            {'unit.tests': {'skipped.alevel': {}, 'sub': {'another': {}}}},
            manager.zone_tree,
        )
        self.assertEqual(
            {'sub', 'skipped.alevel'},
            manager.configured_sub_zones('unit.tests.'),
        )
        self.assertEqual(
            {'another'}, manager.configured_sub_zones('sub.unit.tests.')
        )
>       self.assertEqual(
            set(), manager.configured_sub_zones('another.unit.tests.')
        )
E       AssertionError: Items in the second set but not the first:
E       'sub'
E       'skipped.alevel'
```

ross:
Printing out name and parent at the top of the `configured_sub_zone` while loop and they aren't really what I expected, though still wrapping my head around things:

```python
{'name': 'unit.tests', 'parent': 'unit.tests'}
{'name': '', 'parent': None}
{'name': 'sub.unit.tests', 'parent': 'unit.tests'}
{'name': 'sub', 'parent': 'sub'}
{'name': '', 'parent': None}
{'name': 'another.unit.tests', 'parent': 'unit.tests'}
{'name': 'another', 'parent': None}
```

ross:
OK. I reworked and greatly simplified the overall code that builds the sub-zone list. It all happens up front and the actual request for sub-zones just retrieves the results when needed. 

@brianeclow would you be able to pull down this branch and test your config with it to see if it works as expected?

nakato:
I really like moving large chunks of logic like this into their own functions.

> Printing out name and parent at the top of the `configured_sub_zone` while loop and they aren't really what I expected, though still wrapping my head around things:

> Specifically that `sub` is its own parent.

That's the key name for us to grab 'sub' out of the dictionary at the location we're currently at, I think printing sub_zone_names after the loop would be more clear.


Taking the output of `zone_tree` with formatting we're looking at 
```
{
  'unit.tests': {
    'skipped.alevel': {},
    'sub': {
      'another': {}
    }
  }
}
```

When `parent` points to `sub`, it takes that from the dictionary, then looks for more parents.


I think what happened with #916 is the `str.endswith(...)` in [zone.py:76](https://github.com/octodns/octodns/blob/62caa4bac6f0dc0862584080357498404f030ec6/octodns/zone.py#L76) is being greedy with longer labels.
This can be seen in #918.

That nets me the "under a managed subzone" not the "is a managed subzone" error, I haven't been able to work out how the "is a managed subzone" would appear now that wouldn't have previously, as that requires an exact match.

Looking at `configured_sub_zones` now, as I had not started looking into it yet, I think the `name.endswith(k)` is probably a problem in this function as well, the same as it is in #918.

octodns/manager.py
```diff
@@ -244,6 +227,37 @@ def __init__(self, config_file, max_workers=None, include_meta=False):
...
+                    # If they end with this zone's name them they're a sub
+                    if candidate.endswith(zone):
+                        # We want subs to exclude the zone portion
+                        subs.add(candidate[:-trimmer])
+
+                zone_tree[zone] = subs
+
+            self._zone_tree = zone_tree
+
+        return self._zone_tree
```
nakato, line 259: The output of this is a bit harder to parse in my opinion as it is no-longer tree-like.
Certainly easier to grab children of a zone however as it's a simple map however.

Previously
```
{
  'unit.tests': {
    'skipped.alevel': {},
    'sub': {
        'another': {}
    }
  }
}
```

Replacment:
```
{
  'unit.tests.': {
    'skipped.alevel',
    'sub', 
    'another.sub'  # < Should this one be listed here, as it's a subzone of `sub`
  }, 
  'sub.unit.tests.': {
    'another'
  },
  'another.sub.unit.tests.': set(), 
  'skipped.alevel.unit.tests.': set()
}
```

octodns/manager.py
```diff
@@ -244,6 +227,37 @@ def __init__(self, config_file, max_workers=None, include_meta=False):
...
+
+            # Get a list of all of our zone names
+            zones = list(self.config['zones'].keys())
+            # Sort them from shortest to longest so that parents will always
+            # come before their subzones
+            zones.sort(key=lambda z: len(z))
+            # Until we're done processing zones
+            while zones:
+                # Grab the one we'lre going to work on now
+                zone = zones.pop(0)
```
nakato, line 245: Removing an element from the front of the list is `O(n)` expensive isn't it?
Maybe collections.deque would be a good fit here?

octodns/manager.py
```diff
@@ -244,6 +227,37 @@ def __init__(self, config_file, max_workers=None, include_meta=False):
...
+                trimmer = len(zone) + 1
+                subs = set()
+                # look at all the zone names that come after it
+                for candidate in zones:
+                    # If they end with this zone's name them they're a sub
+                    if candidate.endswith(zone):
```
nakato, line 251: I think this is part of the original bug, if I pass `ssub.unit.tests.` and `sssub.unit.tests.` through as zones I get the following:
```
{
    ...,
    'sub.unit.tests.': {'', 'another', 's'},
    'ssub.unit.tests.': {''},
    'sssub.unit.tests.': set(),
    ...,
```
That's definitely not right, as the tree would see `unit.tests.` have three children, `sub, ssub, sssub`.

octodns/manager.py
ross: Yeah. Looking back at it now I didn't like that the computation/logic happened in two different places/times. That made changes in one place break the other etc which is 😢.

Good point that `zone_tree` no longer makes sense as the name. I'll clean that up. Looks like zone_tree is only used during `configured_sub_zones` so the right solution is probably to ditch the property and do the computing in `configured_sub_zones` itself.  It think that'll even let things do it per zone and only when needed.

octodns/manager.py
ross: On paper it's a problem, but given that we're really only working with sizes on the order of 0-10 it's negligible. Since deque is part of the std lib it seems reasonable to switch to it though.

octodns/manager.py
ross: Yeah. That's pretty much the same issue you found in https://github.com/octodns/octodns/pull/918

ross: Added a test to hit this problem. Looking at solutions now.

ross: This one was trivial, just needed to search for endswith the zone name with a dot prepended. It now work similar to `add_record`'s check so if one is understood the other will make more sense.

brianeclow:
Pulled the octoDNS branch

- Ran non-doit against all zones. No errors; expected results.
- Ran doit on `example.com` zone. No errors; expected results.
- Ran doit on `example2.com` zone. No errors; expected results.

ross:
Going to make another pass over things to think about them and see if I can come up with any other edge/test cases before 🚢ing.

nakato (commented):
This looks good to me.

tests/test_octodns_manager.py
```diff
@@ -712,6 +712,125 @@ class DummyModule(object):
             ),
         )
 
+    def test_subzone_handling(self):
```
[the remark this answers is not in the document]
ross: Yeah, I'm not a fan of how much the tests were done as those big config-based stuff early on (my own doing.) I make it a point to refactor bits out to more easily test them whenever the opportunity arrises.

ross:
@yzguy @viranch would one of you be able to 👀  and 👎 / 👍?

viranch (approved):
Haven't been in this territory but code looks ok 👍

PR #918, opened 2022-08-11
Fix subzone false positives on label longer than subzone

nakato:
If the label of a zone is longer than the subzone, then the matching on
label.endswith(subzone) will improperly match due to checking if the
label endswith any of the subzones.

As such, this means given the two following domains:
1. example.com
2. puff.example.com

If the label "jigglypuff" exists in the "example.com" zone, the
following error will be thrown.

octodns.zone.SubzoneRecordException: Record jigglypuff.example.com. is under a managed subzone

---

I'm not 100% certain this fully addresses #916, as I'm not sure how an exact subzone match is occurring, error message `Record example2.com. a managed sub-zone and not of type NS`, that would not have triggered in 0.9.17 but is in 0.9.18.  The names are obfuscated, so I've had to guess a lot here.

### DIFF
--- a/octodns/zone.py
+++ b/octodns/zone.py
@@ -73,18 +73,22 @@ class Zone(object):
 
         name = record.name
 
-        if not lenient and any((name.endswith(sz) for sz in self.sub_zones)):
+        # Create an exhaustive list of potential subzones for a label
+        parts = name.split('.')
+        name_zones = []
+        for index in range(len(parts)):
+            name_zones.append('.'.join(parts[index:]))
+
+        if not lenient and any(nz in self.sub_zones for nz in name_zones):
             if name not in self.sub_zones:
                 # it's a record for something under a sub-zone
                 raise SubzoneRecordException(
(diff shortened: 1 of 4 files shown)


ross:
If you don't mind take a look at https://github.com/octodns/octodns/pull/917, I think it simplifies things a bit so that it's easier to grok how this process works. I also pulled the method out of `__init__` so that it could be specifically tested more easily. I'll look at adding the test case in this PR to that one to see if it passes.

ross:
Oh, when I went to look at the test case more closely I realized the change here is actually to Zone. 👀

ross:
Ah. The problem this is actually fixing is when records have a name that ends with a subzone, but without the `.`. That's a good catch, though I think I'd like to explore fixing it another way, by leaving the `.` intact on the subzone names. Will explore that for a bit now.

nakato:
Closing #917 fixes what I found here and more, as well as making subzone handling substantially more testable.

PR #920, opened 2022-08-14
Rel 0.9.19 version bump & CHANGELOG update

ross:
## v0.9.18 - 2022-09-14 - Subzone handling

* Fixed issue with sub-zone handling introduced in 0.9.18

## v0.9.18 - 2022-09-09 - Internationalization

* Added octodns.idna idna_encode/idna_decode helpers, providers will need to
  individually add support via these helpers though :-/
* `black` formatting implemented (including .git-blame-ignore-revs)
* --output-provider support for dump to allow configurable dump
  formatting/details
* TLSA record type support
* Subzones support for skipping levels

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.9.18 - 2022-09-14 - Subzone handling
+
+* Fixed issue with sub-zone handling introduced in 0.9.18
+
 ## v0.9.18 - 2022-09-09 - Internationalization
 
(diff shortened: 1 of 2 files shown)


xt0rted:
@ross the dates for the last two versions are the wrong month, they should be `08` instead of `09`

ross:
> @ross the dates for the last two versions are the wrong month, they should be `08` instead of `09`

Doh. Thanks. I'll get it updated in the repo tomorrow. 🚢 has sailed on the releases so not much that can be done with them (definitely don't want to move the tags.)

PR #929, opened 2022-09-06
Record values are first-class objects

ross:
There are backwards compatible as reasonably possible. The real power of this will come as times goes in in our ability to simplify providers. E.g. implement the to/from RFC-style strings once in the value types and remove code that is duplicated many times across providers.

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -9,7 +9,7 @@ from __future__ import (
     unicode_literals,
 )
 
-from ipaddress import IPv4Address, IPv6Address
+from ipaddress import IPv4Address as _IPv4Address, IPv6Address as _IPv6Address
 from logging import getLogger
 import re
 
@@ -326,7 +326,7 @@ class ValuesMixin(object):
     def __init__(self, zone, name, data, source=None):
         super(ValuesMixin, self).__init__(zone, name, data, source=source)
         try:
-            values = data['values']
+            values = [v for v in data['values']]
         except KeyError:
             values = [data['value']]
         self.values = sorted(self._value_type.process(values))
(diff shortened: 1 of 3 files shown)


ross:
Woah, all of the module tests actually pass against this branch. I'm a little surprised. https://github.com/octodns/octodns/actions/runs/3003810303

ross:
Ran some octodns-sync tests against this branch with a couple providers and the results match the modules tests and everything is happy. I've also looked over how the values are being used and it seems compatible with the changes.

In reality the more complex types were already using objects, You can see this by scanning through the `def process` methods before this change https://github.com/octodns/octodns/blob/c11476d79ad7e8889f5bdbe7052ed01640bc97c0/octodns/record/__init__.py, e.g. https://github.com/octodns/octodns/blob/c11476d79ad7e8889f5bdbe7052ed01640bc97c0/octodns/record/__init__.py#L1170-L1172

That's in contract to the simple types like A and AAAA which were being stored as strings and thus can't do anything interesting. e.g. https://github.com/octodns/octodns/blob/c11476d79ad7e8889f5bdbe7052ed01640bc97c0/octodns/record/__init__.py#L1316-L1318

There are others where more interesting things were happening before they became a simple type, e.g.

https://github.com/octodns/octodns/blob/c11476d79ad7e8889f5bdbe7052ed01640bc97c0/octodns/record/__init__.py#L794-L801

This makes them all work the same and have objects there so that we should be able to do things like the following, assuming we add `to_rfc_text`. Similar additions should be possible to handle from rfc text formats. 

```python
    [v.to_rfc_text() for v in record.values]
```

ross:
Looks like there's a module relying on `*Value.data` properties that were removed in 5d2ef02, https://github.com/octodns/octodns-mythicbeasts/blob/805b4af4d5888d480099bbb5be7daaefcfb46412/octodns_mythicbeasts/__init__.py#L399. Putting them back with more deprecation messages/intentions.

ross:
Mostly an under the hood change so that values are always first-class objects (even when they're simple strings) so that we can add more functionality in subsequent PRs (like to/from RFC text.

octodns/record/__init__.py
```diff
@@ -905,13 +909,41 @@ def process(cls, values):
...
+    def tag(self, value):
+        self['tag'] = value
+
+    @property
+    def value(self):
+        return self['value']
+
+    @value.setter
+    def value(self, value):
+        self['value'] = value
```
ross, line 942: Changing these to be dicts allows them to directly be serialized by yaml as-is. It's not a critical change, but it seems more consistent this way...

Majority of these changes are just boilerplate-y bits.

octodns/record/__init__.py
```diff
@@ -1717,34 +1981,67 @@ def process(cls, values):
         return [UrlfwdValue(v) for v in values]
```
ross, line 1981:
```suggestion
        return [cls(v) for v in values]
```

Issue #927, opened 2022-09-04
RFC 2136 support


siilike:
It would be great to also have RFC 2136 support, so octoDNS could also be used with self-hosted DNS servers (Knot, BIND, etc).

ross:
RFC: https://datatracker.ietf.org/doc/html/rfc2136

Looks like [dnspython](https://pypi.org/project/dnspython/) has support for the underlying implementation details. I found an example of it being used [here](https://pypi.org/project/dnspython/) and that looks fairly straightforward.

There's already https://github.com/octodns/octodns-ddns which would be the logical place for this code to live since based on the name of the concept. It currently just implements a `Source` that does a what's my ip lookup. Having a provider that can push things via RFC 2136 seems doable. 

Main question I have atm though is what would be required to see what the provider thinks is the current value for the IP. I'm guessing a dns query to the same server that will be receiving the update is the logical path, but there maybe hiccups with caching/ttls.

yzguy:
@ross I was just thinking about this too and I think honestly if you're doing RFC2136 for updates it kind of seems logical that you'd do a zone transfer for the read bits.

We have the AxfrSource already though so it would be repeating code basically to use it for the ddns bits

ross:
> @ross I was just thinking about this too and I think honestly if you're doing RFC2136 for updates it kind of seems logical that you'd do a zone transfer for the read bits.

Interesting. Guess that makes sense. Would it make sense to require that be the case or does it need to fallback to a query or something?

> We have the AxfrSource already though so it would be repeating code basically to use it for the ddns bits

Hrm. Perhaps it could inherit from the source and ~turn it into a Provider. Don't think anything has done that before, but it it seems like it might be doable.

ross:
pfsense looks to have some decent docs on enabling rfc-2136 support in bind:
- https://docs.netgate.com/pfsense/en/latest/recipes/bind-rfc2136.html
- https://docs.netgate.com/pfsense/en/latest/services/dyndns/rfc2136.html

ross:
It's a hacky mess of testing/poking/prodding, but: 

from bind9 logs

```
06-Sep-2022 03:30:05.212 client @0x7f46c3c01b68 172.22.0.1#60076/key octodns.exxampled.com: signer "octodns.exxampled.com" approved
[...]

PR #930, opened 2022-09-07
(mostly) RFC compliant to/from RR text support for Records

ross:
This PR implements RFC based parsing and encoding of applicable record types. Specifically it adds `parse_rdata_text` and `rdata_text` to all of the value object types and centralizes some of the logic of `AxfrSource` into `Record` (with the help of `ValuesMixin` and `ValueMixin`. 

This implementation should be ~complete in terms of sourcing records from RR/RData. There will likely be further bits added in a subsequent PR once I dig into going the other direction and creating [ZonefileProvider](https://github.com/octodns/octodns/issues/254)

This should have as much testing as AxfrSource did before, hopefully much more since it adds individual/specific testing of the value types.

/cc https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4 for RFC data
/cc https://github.com/octodns/octodns/pull/929 which made all the value types objects allowing the new methods added here
/cc https://github.com/octodns/octodns/issues/927#issuecomment-1238834192 which started me off down the path that lead to this


ross:
WIP bind provider that will be the likely new home of the AXFR stuff now that it's becoming a full-fledged provider. For now it's just a port of the axfr stuff in core with a little bit of refactoring to get rid of the need of inheriting from both BaseSource and BaseProvider when we want to do both AXFR and RFC 2136. So far liking how it's looking. It'll also have a docker-compose setup to run bind for local dev, similar to octodns-powerdns when I get there.

ross:
This is now a much safer/simpler change as it no longer tries to call `parse_rdata_text` from `validate` or the value constructors. Turned out that wasn't viable since there's know what to know if a string value type was rdata encoded or not. This mainly mattered for TXT/SPF etc where the escaping varied. 

That in combination with a gut feeling that it was better to make the parse explicit lead to the change. I'm pretty happy with this now.

ross:
[8b82159](https://github.com/octodns/octodns/pull/930/commits/8b82159ee0b09fcdffde47812a53c087f7adfe82) was accidentally pushed up to this branch, shouldn't be here so reverted it back out.

PR #931, opened 2022-09-10
Update README.md

LuPo:
Adding links to the noteworthy NetBox-DNS plugin and its OctoDNS provider

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,7 @@ If you have a problem or suggestion, please [open an issue](https://github.com/o
   - [`doddo/octodns-lexicon`](https://github.com/doddo/octodns-lexicon): Use [Lexicon](https://github.com/AnalogJ/lexicon) providers as octoDNS providers.
   - [`asyncon/octoblox`](https://github.com/asyncon/octoblox): [Infoblox](https://www.infoblox.com/) provider.
   - [`sukiyaki/octodns-netbox`](https://github.com/sukiyaki/octodns-netbox): [NetBox](https://github.com/netbox-community/netbox) source.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -363,6 +363,7 @@ If you have a problem or suggestion, please [open an issue](https://github.com/o
   - [`doddo/octodns-lexicon`](https://github.com/doddo/octodns-lexicon): Use [Lexicon](https://github.com/AnalogJ/lexicon) providers as octoDNS providers.
   - [`asyncon/octoblox`](https://github.com/asyncon/octoblox): [Infoblox](https://www.infoblox.com/) provider.
   - [`sukiyaki/octodns-netbox`](https://github.com/sukiyaki/octodns-netbox): [NetBox](https://github.com/netbox-community/netbox) source.
+  - [`jcollie/octodns-netbox-dns`](https://github.com/jcollie/octodns-netbox-dns): [NetBox-DNS Plugin](https://github.com/auroraresearchlab/netbox-dns) provider.
```
ross, line 366: Should this replace `sukiyaki/octodns-netbox` as it seems inactive?

LuPo: IMHO sukiyaki/octodns-netbox source is still valuable. It has different scope as it gets the entries from Netbox IPAM IP objects. It's approach is to use description field for storing a list of dns records, while NetBox provides nativelly DNS field that accepts a single valid dns name. With a simple patch one can adapt sukiyaki's source to use the new dns field. 

Netbox-DNS plugin instead deals with zones/records/NS/views and its OctoDNS provider makes it very much usable with OctoDNS.

I am using both (the source and the provider) because at the moment there is no relation between the IP addresses from NetBox IPAM (source for octodns-netbox) and NetBox DNS Plugin records (source/target for octodns-netbox-dns).

Issue #752, opened 2021-08-06
Ignore some records


wolfedale:
Hello!

There is a way to specify some sort of list of records that I would like to ignore? My problem is that some records I'm creating are from another place, another tool, calling GCP API. In this case every time when someone triggers octo-dns pipeline, it's trying to delete all of them. What's the best way to handle it?

ross:
For the most part octoDNS is intended to be used when it owns all the records in the zone and that's the setup which is known to work extremely well and reliably. 

There is an **alpha** processor, [`OwnershipProcessor`](https://github.com/octodns/octodns/blob/master/octodns/processor/ownership.py) which was created to explore this use case. I've not used it for anything "production" and don't have anything where I will in the near future so it's likely to stay alpha for the foreseeable future (until a handful of people have used it in real situations and reported back that it works well.)

It works by creating companion TXT records for each octoDNS managed record. So if you specify

```yaml
hello:
    type: A
    value: 1.2.3.4
```

It will create that record and also a `TXT` record named `_octodns.hello.` which is used to tell octoDNS that it "owns" `hello.` and can modify it. 

The main hiccup with this pattern and something it would be impossible for octoDNS to handle would be error cases like the record being created w/o the ownership TXT or someone or something deleting the ownership TXT. All octoDNS can do is look for and trust the ownership TXT record. If something goes wrong with it human intervention will be required to get things back into a known good state.

See https://github.com/octodns/octodns/pull/751#issue-704832484 for an example of how to configure a processor.

hsmade:
A use case for this, is when you have multiple data centers, and switch between them. This is typically done from monitoring systems. In this case, octoDNS should control the zone, but ignore changes on some records, as it would revert what the monitoring has switched.

hsmade:
I'd be happy to do a PR for this, if you are open for it.

ross:
As [stated above](https://github.com/octodns/octodns/issues/752#issuecomment-894237341) this is supported with the `OwnershipProcessor` so no PR is necessary.

johnlane:
We have a use-case where a zone in Amazon Route53 has some "static" records that we want to manage with OctoDNS and some "dynamic" records that will be managed by Amazon EKS. I'd like to specify somehow that OctoDNS should ignore records matching, say, a regex that are found to exist on the DNS but are not represented in the OctoDNS zone config (i.e. a record present on route53 that OctoDNS doesn't know about and would otherwise want to delete).

[...]

PR #932, opened 2022-09-11
Processor impl that filters on name

ross:
Docstring have more details, but bascially this adds two new filter style processors that work on the record names allowing or rejecting based on which type is selected things that match the prescribed patterns. Patterns can be exact string matches or full regexes. 

/cc https://github.com/octodns/octodns/issues/752#issuecomment-1239726527 which requested this @sjparkinson


sjparkinson:
Thank you @ross! The pattern matching might prove invaluable too.

PR #935, opened 2022-09-14
IDNA support for Record values holding fqdns

ross:
As of #915 Zone and Record names now handle IDNA values transparently. This PRs does the same for Record values that hold FQDNs: ALIAS, CNAME, DNAME, PTR, MX, NS, and SRV.

/cc https://github.com/octodns/octodns/pull/929 which this builds on, that added zone and record name IDNA support, this brings record values into the mix
/cc https://github.com/octodns/octodns/pull/915 which this replaces the need for (in a more full fledged way.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   decoded form. Both forms should be accepted in command line arguments.
   Providers may need to be updated to display the decoded form in their logs,
   until then they'd display the IDNA version.
+* IDNA value support for Record types that hold FQDNs: ALIAS, CNAME, DNAME, PTR,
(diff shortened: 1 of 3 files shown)


ross:
Not this does not display the decoded versions of the values when outputting plans or in logs.  Doing so in a backwards compatible way got pretty complicated pretty fast when trying to deal with keeping things backwards compatible for custom record types. https://github.com/octodns/octodns/compare/idna-values-display?expand=1 has something that works for core octoDNS as an example, but I'm not going forward with it. I plan to address this in the future as part of (hoped/planned) work on improved record displaying.

PR #940, opened 2022-09-16
Do not allow ALIAS to coexist, same as CNAME. Iprove error messages

ross:
CNAMEs were checked, but ALIASes were not. This also improves the error messages to always talk in terms of the record that's being added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@
 * Add TtlRestrictionFilter processor for adding ttl restriction/checking
 * NameAllowlistFilter & NameRejectlistFilter implementations to support
   filtering on record names to include/exclude records from management.
+* ALIAS should not be allowed to coexist with other records, same as CNAME,
+  fix/improve error messages when there are problems with them
 
 ## v0.9.19 - 2022-08-14 - Subzone handling
 
--- a/octodns/zone.py
+++ b/octodns/zone.py
@@ -99,17 +99,23 @@ class Zone(object):
             raise DuplicateRecordException(
                 f'Duplicate record {record.fqdn}, ' f'type {record._type}'
             )
-        elif not lenient and (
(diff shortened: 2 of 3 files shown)


yzguy (approved):

yzguy:
I think ALIAS's can exist with other records though, their very existence was really because CNAME can't have other records with it afaik.

ross:
> I think ALIAS's can exist with other records though, their very existence was really because CNAME can't have other records with it AFAIK.

octoDNS's concept of ALIAS only applies to the zone APEX, basically a top-level CNAME since the specs don't allow that.

> In the cases I'm aware of, the provider resolves the IP itself before returning it as an A record

ross:
Submitted too soon:

Yeah. That seems to be how they've implemented it. My thinking was that it was similar to CNAMEs in that it wouldn't make sense to have an A record live alongside an ALIAS at the top of a zone. 

I guess it may be a bit more nuanced than since they would live along side NS records. It's less clear what would/should happen with things like TXT records. Afaik those can't live along side of CNAMEs and I think if you query a TXT on a node with a CNAME it should follow and query the target for TXT. I would hope/assume that ALIAS does the same, but I guess that's up to the provider's implementation details.

ross:
Looks like at least one provider's test cases has an ALIAS coexisting https://github.com/octodns/octodns/actions/runs/3070343650/jobs/4959969244.

ross:
Some quick testing with NS1 and it looks like they don't support following an apex ALIAS to a TXT. Works fine for A/AAAA. 

You can create A and AAAA records alongside an ALIAS at which point it seems like the ALIAS just stops working and you get your A/AAAA answers. It may work for the other if you only have one. 

This is starting to feel like a minefield and something that's probably best left as-is. Closing. Thanks for the review @yzguy.

PR #942, opened 2022-09-22
Update CONTRIBUTING branch name to main

Ariana-Hlavaty-i2:
Update document to ensure correct branch name.


ross (approved):
Good catch. Thanks

Issue #954, opened 2022-11-08
Add root NS filter in the processors


sevencastles:
Hi,
I use octodns to sync between providers, from my own API to both PowerDNS and R53 (multi-authoritative). In my use case it was handy **not** to have root NS support, because I could set the NS manually in the PowerDNS provider, whilst R53 automatically created the right root NS records for a zone. Now that the support is introduced, R53 zones gets the root NS from my API, which breaks lots of stuff. 

I saw that octodns can filter the types of record being synced. I would like to keep syncing NS records (for zones delegation) but also to not sync root NS records. How could I achieve that? Is it possible to exclude the root NS sync somehow (either by a filter processor or by making the support optional)? 

Thanks in advance

ross:
> from my own API

What do you mean by your own API?  

There's a `supports_root_ns` parameter to `YamlProvider` that if configured to `False` would omit root NS records configured via yaml and then leave them unmanaged.

If you have a custom provider code pulling data from your own API you can set `SUPPORTS_ROOT_NS=False` on that provider to accomplish the same.

I can look at adding a processor that filters them out as well, but just want to nail down the details of your use case to make sure that it'd cover it.

sevencastles:
Hi @ross . I replied [here](https://github.com/octodns/octodns/pull/955#issuecomment-1311822243). Thanks for taking care of this!

PR #955, opened 2022-11-09
Add IgnoreRootNsFilter w/tests

ross:
Simple filter to ignore Root NS records that can be used in cases where one or more providers involved in a sync support Root NS records but the user wishes to leave them unmanaged.

/cc Fixes https://github.com/octodns/octodns/issues/954 @sevencastles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v1.0.0.rc0 - 2022-??-?? - First of the ones
+
+#### Noteworthy changes
+
+#### Stuff
+
+* Added simple IgnoreRootNsFilter
+
 ## v0.9.21 - 2022-10-16 - Last of the oughts
 
 * Shim AxfrSource and ZoneFileSource post extraction into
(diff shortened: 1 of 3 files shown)


viranch:
I may be recalling wrong, but didn't we build root NS support in a way that if the YAML doesn't include root NS record, it will be left untouched in the target?

sevencastles:
Hi @ross and @viranch, sorry for my late reply!
My use case is using an internal API (via a custom provider) as source, and PowerDNS and Route53 as targets. I can confirm that the situation is exactly as Ross depicted: since we don't use YAML as source we found ourselves struggling. I worked around the issue by adding an extra option to the custom provider to skip any root NS record. It works so far, but I am glad you've added a filter and we'll make sure to use it in the near future.

Thanks again for the quick support!

ross:
> I worked around the issue by adding an extra option to the custom provider to skip any root NS record.

If the provider has `SUPPORTS_ROOT_NS=False` you should get the same result.

I still think this filter makes sense for those using non-custom providers as sources that don't give a choice in root NS support.

sevencastles:
The point is that I am fine with syncing root NS records from my source to PowerDNS, but not to Route53. Perhaps I could have made `SUPPORTS_ROOT_NS` a variable, but my approach back then looked cleaner. Anyway I'll make sure to use the filter functionality. Thanks

PR #956, opened 2022-11-15
1.0 RC: Shim removal

ross:
On the way to cutting a 1.0 RC, remove the shims that have been deprecated for a long time now.

* Provider, Source, and Processor shims removed, they've been warnings for > 1yr.  Everything should be using and referring to provider-specific modules now.
* Zone._remove_record shim removed in favor of long standing Zone.remove_record


yzguy (approved):
Can't believe it's already been a year!

ross:
Looks like a lot of module tests use `Zone._remove_record` and haven't been udpated. Given that it's just a method alias I'm fine with leaving it lying around, but 😦.

ross (commented):
Bump for CI

PR #959, opened 2022-11-20
Manager.get_zone used everywhere rather than direct create

ross:
Minor cleanup/refactoring around the way Manager creates Zone objects, always uses `get_zone`.


ross (commented):
/ci modules

PR #961, opened 2022-11-28
Add custom zone file name support

junnhy5:
We needed to split our area into several sub-files, we did not want to go through the area name for a given file since we have the following tree structure.

```bash
zones/example.org/root-records.yaml
zones/example.org/project1/records.yaml
zones/example.org/project2/records.yaml
zones/example.org/project3/example.org.yaml
```

And in zone config for example:

```yaml
---
providers:
  root-records:
    directory: zones/example.org/
    file_name: "root-record"
    default_ttl: 3600
    enforce_order: False
  project1:
    directory: zones/example.org/project1/
    file_name: "records"
    default_ttl: 60
  project2:
    directory: zones/example.org/project2/
    file_name: "records"
    default_ttl: 60
  project3:
    directory: zones/example.org/project3/
    default_ttl: 60
  route53:
    class: octodns_route53.Route53Provider
    access_key_id: env/AWS_ACCESS_KEY_ID
    secret_access_key: env/AWS_SECRET_ACCESS_KEY
    max_changes: 100

zones:
  example.org.:
    sources:
      - root-records
      - project1
      - project2
    targets:
      - route53
```

That's why I took the liberty of adding this feature which allows you to split an area into several sub-files when it is very large.

### DIFF
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -21,6 +21,10 @@ class YamlProvider(BaseProvider):
         class: octodns.provider.yaml.YamlProvider
         # The location of yaml config files (required)
         directory: ./config
+        # Specify a custom zone file name
+        # File present in the directory declared in "directory" without specifying the extension.
+        # (optional, default empty)
+        file_name: "test"
         # The ttl to use for records when not specified in the data
         # (optional, default 3600)
         default_ttl: 3600
@@ -105,11 +109,13 @@ class YamlProvider(BaseProvider):
     SUPPORTS_DYNAMIC = True
     SUPPORTS_POOL_VALUE_STATUS = True
     SUPPORTS_MULTIVALUE_PTR = True
+    FILE_NAME = ""
 
     def __init__(
         self,
         id,
         directory,
+        file_name="",
         default_ttl=3600,
         enforce_order=True,
         populate_should_replace=False,
@@ -129,6 +135,7 @@ class YamlProvider(BaseProvider):
             populate_should_replace,
         )
         super().__init__(id, *args, **kwargs)
+        self.file_name = file_name
         self.directory = directory
         self.default_ttl = default_ttl
         self.enforce_order = enforce_order
@@ -198,7 +205,11 @@ class YamlProvider(BaseProvider):
 
         before = len(zone.records)
         utf8_filename = join(self.directory, f'{zone.decoded_name}yaml')
-        idna_filename = join(self.directory, f'{zone.name}yaml')
+
+        if self.file_name == "":
+            idna_filename = join(self.directory, f'{zone.name}yaml')
+        else:
+            idna_filename = join(self.directory, f'{self.file_name}.yaml')
 
         # we prefer utf8
         if isfile(utf8_filename):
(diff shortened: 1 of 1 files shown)


ross:
Hi @junnhy5. The setup you're after is already supported by octoDNS. The directories and files would look like:

```
zones/example.org/example.org.yaml
zones/example.org/project1/example.org.yaml
zones/example.org/project2/example.org.yaml
zones/example.org/project3/example.org.yaml
```

Otherwise your config should work as-is. As you saw multiple sources just "append" to the zone one after another. The reason it's preferable to use the standard setup rather than a hard-coded filename is that the filename would only allow one zone to exist per provider since the directory and exact filename are specified. 

For an example say you have two zones that you want to break up by project, example.org and example.com. The files would look like:

```
zones/example.org/example.org.yaml
zones/example.org/example.com.yaml
zones/example.org/project1/example.org.yaml
zones/example.org/project1/example.com.yaml
zones/example.org/project2/example.org.yaml
zones/example.org/project2/example.com.yaml
zones/example.org/project3/example.org.yaml
zones/example.org/project3/example.com.yaml
```

With the `file_name` parameter you'd have to create a completely separate set of providers to support this where as the standard way supports any number of zones.

junnhy5:
Our use case is a bit particular, indeed, we would need this directive to separate the root records from the rest while sharing inter-project information.

Hence the interest of having added this functionality.

Moreover, having example.org.yaml in each folder may lose people who will have to contribute on the OctoDNS repository we have.

ross:
Given the specifics of the use case the functionality is probably best implemented with a custom provider. 

As-is that'd be a bit more effort than ideal so it may make sense to add a `get_filenames` helper method to `YamlProvider`  
That'd probably look something like:

```python
    def get_filenames(self, zone):
        utf8_filename = join(self.directory, f'{zone.decoded_name}yaml')
        idna_filename = join(self.directory, f'{zone.name}yaml')
        return utf8_filename, idna_filename
```

And then these two lines could be replaced with:
https://github.com/octodns/octodns/blob/498d7928235c84c42053ba091ffc7a000bb0c9bf/octodns/provider/yaml.py#L200-L201.

```python
        utf8_filename, idna_filename = self.get_filenames(zone)
```

That should allow you to inherit from `YamlProvider` to add your `file_name` param and then override `get_filenames` to add your specific logic. 

Lmk if you'd like me to get the PR together for the refactoring or if you'd like to do it.

ross:
> I just updated the PR in this sense, does this way of overloading the value fit you?

I'm fine with repurposing the PR since the discussion/context will be here with the changes. The refactoring bits look good.

In the comment [here](https://github.com/octodns/octodns/pull/961#issuecomment-1329779164) I was saying that the refactoring would be the only portion to go into octoDNS itself though. 

You would then create a custom provider in your setup to implement the non-standard logic, something like:

```python

class NamedFileYamlProvider(YamlProvider):

    def __init__(self, id, file_name, *args, **kwargs):
        super().__init__(id, *args, **kwargs)
        self.file_name = filename

    def get_filenames(self, zone):
        # your custom handling/logic goes here
        ...
```

ross:
As of https://github.com/octodns/octodns/pull/966, the trivial provider above should be possible https://github.com/octodns/octodns/pull/961#issuecomment-1330942960.

Issue #967, opened 2022-12-22
Support DS record


baest:
I know this is properly not the most commonly used record type, but we have DS records in our record set and it will be nice to have it be supported.
Example content of a DS record:
`2371 13 2 499E0967FB04F531204C56F6E8863FFF7CCEEED1469E8F26D85E2E42 54DD0CBB`

yzguy:
Relevant Issue about DNSSEC support: #176

baest:
Basically the DS record is put in parent zone and is used to validate the DNSKEYs in your zone. So DS will normally not be a record used, since it will be uploaded to the domain registrar. But in the case where you have signed a sub-zone then it becomes relevant. DNSKEYs are often auto generated I think (and that is what #176 says), so for me it is no problem at all to exclude them.
Again this is possible an edge use-case, but on the other hand it shouldn't require a lot of work so could be considered. I can possibly make a PR if there is a chance of it getting merged :)

ross:
Not opposed to supporting the `DS` record type to support the use case of publishing the hash of the KSK. 

I've looked at overall DNSEC support a number of times, documented in https://github.com/octodns/octodns/issues/176, but there were also internal GitHub investigations/discussion. In the end DNSEC seems pretty flawed opertationally, at least in the way it's been implemented. I wish providers would let you manage your own keys and just provide them to them. Then we could actually get working DNSSEC across multiple providers. 

Anyway, happy to review a PR if one pops up. Otherwise I've left this tab open as a TODO list item when I get around to it.

ross:
This is ready for people to submit PRs for providers to add support.

PR #968, opened 2023-01-02
Add support for DsRecord type

ross:
Added new `DsRecord` type (provider support will be added over time)

/cc Fixes https://github.com/octodns/octodns/issues/967 
/cc https://www.rfc-editor.org/rfc/rfc4034.html#section-2.1
/cc @baest @yzguy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 #### Stuff
 
+* Added new DsRecord type (provider support will be added over time)
 * Added simple IgnoreRootNsFilter
 * Minor refactor on YamlProvider to add get_filenames making it a bit easier to
   create specialized providers inheriting from it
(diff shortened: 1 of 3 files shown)


viranch:
I wonder if (in another PR) it might be possible to split this rather big `record/__init_.py` into individual files for each record type.

baest:
Looks very good 👍 Thanks for doing this (and so quickly too)

ross:
> I wonder if (in another PR) it might be possible to split this rather big `record/__init_.py` into individual files for each record type.

Yeah. I actually started breaking it up in a branch a while back and either got distracted or ran into a hiccup or something as I didn't finish it off. I also want to break up/clean up the tests as well.

ross:
Since this is self contained, going to go ahead and merge it. Will poke around at breaking things up.

Issue #977, opened 2023-02-10
SPF processor to error on too many DNS lookups


sjparkinson:
We've been looking into validating our TXT SPF records to ensure they are valid and don't exceed the limits.

I've found https://www.mailhardener.com/blog/spf-lookup-limit-explained a useful reference.

Using a new processor seemed the best approach and I wanted to reach out to see if you think it'd be valuable to include here? No worries if not, I can also put a new repository together in the Financial-Times org.

Here's what I came up with in the initial go at this, not by best code but it does validate the approach, and it depends on `dnspython`:

```python
from logging import getLogger

import dns.resolver

from ..record import TxtRecord
from ..zone import Zone
from .base import BaseProcessor, ProcessorException


class SpfLookupException(ProcessorException):
    pass


class SpfProcessor(BaseProcessor):
    log = getLogger('SpfProcessor')

    def __init__(self, name):
        self.log.debug(f"SpfProcessor: {name}")
        super().__init__(name)

    def process_source_zone(self, zone: Zone, *args, **kwargs):
        self.log.debug(f"process_source_zone: {zone.name}")
        for record in zone.records:
            if record._octodns.get('lenient'):
                continue

            if type(record) == TxtRecord and any(
                value.startswith('v=spf1') for value in record.values
            ):
                lookups = 0
                includes = []

                for value in record.values:
                    includes.extend(
                        [
                            d.removeprefix('include:')
                            for d in value.split(' ')
                            if d.startswith('include:')
                        ]
                    )
                    self.log.debug(
                        f"process_source_zone: found {includes} includes in {record.fqdn}"
                    )

                while (
                    lookups < 10
                    and len(includes) > 0
                    and lookups < len(includes)
                ):
                    answer = dns.resolver.query(includes[lookups], 'TXT')
                    lookups += 1

                    self.log.debug(
                        f"{lookups} lookups of 10 allowed for {record.fqdn}"
                    )

                    for txt in answer:
                        if txt.to_text().startswith('"v=spf1'):
                            self.log.debug(
                                [
                                    d.removeprefix('include:')
                                    for d in txt.to_text().split(' ')
                                    if d.startswith('include:')
                                ]
                            )
                            includes.extend(
                                [
                                    d.removeprefix('include:')
                                    for d in txt.to_text().split(' ')
                                    if d.startswith('include:')
                                ]
                            )

                if lookups == 10:
                    raise SpfLookupException(
[...]

PR #978, opened 2023-02-16
Add SPF processor to validate SPF values in TXT records

sjparkinson:
Use this processor to validate SPF values in TXT records

The processor will error on...

* more than one SPF value in a TXT record
* more than 10 DNS lookups caused by an SPF value
* any reference to the deprecated `ptr` mechanism

The processor uses recursion to handle counting the number of DNS lookups caused by `include` mechanisms. I'd be happy to hear ideas on how that could be simplified!

Another validation I'd be interested in, but is likely separate to this, is a limit on the number of values in any TXT record. I'm aware of issues occuring when the value for a DKIM key is beyond the ~50th element in a TXT record.

**Usage:**

```yaml
processors:
  spf:
    class: octodns.processor.spf.SpfDnsLookupProcessor

zones:
  example.com.:
    sources:
      - config
    processors:
      - spf
    targets:
      - route53
```

**Todo:**

- [x] Add documentation to the class
- [x] Test for more than one level of `include` mechanism
- [x] Handle chunked values in resolver answer, e.g. from `dig _spf.mailgun.org TXT`
- [x] ~Check for TXT record size, https://datatracker.ietf.org/doc/html/rfc7208#section-3.4~ Doesn't seem much of a concern? https://serverfault.com/questions/1000170/size-of-spf-with-other-txt-records
- [x] Handle the `ptr` mechanism. Looks like it is deprecated https://datatracker.ietf.org/doc/html/rfc7208#section-5.5, could be an error?

### References

* #977

### DIFF
--- a/tests/test_octodns_processor_spf.py
+++ b/tests/test_octodns_processor_spf.py
@@ -10,9 +10,83 @@ from octodns.zone import Zone
 
 
 class TestSpfDnsLookupProcessor(TestCase):
+    def test_get_spf_from_txt_values(self):
+        processor = SpfDnsLookupProcessor('test')
+        record = Record.new(
+            Zone('unit.tests.', []),
+            '',
+            {'type': 'TXT', 'ttl': 86400, 'values': ['v=DMARC1\; p=reject\;']},
+        )
+
+        self.assertEqual(
+            'v=spf1 include:_spf.google.com ~all',
+            processor._get_spf_from_txt_values(
+                [
+                    'v=DMARC1\; p=reject\;',
+                    'v=spf1 include:_spf.google.com ~all',
+                ],
+                record,
+            ),
+        )
+
+        with self.assertRaises(SpfValueException):
+            processor._get_spf_from_txt_values(
+                [
+                    'v=spf1 include:_spf.google.com ~all',
+                    'v=spf1 include:_spf.google.com ~all',
+                ],
+                record,
+            )
+
+        self.assertEqual(
+            'v=spf1 include:_spf.google.com ~all',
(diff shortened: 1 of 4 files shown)


ross (approved):
Looks good. Read over it and nothing specific jumped out at me once I was correctly seeing out `lookups` was being managed/counted.

tests/test_octodns_processor_spf.py
```diff
@@ -0,0 +1,403 @@
...
+        # Used in logging
+        record = Record.new(
+            Zone('unit.tests.', []),
+            '',
+            {'type': 'TXT', 'ttl': 86400, 'values': ['']},
+        )
+
+        self.assertIsNone(
+            processor._get_spf_from_txt_values(
+                record, ['v=DMARC1\; p=reject\;']
```
ross, line 26: Looks like py3.7 and 3.8 aren't happy with the escape sequences in here: 

```
=============================== warnings summary ===============================
tests/test_octodns_processor_spf.py:26
  /home/runner/work/octodns/octodns/tests/test_octodns_processor_spf.py:26: DeprecationWarning: invalid escape sequence \;
    record, ['v=DMARC1\; p=reject\;']
...
```

ross:
> Yeah, that I'm not the biggest fan of either, took two of us to figure it out to be honest!

Only other option that came to mind was using a queue/stack to push lookups onto and pop them off of. I believe that would let things avoid the recursion and just do a `while stuff_to_process`, but without coding it out no clue if that'd be preferable.

ross (approved):

PR #979, opened 2023-02-16
Add `Financial-Times/octodns-fastly` to the resources

sjparkinson:
Also fixes an anchor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Most of the things included in OctoDNS are providers, the obvious difference bei
 
 The `class` key in the providers config section can be used to point to arbitrary classes in the python path so internal or 3rd party providers can easily be included with no coordination beyond getting them into PYTHONPATH, most likely installed into the virtualenv with OctoDNS.
 
-For examples of building third-party sources and providers, see [Related Projects & Resources](#related-projects--resources).
+For examples of building third-party sources and providers, see [Related Projects & Resources](#related-projects-and-resources).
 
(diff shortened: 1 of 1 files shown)


ross (approved):
LGTM. Happy to host it under the `octodns/` if you'd prefer.

ross:
> That would be good actually. Might encourage more contributions in the long run.

I'll get a PR going in https://github.com/octodns/terraformed in the next couple of days and we can get it moved after that.

ross:
So I have things set up, but realized it may be preferable to move the repo rather than create a new one and push the code up to i since there's historic PRs etc. Do you have a preference? Do you have enough perms on the FT org to do that if granted enough on the octoDNS side? If it's more trouble than it's worth we can just push the code and go from there.

ross:
> Seems worth preserving the history over one big "initial commit"!

Ah, I was meaning `git push` which would keep the code history, but it would lose the PRs. I'll delete the empty repo over there and figure out what's required to let you move it.

ross:
OK. Deleted and invited you to the org with what supposedly should be enough perms to transfer the repo over. Once it's moved we'll get all the details sorted.

ross:
Cool. I'll open up PRs/issues in https://github.com/octodns/octodns-fastly for anything further. Thanks.

ross:
Can you email me at rwmcfa1@gmail.com. I have a question that doesn't belong in an issue :grin:

PR #980, opened 2023-02-17
Update README sample to reflect extracted providers

adutton:
https://github.com/octodns/octodns#updating-to-use-extracted-providers changed the class names to use but that wasn't reflected in the sample code


ross (approved):
Doh

PR #983, opened 2023-02-20
Auto arpa tweaks

ross:
Review feedback from https://github.com/octodns/octodns/pull/974#pullrequestreview-1264833837

/cc https://github.com/octodns/octodns/pull/974#pullrequestreview-1264833837 @viranch

### DIFF
--- a/docs/auto_arpa.md
+++ b/docs/auto_arpa.md
@@ -21,13 +21,21 @@ manager:
     ttl: 1800
 ```
 
-Once enabled a singleton `AutoArpa` instance, `auto-arpa`, will be added to the pool of providers and globally configured to run as the very last global processor so that it will see all records as they will be seen by targets. Further all zones ending with `arpa.` will be held back and processed after all other zones have been completed so that all `A` and `AAAA` records will have been seen prior to planning the `arpa.` zones. 
+Once enabled a singleton `AutoArpa` instance, `auto-arpa`, will be added to the pool of providers and globally configured to run as the very last global processor so that it will see all records as they will be seen by targets. Further all zones ending with `arpa.` will be held back and processed after all other zones have been completed so that all `A` and `AAAA` records will have been seen prior to planning the `arpa.` zones.
 
 In order to add `PTR` records for a zone the `auto-arpa` source should be added to the list of sources for the zone.
 
 ```yaml
+# Zones are matched on suffix so `0.10.in-addr.arpa.` would match anything
+# under `10.0/16` or `0.8.e.f.ip6-.arpa.` would match any IPv6 address under
+# `fe80::`, 0.0.10 here matches 10.0.0/24.
 0.0.10.in-addr.arpa.:
   sources:
+    # In most cases you'll have some statically configured records combined in
+    # with the auto-generated records as shown here, but that's not strictly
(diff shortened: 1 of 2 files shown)


docs/auto_arpa.md
```diff
@@ -14,16 +14,28 @@ Alternatively the value can be a dictionary with configuration options for the A
 ---
 manager:
   auto_arpa:
+    # Replace duplicate records rather than throw an error, default is False
+    # which throws an error
+    replace: False
```
viranch, line 19: It would be better to offer a choice whether the config takes precedence or auto-generated zone. Perhaps something to do with the order of sources?

viranch:
LGTM after that one comment!

viranch:
Long term, It will be nice to provide a mechanism to write the auto-generated arpa zone into a config and then octodns treats that as the source. The zone yamls have the potential to be treated as the single source of truth for all DNS records and the auto arpa zone will be missing from that source of truth.

So instead of auto-arpa -> target, we'd have auto-arpa -> yaml -> target; wherein auto-arpa -> yaml part can be done locally by the user followed by usual git commit. I suppose it could be done with a combination of octodns zone configs (`sources: auto-arpa; targets: config` then `sources: config; targets: $provider`) but a clean documented mechanism could be valuable.

docs/auto_arpa.md
ross: Yeah. `YamlProvider` supports the replace idea as well so that would be better managed by flipping the order rather than having to add further configuration to all the providers involved.

That did actually make me look at the naming in `YamlProvider` and realize that it was `populate_should_replace` so I've updated this PR to use the same naming here for consistency's sake.

ross:
> Long term, It will be nice to provide a mechanism to write the auto-generated arpa zone into a config and then octodns treats that as the source. 

That's an interesting way to static-ify dynamic sources. Should be easily possible already by using the "config" source as a target for the arpa zones, maybe in the same octoDNS config or perhaps a dedicated one (safer/simpler.) 

> The zone yamls have the potential to be treated as the single source of truth for all DNS records and the auto arpa zone will be missing from that source of truth.

Yeah. That'd come down to how static an env you're running. GitHub's was WAY too dynamic to have the goal of everything being in config files in git. There were a lot of dynamic sources: hosts, k8s sources, etc. and we couldn't require going through the whole PR flow just to get new stuff to show up.

> So instead of auto-arpa -> target, we'd have auto-arpa -> yaml -> target; wherein auto-arpa -> yaml part can be done locally by the user followed by usual git commit. I suppose it could be done with a combination of octodns zone configs (sources: auto-arpa; targets: config then sources: config; targets: $provider) but a clean documented mechanism could be valuable.

Yeah. That's definitely a case where documenting it makes more sense than adding the complications to do something "automatically," similar to the replace ordering stuff up above. There's no end to the documentation that should exist for octoDNS :-/

docs/auto_arpa.md
ross: That doesn't really seem directly related to auto-arpa as it would apply to situations with YamlProvider as well so that assumption was that it'd be documented in another location that was more generally applicable once that's figured out. I have hopes to spend time on documentation/examples in the ~future.

PR #984, opened 2023-02-24
prepend zone_name with . to ensure it matches on address boundary

dethmetaljeff:
Existing logic matches with an endswith which will incorrectly match zones.  Zone matching needs to be done with respect to the subnet.

example:
66.5.20.10.in-addr.arpa. matches zone 0.10.in-addr.arpa.

but it should actually match 20.10.in-addr.arpa. or 5.20.10.in-addr.arpa., etc

### DIFF
--- a/octodns/processor/arpa.py
+++ b/octodns/processor/arpa.py
@@ -48,7 +48,7 @@ class AutoArpa(BaseProcessor):
         zone_name = zone.name
         n = len(zone_name) + 1
         for arpa, fqdns in self._records.items():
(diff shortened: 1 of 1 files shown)


ross (commented):
Nice catch. This is probably a good case for a couple of unit test calls to make sure that things like your example `20.10.*` don't match.

ross (approved):
Thanks. Tests added.

dethmetaljeff:
> 

Thank you for adding the tests!

Issue #989, opened 2023-03-26
Explore a validation for dynamic rules that encourages them to go from most to least specific


ross:
https://github.com/octodns/octodns-ns1/issues/36 is tracking some potential issues with dynamic rules and geo's with NS1, but while looking into it I realized the example config wasn't ordered most to least specific which would generally be the "best practice." 

I'm not even sure that's said anywhere. If not it should be in the dynamic doc. It'd also be worth exploring a validation check for the ordering of dynamic rules to check their order. The validation can always be disabled if someone has a reason to do otherwise.

- [x] Make sure the best practices are documented
- [x] Explore a validation that checks for reusing targeted locs
- [x] Explore a validation that checks for targeting a more specific loc inside of a previous one
- [x] Explore a validation that verifies there's a "catch-all" rule at the end

/cc https://github.com/octodns/octodns-ns1/issues/36

viranch:
Talking of best practice, I would also make a case for automatically appending a catch-all rule (in absence of one) which points to the top-level `value`/`values`.

ross:
> Talking of best practice, I would also make a case for automatically appending a catch-all rule (in absence of one) which points to the top-level `value`/`values`.

Feel like that'd best be a warning/error as there aren't any other cases I can think of where octoDNS modifies things in that sort of way.

ross:
2 of the 4 checkboxes are done in https://github.com/octodns/octodns/pull/991, will pick up the other 2 in there later/tomorrow.

istr:
I did some more research and thought over the discussion we had around the issues in the ns1 provider.

As a conclusion, I think that the concept of "sorted `rules`" does not make much sense, and is -- at best -- an implementation artifact. The _declarative semantics_ should reflect a configuration that is unambiguous and provides a straight "most specific rule wins" logic. Warnings generated if there are issues should reflect this logic as well.

Here are some reasons:

1. none of the currently implemented providers with `dynamic` facilities (apart from ns1) have a logic that would allow "ordered" rulesets
    - route53 explicitly selects the most specific target. "If you create separate records for overlapping geographic regions—for example, one record for North America and one for Canada—priority goes to the smallest geographic region."
      https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-geo.html
      Any kind of rule order here is pointless in the good case and misleading in bad cases.
     - ns1 takes the order of rules into account. However, tracking is _not natively supported_ (a workaround with annotations is used to track and sync the order)
[...]

PR #991, opened 2023-03-28
Improved dynamic validations

ross:
This PR validates that rules do not target geos that have already been used and ensures that more specific location within a previously targeted geo cannot be targeted in rules.

/cc https://github.com/octodns/octodns/issues/989 which this partially addresses
/cc @viranch


istr:
I would still argue that the "most natural" _last catchall_ rule in a record is the primary value.

What you are proposing is along the lines of
```yaml
geo:
  - dynamic:
    pools:
      eu:
        values:
          - status: up
            value: 1.2.3.4
      eu_E:
        values:
          - status: up
            value: 4.5.6.7
    rules:
      - geos:
          - EU-PL
          - EU-LT
        pool: eu_E
      - pool: eu  # catch-all, no geo, best practice
  ttl: 60
  type: A
 ```
While this is what matches the _current implementation_ of some providers, I would still like to propose a convention that does not need to impose another constraint on `rule` entries:

```yaml
geo:
  - dynamic:
    pools:
      eu_E:
        values:
          - status: up
            value: 4.5.6.7
    rules:
      - geos:
          - EU-PL
          - EU-LT
        pool: eu_E
  ttl: 60
  type: A
  values:
    - 1.2.3.4  # mandatory global catch all value
```

This construct already works for _some_ provider implementations (for example, constellix), but not for all (for example, ns1).

So, instead of trying to check for a catch-all pool without `geos` I would propose to

- make `geos` mandatory for each `rule` entry
- have at least one mandatory `value` entry for `dynamic` records that serves as a global catch-all for all providers.

I think that the implementation of this logic is less complicated, less error-prone, more easy to read and comprehend and unambiguous. Just my 2 cents, ymmv. :slightly_smiling_face:

istr:
As an addition: with a construct that has a _mandatory value_, the fall-back to providers that do not provide dynamic records _at all_ is really straight-forward. Just issue a warning that the provider does not support `dynamic` and proceed with the plain old `values`. That way you can think of `dynamic` as a "nice-to-have" add-on. No need to try to figure out the default from an unsupported feature for providers that don't support it to start with...

istr:
> > As an addition: with a construct that has a _mandatory value_, the fall-back to providers that do not provide dynamic records _at all_ is really straight-forward. Just issue a warning that the provider does not support `dynamic` and proceed with the plain old `values`. That way you can think of `dynamic` as a "nice-to-have" add-on. No need to try to figure out the default from an unsupported feature for providers that don't support it to start with...
> 
> This is the current behavior/intention. This discussion does make me realize that the main `values` section could probably use a better example and some comments about this as it's probably not laid out well anywhere.

Good to know. Then at least _some_ implementations are faulty and do not follow that intention. For example, the ns1 provider simply _drops_ or _overwrites_ the values given in `values` with the default pool. That is to say: in the following example the `values` will never be used for ns1 (the implementation _replaces_ them with the default pool), whereas constellix sets them as a global fallback (as intended).

```yaml
geo:
  - dynamic:
      pools:
        eu:
          values:
            - status: obey
              value: 1.2.3.4
        eu_E:
          values:
            - status: obey
              value: 4.5.6.7
      rules:
        - geos:
            - EU-PL
            - EU-LT
          pool: eu_E
        - pool: eu  # catch-all, no geo, best practice
    ttl: 60
    type: A
    values:
      - 9.8.7.6
      - 5.4.3.2
      - 4.3.2.1
      - 1.1.1.1
```

That is to say that what is given as an example here https://github.com/octodns/octodns/blob/main/docs/dynamic_records.md _only_ works because the default pool and the `values` share _the same_ (redundant) pool of IPs.

The argument with "everything falls back to a single point in case of failure" falls apart, because you can still have _multiple values_ within `values`.

> The "best practice" is to have all of the IPs or at least a large subset of them in the ultimate fallback list of values. Thus if all health checks are failing all (or a lot) of the global IPs are returned and requests can just try and see if any of them work.

Ok, but why not simply put them into `values` instead of a special fall-back pool that would need to be configured with an `up` policy anyway to make sense? The effect is the same: if monitoring fails, everything falls back to some non-monitored default. Having this in `values` has the great advantage that it is completely in line with a non-dynamic setup.

Maybe there is something I overlook in the two-step concept of having 

- a monitored catch-all pool that might fail
- a global fall-back that catches failures of all pools including a failing catch-all pool?

What would you gain with this?

In any scenario I can think of, you would either make the catch-all pool non-monitored (and can then ditch the fall-back) or simply ditch the catch-all pool and use the non-monitored global fall-back...

In case you configure your fall-back with _all_ of the IPs of _all_ your pools, then you would have the catch-all pool in there as well (as a subset), so redundancy again.

ross:
> Then at least _some_ implementations are faulty and do not follow that intention. For example, the ns1 provider simply _drops_ or _overwrites_ the values given in `values` with the default pool. That is to say: in the following example the `values` will never be used for ns1 (the implementation _replaces_ them with the default pool), whereas constellix sets them as a global fallback (as intended).

It's been years since I was deeply in the details of the NS1 dynamic code, but that doesn't match my memory. 

`record.values` are pulled out into `default_values`

https://github.com/octodns/octodns-ns1/blob/bc542be4ac73d3914adcfd5d1ce2f56ff484e119/octodns_ns1/__init__.py#L1453-L1457

which are passed into `_add_answers_for_pool` for each region. that method adds them as the final set of answers w/o a health check (ultimate fallback) and marks them with the note `--default--`. The code above that walks through the pools in fallback order from the region's starting pool down to the catch-all. All of which should be health checked.

https://github.com/octodns/octodns-ns1/blob/bc542be4ac73d3914adcfd5d1ce2f56ff484e119/octodns_ns1/__init__.py#L1310-L1322

There's code in the opposite direction that pulls them out and turns them into the `record.values` during populate.

> That is to say that what is given as an example here https://github.com/octodns/octodns/blob/main/docs/dynamic_records.md _only_ works because the default pool and the `values` share _the same_ (redundant) pool of IPs.

That's not my read of the code atm. 

For a concrete example:

```yaml
aaaa:
  dynamic:
    pools:
      ams:
        values:
          - value: 2601:642:500:e210:62f8:1dff:feb8:9471
      iad:
        values:
          - value: 2601:642:500:e210:62f8:1dff:feb8:9472
          - value: 2601:642:500:e210:62f8:1dff:feb8:9473
      lax:
        values:
          - value: 2601:642:500:e210:62f8:1dff:feb8:9474
      sea:
        values:
          - value: 2601:642:500:e210:62f8:1dff:feb8:9475
            weight: 1
          - value: 2601:642:500:e210:62f8:1dff:feb8:9476
            weight: 2
    rules:
      - geos:
        - EU-GB
        pool: lax
      - geos:
        - EU
        pool: ams
      - geos:
          - NA-US-CA
          - NA-US-NC
          - NA-US-OR
          - NA-US-WA
        pool: sea
      - pool: iad
  type: AAAA
  values:
    - 2601:642:500:e210:62f8:1dff:feb8:947a
    - 2601:644:500:e210:62f8:1dff:feb8:947a
```

The results of `_add_answers_for_pool` for the "ams" set are:

```python
[{'answer': ['2601:642:500:e210:62f8:1dff:feb8:9471'],
  'meta': {'note': 'fallback: from:ams__georegion pool:ams',
           'priority': 1,
           'up': {'feed': '5374aadbfcf26124dd0cf060'},
           'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:642:500:e210:62f8:1dff:feb8:947a'],
  'meta': {'note': 'from:--default--', 'priority': 2, 'up': True, 'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:644:500:e210:62f8:1dff:feb8:947a'],
  'meta': {'note': 'from:--default--', 'priority': 2, 'up': True, 'weight': 1},
  'region': 'ams__georegion'}]
```

In this case there's no `fallback` so things go straight from the pool's ips (health checked) to the record.values (not health checked.)

If I throw a `fallback: iad` into the `ams` pool the answers become:

```python
[{'answer': ['2601:642:500:e210:62f8:1dff:feb8:9471'],
  'meta': {'note': 'fallback:iad from:ams__georegion pool:ams',
           'priority': 1,
           'up': {'feed': '5374aadbfcf26124dd0cf060'},
           'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:642:500:e210:62f8:1dff:feb8:9472'],
  'meta': {'note': 'fallback: from:ams__georegion pool:iad',
           'priority': 2,
           'up': {'feed': 'acede3cc6fd66509ab695559'},
           'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:642:500:e210:62f8:1dff:feb8:9473'],
  'meta': {'note': 'fallback: from:ams__georegion pool:iad',
           'priority': 2,
           'up': {'feed': 'b86891a7ad70c8e784549be8'},
           'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:642:500:e210:62f8:1dff:feb8:947a'],
  'meta': {'note': 'from:--default--', 'priority': 3, 'up': True, 'weight': 1},
  'region': 'ams__georegion'},
 {'answer': ['2601:644:500:e210:62f8:1dff:feb8:947a'],
  'meta': {'note': 'from:--default--', 'priority': 3, 'up': True, 'weight': 1},
  'region': 'ams__georegion'}]
```

ross:
> Ok, but why not simply put them into `values` instead of a special fall-back pool that would need to be configured with an `up` policy anyway to make sense? The effect is the same: if monitoring fails, everything falls back to some non-monitored default. Having this in `values` has the great advantage that it is completely in line with a non-dynamic setup.

The `record.values` IPs are the one added to the answers as the lowest priority w/o a health check so that they're always served if everything above them, with health checks normally, is failing. 

I think I've confused things by talking about fallbacks and routing rules at the same time. `catch-all` is a rules concept and default values are a fallback mechinism.

A request will always match at most 1 **rule**. That will direct things to a specific **pool** and from there zero or more fallbacks will happen based on the `fallback` pool option and pools health status.

The `catch-all` rule is the thing that applies to requests that don't match any other rules. There are (potentially) IPs that don't geo-locate at all b/c they're unknown or special cases.

istr:
> I just added a diagram showing the example records rules and pools flow.

Great, that helps _a lot_ -- as well as the new comments in the documentation. :+1:

viranch:
This makes sense in practice. One theoretical problem I see is in the following setup:

```
...
    rules:
      - geos:
        - EU-GB
        - AS
        pool: one
      - geos:
        - AS-IN
        - EU
        pool: two
...
```

IIRC multiple rules cannot point to the same pool, so the above can't be split into country-only followed by continent-only rules.

ross:
OK. I think things have settled here and what's in this PR lines up with currently recommended/implemented best practices. There's some chance that may change in the future, but until then this docs the way things (are supposed to) work.

ross:
Doh, since this didn't get a review all the module tests didn't run, looks like some of the specific testing cases fail in 
* azure - https://github.com/octodns/octodns/actions/runs/4623780200/jobs/8177940871?pr=995. 
* gcore - https://github.com/octodns/octodns/actions/runs/4623780200/jobs/8177941809?pr=995
* ns1 - https://github.com/octodns/octodns/actions/runs/4623780200/jobs/8177942135?pr=995

Will dig into those ~soon

PR #993, opened 2023-04-02
Dynamic subnets

viranch:
Support for subnet targeting in addition to geofencing. Ns1Provider support in https://github.com/octodns/octodns-ns1/pull/44


octodns/record/dynamic.py
```diff
@@ -70,6 +71,10 @@ def __init__(self, i, data):
             self.data['geos'] = sorted(data['geos'])
         except KeyError:
             pass
+        try:
+            self.data['subnets'] = sorted(data['subnets'])
+        except KeyError:
+            pass
```
ross, line 77: Can both `geos` and `subnets` be applied at once? Having a hard time coming up with use cases that make sense, but it seems like maybe there's really esoteric cases.

viranch: In practice it's likely that subnet rules are more specific hence appear above geo rules per best practice, but it's also easier to allow any combination of subnets & geos to be used in rules from an implementation perspective.

ross: Looking at the [NS1 implementation](https://github.com/octodns/octodns-ns1/pull/44/files#diff-05d7563145e3351ff130bda5f037afe0c98f7e55129508c3fd4474ab8d8053baR386-R395) it looks like the subnet filterchain comes before region & country. That would mean that a subnet match always overrides a (potential later) geo match.

My worry here is that we don't have a clean way to write a validation for this since we have no clue where the provider will geo-locate the ips in the block to (if it even is a single place.)

Do any other providers (that you know of) support subnet targeting? Historically I've tried to look at how things would be implemented across any provider(s) that could do it before settling on an schema/implementation to avoid coding ourselves into a corner. 

/cc https://github.com/octodns/octodns/pull/991#issuecomment-1491018659 which implements related geos validations

viranch: Yes, NS1's implementation has an implicit assumption that a subnet match is always more specific than a geo match. I can't think of a practical use-case where someone would want a geo match to override a subnet match. I don't think subnets can span country borders.

As for validation, I think its safe to expect that all subnet-only rules be put ahead of all other rules (subnet+geo, geo-only, catch-all).

Azure also supports [subnet targeting](https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-routing-methods#subnet-traffic-routing-method). It can be implemented to give priority to subnet matches over geo or also the other way round, but I think the former makes more sense practically.

viranch: [Route53 also seems to support it](https://aws.amazon.com/blogs/networking-and-content-delivery/introducing-ip-based-routing-for-amazon-route-53/) but I'm not sure of how it handles subnet vs geo match.

ross: > I don't think subnets can span country borders.

In a case where it was known and there was a clean separation I'm sure things would be broken up into smaller subnets. 

I'm also imagine there are cases where a provider has customers on both sides of a geographic border being served up the same subnet, but as far as routing things to the best DC for performance goes while that would be fine. If you were trying break things up for legal or compliance reasons it wouldn't, but trying to do that with geo-location of ips is probably going to be a bad idea in the first place.

> As for validation, I think its safe to expect that all subnet-only rules be put ahead of all other rules (subnet+geo, geo-only, catch-all).

Wherever this lands it should be documented and added to the validations to encourage the "best practice." Probably much to @istr satisfaction this would break the ordering of rules concept and be more in-line with his preferred specific-to-general implicit ordering. So long as this is documented clearly the change for it won't be the end of world. we'd need to decide/implement a broader change elsewhere after letting it further marinate. 




> Azure also supports [subnet targeting](https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-routing-methods#subnet-traffic-routing-method). It can be implemented to give priority to subnet matches over geo or also the other way round, but I think the former makes more sense practically.

👍 

> [Route53 also seems to support it](https://aws.amazon.com/blogs/networking-and-content-delivery/introducing-ip-based-routing-for-amazon-route-53/) but I'm not sure of how it handles subnet vs geo match.

Guess we'd have to run some tests. At some point to build out some tooling for running geo surveys, both in an automated fashion (~testing) and more interactive ipython notebook research style.

istr: @ross @viranch constellix supports ASN based filtering as well: https://constellix.com/news/asn-filters

From the current research that I did, I recall that all providers who support ASN based filtering (if I get that right: all of the providers who support dynamic) check for these first (before country and continent). Same might be true for subnets. Let's double-check.

@ross 
> Can both geos and subnets be applied at once? Having a hard time coming up with use cases that make sense, but it seems like maybe there's really esoteric cases.

I think one common use case is that you might want to block / limit / treat differently ASNs or subnets that are known to be VPN providers and thus break geo-based DRM limitations.

istr: > I think one common use case is that you might want to block / limit / treat differently ASNs or subnets that are known to be VPN providers and thus break geo-based DRM limitations.

To be clear -- I explicitly _do not welcome_ these use cases, but I think that they are common or are becoming common anyway... :slightly_smiling_face:

viranch: I have added [subnet ordering validations](https://github.com/octodns/octodns/pull/993/commits/96f09cb4e50c304f66915bf43c5235d122e65ec1) as well as [rule ordering validations](https://github.com/octodns/octodns/pull/993/commits/ae781f9b4f678152ff5141af098aa6d77388c6f4) per the above discussion.

viranch:
@ross lmk when you've had a chance to review and okay the general approach in this PR, I'll add tests afterwards and let this bake for more 👀.

ross (commented):
Makes sense to me.

octodns/provider/base.py
```diff
@@ -59,28 +59,69 @@ def _process_desired_zone(self, desired):
...
+                        # drop subnet rules in reverse order so indices don't shift during rule deletion
+                        rules = record.dynamic.rules
+                        for i in subnet_rules[::-1]:
+                            rule = rules[i].data
+                            if rule.get('geos'):
+                                del rule['subnets']
+                            else:
+                                del rules[i]
+
+                        # drop any pools rendered unused
```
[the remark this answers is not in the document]
viranch: The preceding `supports_warn_or_except` calls log the appropriate warnings already https://github.com/octodns/octodns/blob/221367b06a09eb702f19f35f37873fa9ff048a57/octodns/provider/base.py#L95-L97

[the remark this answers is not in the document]
viranch: Done https://github.com/octodns/octodns/pull/993/commits/a151ae31c947423ce1a7fe48a12a8579bf7ffa85

viranch:
Ready for final review with full test coverage in.

ross (approved):

Issue #1003, opened 2023-05-18
Invalid host name accepted during plan but fails push


tatnat:
When trying to sync a zone file with an invalid record (in this case there's a leading space in the zone name), the validation passes during plan but then Route53 rejects the change.

Can this record be skipped/ignored like others when running `lenient`?

Versions:

```
octodns==09.21
octodns-route53==0.0.5
```

---

Manager config:

```
zones:
  test.io.:
    lenient: True
    sources:
      - targetyaml
    targets:
      - route53
```

Zone file with YAML records (`test.io.yaml`):
* problematic zone name `' abcd'`

```
aaaa:
  type: NS
  value: aaaa.io.
' abcd':
  type: CNAME
  value: abcd.io.
bbbb:
  type: A
  value: 1.2.3.4
```

Logs:

```
INFO:Manager:sync: eligible_zones=[], eligible_targets=[], dry_run=False, force=False, plan_output_fh=<stdout>
INFO:Manager:sync:   zone=test.io.
INFO:Manager:sync:   sources=['targetyaml'] -> targets=['route53']
INFO:YamlProvider[targetyaml]:populate:   found 3 records, exists=False
INFO:Route53Provider[route53]:plan: desired=test.io.
INFO:Route53Provider[route53]:root NS record in existing, but not supported or not configured; ignoring it
INFO:Route53Provider[route53]:populate:   found 1 records, exists=True
WARNING:Route53Provider[route53]:root NS record supported, but no record is configured for test.io.
INFO:Route53Provider[route53]:root NS record in existing, but not supported or not configured; ignoring it
INFO:Route53Provider[route53]:plan:   Creates=3, Updates=0, Deletes=0, Existing Records=0
INFO:Plan:
********************************************************************************
* test.io.
********************************************************************************
* route53 (Route53Provider)
*   Create <CnameRecord CNAME 3600,  abcd.test.io., abcd.io.> (targetyaml)
*   Create <NsRecord NS 3600, aaaa.test.io., ['aaaa.io.']> (targetyaml)
*   Create <ARecord A 3600, bbbb.test.io., ['1.2.3.4']> (targetyaml)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing Records=0
********************************************************************************


INFO:Route53Provider[route53]:apply: making 3 changes to test.io.
INFO:Route53Provider[route53]:_apply: zone=test.io., len(changes)=3
[...]

PR #1004, opened 2023-05-19
Validations to ensure Record.name and Zone.name have no whitespace

ross:
/cc Fixes https://github.com/octodns/octodns/issues/1003 @tatnat


yzguy:
Nice, like most things I'm surprised stuff like this hasn't come up until now :P
I like the test method too, very clean.

yzguy (approved):

ross:
Bind failure is just a result of a test case looking for a name w/o quotes. I'll look at updating that test not to care.

tatnat:
Quick turnaround, thanks!

Issue #1002, opened 2023-05-17
RC version normalized


viranch:
After merging https://github.com/octodns/octodns/pull/1001, I ran `./script/release`. It created the proper git tag:
```
To github.com:octodns/octodns.git
 * [new tag]         v1.0.0.rc0 -> v1.0.0.rc0
Tagged and pushed v1.0.0.rc0
```
But then setuptools normalized the version to drop the last dot:
```
/tmp/build-env-6ta6ghhh/lib/python3.10/site-packages/setuptools/dist.py:519: InformationOnly: Normalizing '1.0.0.rc0' to '1.0.0rc0'
```
The twine commands failed due to this normalization:
```
Successfully built octodns-1.0.0rc0.tar.gz and octodns-1.0.0rc0-py3-none-any.whl
ERROR    InvalidDistribution: Cannot find file (or expand pattern): 'dist/*1.0.0.rc0.tar.gz'
```

I manually ran twine commands to get the release out but its with the normalized version: https://pypi.org/project/octodns/1.0.0rc0/. How much does it matter to have the git tag match this normalized version? More generally, we may want to handle this in the script if this is an important use-case.

Also, `script/sdist` contains a `0.*` version match https://github.com/octodns/octodns/blob/main/script/sdist#L13 that might need fixing too.

ross:
> How much does it matter to have the git tag match this normalized version? 

I think ideally they'd match. I guess if setuptools doesn't like the 0.rc0 we should probably just use what it prefers?

> More generally, we may want to handle this in the script if this is an important use-case.

Probably should fix it in the naming/process/script, whichever needs to happen.

> Also, `script/sdist` contains a `0.*` version match https://github.com/octodns/octodns/blob/main/script/sdist#L13 that might need fixing too.

👍 

Will poke around at this soonish unless you're planning to take a look.

viranch:
I'm not sure if/when I'll have time to look at this.

ross:
OK. I'll try and take a pass at it this week.

PR #1005, opened 2023-05-19
Make the in-repo __VERSION__ match what the most recent release used

ross:
Updates the in-repo version string to match what was recently released, namely fixing the RC part of the version number to be part of the last digit and not it's own `.`. 

Unfortunately there's nothing we can do to get the version to match in the already cut release candidate, but it'll be correct from now on.

More information in https://github.com/octodns/octodns/issues/1002.

/cc Fixes https://github.com/octodns/octodns/issues/1002


viranch (approved):
LGTM!

PR #1014, opened 2023-06-22
Add zone name to TooMuchChange during plan

tatnat:
# Summary
As the title indicates, this allows easier troubleshooting when running into a plan that errors on TooMuchChange with multiple workers running concurrently.


ross (approved):
Good call

Issue #988, opened 2023-03-21
Importing TinyDNS records ignores all the ones with a trailing dot


dbuenoparedes:
A basic example to show this behavior:

- source TinyDNS file (in `tinydns/test_example` directory):
```
=test1.example.com.:10.0.0.101:3600
=test2.example.com.:10.0.0.102:3600
Calias-test1.example.com:test1.example.com:3600
Calias-test2.example.com:test2.example.com:3600
```
- octodns configuration file (named `test_example.yml`):
```
---
providers:
  test_example:
    class: octodns.source.tinydns.TinyDnsFileSource
    directory: tinydns/test_example
  yaml_test_example:
    class: octodns.provider.yaml.YamlProvider
    directory: octodns/test_example

zones:

  example.com.:
    sources:
      - test_example
    targets:
      - yaml_test_example
```

Then I run the `octodns-dump` command to translate all TinyDNS entries into a Yaml format:
- command: `octodns-dump --config-file=test_example.yml --output-dir octodns/test_example example.com. test_example`
- output:
```
2023-03-21T15:40:39  [140500659830784] INFO  Manager __init__: config_file=test_example.yml (octoDNS 0.9.21)
2023-03-21T15:40:39  [140500659830784] INFO  Manager _config_executor: max_workers=1
2023-03-21T15:40:39  [140500659830784] INFO  Manager _config_include_meta: include_meta=False
2023-03-21T15:40:39  [140500659830784] INFO  Manager __init__: global_processors=[]
2023-03-21T15:40:39  [140500659830784] INFO  Manager __init__: provider=test_example (octodns.source.tinydns 0.9.21)
2023-03-21T15:40:39  [140500659830784] INFO  Manager __init__: provider=yaml_test_example (octodns.provider.yaml 0.9.21)
2023-03-21T15:40:39  [140500659830784] INFO  Manager dump: zone=example.com., output_dir=octodns/test_example, output_provider=None, lenient=False, split=False, sources=['test_example']
2023-03-21T15:40:39  [140500659830784] INFO  Manager dump: using custom YamlProvider
2023-03-21T15:40:39  [140500659830784] INFO  TinyDnsFileSource[test_example] populate:   found 2 records
2023-03-21T15:40:39  [140500659830784] INFO  YamlProvider[dump] plan: desired=example.com.
2023-03-21T15:40:39  [140500659830784] WARNING YamlProvider[dump] root NS record supported, but no record is configured for example.com.
[...]

PR #1019, opened 2023-06-29
Correctly handle trailing . in TinyDNS source fqdns

ross:
/cc Fixes https://github.com/octodns/octodns/issues/988 @dbuenoparedes


ross (commented):
/ci bump

Issue #987, opened 2023-03-20
Support relative record values, supress FQDN validation


rafaelfuchsbr:
octoDNS is currently checking if domain name values are FQDN, but, according to [RFC1035](https://datatracker.ietf.org/doc/html/rfc1035), it should support relative domain names.

> &lt;domain-name&gt;s make up a large share of the data in the master file.
The labels in the domain name are expressed as character strings and
separated by dots.  Quoting conventions allow arbitrary characters to be
stored in domain names.  Domain names that end in a dot are called
absolute, and are taken as complete.  Domain names which do not end in a
dot are called relative; the actual domain name is the concatenation of
the relative part with an origin specified in a $ORIGIN, $INCLUDE, or as
an argument to the master file loading routine.  A relative name is an
error when no origin is available.

I understand that the `lenient` option allow me to move forward with the process, but it's still showing the warning message in a way that can lead people to believe that this is an error/invalid value, which is not the case.

ross:
I think this came through while I was traveling and I managed to miss it then.

I think I'd personally land at using the FQDN being a best practice even if it's not required and from that POV the validation would be encouraging that.

At the same time I'm not a fan of throwing an error when someone doesn't w/o using lenient. The FQDN check was probably originally intended more to just validate that things are sane more than specifically a FQDN. 

Will have to think about this one a bit.

ross:
> octoDNS is currently checking if domain name values are FQDN, but, according to [RFC1035](https://datatracker.ietf.org/doc/html/rfc1035), it should support relative domain names.

Can you quote/copy-n-paste the section of this you feel like is related to relative values?

I started doing some deeper testing on https://github.com/octodns/octodns/pull/1023 and so far 100% of providers are putting `.` on the end of the relative values when they serve them out. E.g. Azure:

<img width="325" alt="Screen Shot 2023-07-18 at 4 21 21 PM" src="https://github.com/octodns/octodns/assets/12789/0f0e4e50-6b1f-4058-8f05-c0d455adc0ad">

...

<img width="839" alt="Screen Shot 2023-07-18 at 4 20 59 PM" src="https://github.com/octodns/octodns/assets/12789/c94d05d6-b609-40a2-97b1-e8d2a41fa916">


When queried:

```console
(env) coho:octodns ross$ dig +short relative.xormedia.com
www.
```

And when I hit it with curl/wget/etc:


```console
(env) coho:octodns ross$ curl  http://relative.xormedia.com
[...]

PR #1023, opened 2023-07-17
Add support for relative target/exchange/nameserver values

ross:
This adds support and tests for relative values as target (CNAME, NS, ...), exchanges (MX), etc. 

They are validations that will object by default and can be turned into warnings with `lenient`. This should be completely backwards compatible.

/cc Fixes https://github.com/octodns/octodns/issues/987 @rafaelfuchsbr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v1.0.0.rc2 - 2023-??-?? -
+## v1.0.0.rc1 - 2023-07-20 - The last one before the 1s
 
 * Record and Zone validation now ensures there's no whitespace in names
 * OwnershipProcessor managed records always add w/lenient=True, this allows
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Correctly handle FQDNs in TinyDNS config files that end with trailing .'s
 * Complete rewrite of TinyDnsBaseSource to fully implement the spec and the ipv6
   extensions
+* Allow relative target, nameserver, or exchange values
 
 ## v1.0.0.rc0 - 2023-05-16 - First of the ones
 
--- a/octodns/__init__.py
+++ b/octodns/__init__.py
@@ -1,3 +1,3 @@
 'OctoDNS: DNS as code - Tools for managing DNS across multiple providers'
 
-__VERSION__ = '1.0.0rc0'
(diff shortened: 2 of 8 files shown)


ross:
Main question here is whether we should encourage fqdn with a validation warning/error or just allow them outright. The PR currently landed on encourage FQDNs as a best practice. 

After testing with some providers though so far they're all appending `.`'s onto the end of the relative names when queried via DNS:

```console
(env) coho:octodns ross$ dig +short relative.xormedia.com
www.
```

Things do come back from the API without the `.`, but looks like at least some of the providers append the `.` themselves and always end up with changes:

```
* exxampled.com.
* ns1 (Ns1Provider)
*   Update
*     <CnameRecord CNAME 3600, relative.exxampled.com., www.> ->
*     <CnameRecord CNAME 3600, relative.exxampled.com., www> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing Records=3
```

Anyway, going to back off on trying to get this into `1.0.0rc1` and cut that release w/o this change.

Not [super clear what the use cases for this is](https://github.com/octodns/octodns/issues/987#issuecomment-1641122007) and it looks like it would require changes to providers to support as well.

istr:
@ross I am a little bit afraid that we would need to update lots of (or all) of the providers to support this.
You already said the same back in July:

> Not [super clear what the use case for this is](https://github.com/octodns/octodns/issues/987#issuecomment-1641122007) and it looks like it would require changes to providers to support as well.

I think that the primary use case for relative values is a shortcut/simplification for complicated master files.

When serving or submitting to providers through their API, the merging needs to be done  no matter what.

In our case, we would either need to implement the merging in the core. In this case, the few providers like `octodns-bind` with native support for relative entries will have to revert the merge. The other option would be to implement this in the vast majority of providers (since most of them do not support relative names through their API or even interpret relative values as absolute values, making things complicated/worse).

tl;dr - I would refrain from implementing this, as it could have nasty/breaking implications for many providers.

istr (commented):
_If_ we implement this, I would change the check for "is relative" to be `endswith('.')` throughout the code. Maybe even tuck it away in a single function to be sure that there is a consistent check.

ross:
Yeah. My plan for now at least, is to just let this sit here to see if the original discussion went anywhere or I was misunderstanding something. The assumption was that it'd likely just close out once the stale action decided it was time.

PR #1024, opened 2023-07-18
Version 1.0.0rc1 version bump and changelog update

ross:
## v1.0.0.rc1 - 2023-07-20 - The last one before the 1s

* Record and Zone validation now ensures there's no whitespace in names
* OwnershipProcessor managed records always add w/lenient=True, this allows
  ownership to be marked in the same zone for delegation NS records.
* octodns-report access --lenient flag to allow running reports with records
  sourced from providers with non-compliant record data.
* Correctly handle FQDNs in TinyDNS config files that end with trailing .'s
* Complete rewrite of TinyDnsBaseSource to fully implement the spec and the ipv6
  extensions
* Allow relative target, nameserver, or exchange values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v1.0.0.rc2 - 2023-??-?? -
+## v1.0.0.rc1 - 2023-07-20 - The last one before the 1s
 
 * Record and Zone validation now ensures there's no whitespace in names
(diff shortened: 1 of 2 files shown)


ross:
Fsck! Looks like I forgot to change the target branch on this. Filing a new one now.

ross:
Fixed version https://github.com/octodns/octodns/pull/1028

PR #1026, opened 2023-07-21
Dynamic zone configs

ross:
This PR adds the ability for octoDNS to dynamically build a list of zones and their config to be pushed to targets. 

It should simplify cases where there's a large number of zones being managed with the same sources and targets by adding a single wildcard zone with the prototype config and then dynamically sourcing the names of zones. The zone config would look something like:

```yaml
'*':
  sources:
  - config
  targets:
  - ns1
  - route53
```

The manager sees this wildcard as it's getting started with the sync process and then queries the configured source(s) to get a list of zones by calling `source.list_zones`. This is a new provider method and as of right no only `YamlProvider` would support it, but it's something that can be added over time to all of the providers where it makes sense. If a zone is already explicitly configured a log message is printed and it's ignored. All other zones returned by `list_zones` are added to the pool of zones to be configured. Everything else works normally in terms of `eligible_*`.

Multiple wildcard entries can be included in the config, they just need to start with a `*`, so things like the following would be valid:

```yaml
'*.dual':
  sources:
  - dual-config
  targets:
  - ns1
  - route53

`*.single'
  sources:
  - single-config
  targets:
  - ns1
  - route53
```

Note that there are 2 config providers above that would presumably have different files in their directories, otherwise the first would define all the zones and the second would do nothing b/c they all already exist. It'd also be possible to source everything from two non-config providers with different zones and combine them, e.g.:

```yaml
'*.azure':
  sources:
  - azure
  targets:
  - ns1

`*.route53'
  sources:
  - route53
  targets:
  - ns1
```

* [x] Sketch out a POC of the approach
* [x] Write tests for the basic sync functionality
* [x] Investigating adding support for the concept to dump (and maybe other commands)
* [x] Document the new functionality


HLFH:
Hello,

I wanted to migrate from DNSControl to OctoDNS because they don't fully support the OVH provider.
https://github.com/StackExchange/dnscontrol/pull/2485
https://github.com/StackExchange/dnscontrol/issues/2488
https://github.com/StackExchange/dnscontrol/issues/1640
https://github.com/StackExchange/dnscontrol/issues/1235

But I would need some kind of similar feature for `dnscontrol get-zones` to retrieve all domain zones and resource records from an account with the OVH provider.

ross:
> But I would need some kind of similar feature for `dnscontrol get-zones` to retrieve all domain zones and resource records from an account with the OVH provider.

Should be possible to implement it. PRs welcome as no clue when I'll get to it. This should be roughly what the functionality will end up looking like and I'd imagine nothing meaningful will change about the `list_zones` call that the providers will need to implement.

ross:
This is generally ready. Just going to hold on marking it so and merging until post 1.0

ross:
One of the most powerful improvements here is with `octodns-dump` where we can now "mirror" a providers config in `YamlProvider` which should provide for a very smooth octoDNS transition. You can create a basic config file with the provider details, a single dynamic zone config, and the use octoDNS dump to create zone files for all of the zones the provider has configured. Immediately after that dump a sync could be run to push the complete list of dumped zones (still dynamically configured) to one or more providers.

ross:
https://github.com/octodns/octodns-ns1/pull/59 is a quick pass at adding support to `Ns1Provider` where I have useful/interesting data for testing with, that testing lead to the changes in https://github.com/octodns/octodns/pull/1026/commits/20ace6008ccc58f60ef9f5971d2d9fa705f4ca71

viranch (approved):
This is awesome, I love it! Our config.yaml is already getting out of hands with 2k+ zones and this will be a very welcome feature.

Issue #1027, opened 2023-07-26
Support file name and line number in ValidationError message


joschi36:
Type: Feature Request/Idea

I am currently building an OctoDNS GitHub Action with Python and wanted to use the [annotation feature](https://github.com/orgs/community/discussions/25002#discussioncomment-3246152) of GitHub. Unfortunately, the ValidationError error message only includes the zone and the error.

Current error message:
```
octodns.record.ValidationError: Invalid record test.com.
  - invalid IPv4 address "abc"
```

idea for future error message:
```
zones/test.com.yaml:L10:
octodns.record.ValidationError: Invalid record test.com.
  - invalid IPv4 address "abc"
```

Maybe this could also be implemented as a separate tool, like a linter perhaps? This would also open a whole range of new possibilities, like for instance give some custom linting rules (e.g. TTL not allowed lower/higer than X)

ross:
Like the ideal. Figuring out how to go about it would probably require a fair bit of thinking/planning. The biggest issue would be getting the line number from the yaml parser. AFAIK it'd require something custom to get the line number of yaml elements as they are parsed. There are some results for that, e.g. https://stackoverflow.com/questions/13319067/parsing-yaml-return-with-line-number so it may be doable. 

After that it'd require figuring out a clean way to plumb the data through `Records.new` to get it to `Record.validate`, probably as an optional parameter. 

Ideally other types of sources would be supported as well, not just yaml, so there's probably some thinking/planning around the context being arbitrary so that other providers can pass in the context info that makes sense for them.

ross:
Fixed by https://github.com/octodns/octodns/pull/1029

PR #1029, opened 2023-07-29
config/validation errors with source context

ross:
This sketches out a system for including context in error messages about configuration or validation. It's implemented for YamlProvider here, but nothing would stop other providers from passing in context if there's something useful for them to provide, maybe things like account ids or provider specific/internal ids...

For now this is just a POC. It needs a bunch of cleanup and testing before it'd be ready to move forward. Definitely a post 1.0 release change. Pretty happy with how it's working though and was cleaner to get going that I expected.

### Invalid Provider Config

```console
...
  File "/Users/ross/octodns/octodns/octodns/manager.py", line 202, in _config_providers
    raise ManagerException(
octodns.manager.ManagerException: Provider ns1 is missing class, config/dev.yaml, line 26, column 5
```

```console
...
  File "/Users/ross/octodns/octodns/octodns/manager.py", line 219, in _config_providers
    raise ManagerException(
octodns.manager.ManagerException: Incorrect provider config for ns1, config/dev.yaml, line 26, column 5
```

### Record errors and validation

```console
...
  File "/Users/ross/octodns/octodns/env/lib/python3.10/site-packages/yaml/constructor.py", line 100, in construct_object
    data = constructor(self, node)
  File "/Users/ross/octodns/octodns/octodns/yaml.py", line 55, in _construct
    raise ConstructorError(
yaml.constructor.ConstructorError: keys out of order: expected ttl got type at ./config/exxampled.com.yaml, line 3, column 5
```

```console
...
    self._populate_from_file(filename, zone, lenient)
  File "/Users/ross/octodns/octodns/octodns/provider/yaml.py", line 175, in _populate_from_file
    record = Record.new(
  File "/Users/ross/octodns/octodns/octodns/record/base.py", line 63, in new
    raise Exception(msg)
Exception: Unknown record type: "AAAAA", ./config/exxampled.com.yaml, line 7, column 5
```

```console
...
  File "/Users/ross/octodns/octodns/octodns/record/base.py", line 75, in new
    raise ValidationError(fqdn, reasons, context)
octodns.record.exception.ValidationError: Invalid record "cname.exxampled.com.", ./config/exxampled.com.yaml, line 14, column 3
  - CNAME value "foo.bar" missing trailing .
```

/cc https://github.com/octodns/octodns/issues/1027 which lead to this work
/cc @joschi36 who suggested it

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -254,15 +254,14 @@ class Manager(object):
     def _config_plan_outputs(self, plan_outputs_config):
         plan_outputs = {}
         for plan_output_name, plan_output_config in plan_outputs_config.items():
-            context = getattr(plan_output_config, 'context', None)
+            context = getattr(plan_output_config, 'context', '')
             try:
                 _class = plan_output_config.pop('class')
             except KeyError:
                 self.log.exception('Invalid plan_output class')
-                msg = f'plan_output {plan_output_name} is missing class'
-                if context:
-                    msg += f', {context}'
-                raise ManagerException(msg)
+                raise ManagerException(
+                    f'plan_output {plan_output_name} is missing class, {context}'
+                )
             _class, module, version = self._get_named_class(
                 'plan_output', _class, context
             )
@@ -281,10 +280,9 @@ class Manager(object):
                     )
             except TypeError:
                 self.log.exception('Invalid plan_output config')
-                msg = f'Incorrect plan_output config for {plan_output_name}'
-                if context:
-                    msg += f', {plan_output_config.context}'
-                raise ManagerException(msg)
+                raise ManagerException(
+                    f'Incorrect plan_output config for {plan_output_name}, {context}'
+                )
 
         return plan_outputs
 
@@ -324,10 +322,9 @@ class Manager(object):
             self.log.exception(
                 '_get_{}_class: Unable to import module %s', _class
             )
-            msg = f'Unknown {_type} class: {_class}'
-            if context:
-                msg += f', {context}'
-            raise ManagerException(msg)
+            raise ManagerException(
+                f'Unknown {_type} class: {_class}, {context}'
+            )
 
         try:
             return getattr(module, class_name), module_name, version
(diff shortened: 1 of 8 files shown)


octodns/manager.py
```diff
@@ -190,9 +190,11 @@ def _config_providers(self, providers_config):
             except KeyError:
                 self.log.exception('Invalid provider class')
                 raise ManagerException(
-                    f'Provider {provider_name} is missing class'
+                    f'Provider {provider_name} is missing class, {provider_config.context}'
```
ross, line 193: All of the yaml sourced config in manager will only ever be loaded by our custom yaml loaders, `SortEnforcingLoader` or `ContextLoader`, so it's safe to assume they'll be `ContextDict`s.

octodns/record/base.py
```diff
@@ -46,24 +46,33 @@ def new(cls, zone, name, data, source=None, lenient=False):
             reasons.append('invalid record, whitespace is not allowed')
 
         fqdn = f'{name}.{zone.name}' if name else zone.name
+        context = getattr(data, 'context', None)
```
ross, line 49: In contrast `Record` stuff may get data from our yaml infra or it may be coming from other providers that do not use `ContextDict` so we always need check to see if the attr exists gracefully, and then only use the context if there was a value (below.)

joschi36:
@ross Very nice. Unfortunately, I'm unavailable for testing it in the next three weeks. But definitely very cool you've responded so quickly with a PR. I will test it and try to integrate it into GitHub Actions.
One question: As far as I understand, this would only output the first found failure and not list all like a linter, correct? If yes can we maybe implement this functionality into octodns-validate? WDYT?

ross:
> As far as I understand, this would only output the first found failure and not list all like a linter, correct? 

That is correct.

> If yes can we maybe implement this functionality into octodns-validate? WDYT?

Would have to think about it/look into how it might work. All the errors are throw as exceptions and there's not really way to continue and stack them up.

ross (commented):
/ci modules bump

PR #1030, opened 2023-07-29
Require python 3.8, remove a p3.7 import work-around, address other minor TODOs as prep for 1.0

ross:


ross (commented):
/ci

ross:
Failure seems unrelated. netbox has been failing for a while now. Going to remove it in a separate PR.

PR #1031, opened 2023-07-30
Comment out netbox modules ci run

ross:


ross:
Started failing in the middle of https://github.com/octodns/octodns/pull/1020 and has consistently failed since. there are no references to tinydns in netbox, https://github.com/search?q=repo%3Asukiyaki%2Foctodns-netbox%20tinydns&type=code, and I don't really see anything else that seems likely to cause the problem.

ross (commented):
/ci bump modules

PR #1032, opened 2023-07-30
Deprecate geo records in favor of dynamic

ross:
`dynamic` has been around a long long time now and is far superior to the old geo setup. Formally deprecating them as we move into 1.0 with the plan to remove the functionality when we hit 2.x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v1.0.0 - 2023-??-?? - The One
 
+#### Noteworthy changes
+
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci bump modules

ross:
Heh, there's a test in ns1 that's failing b/c it's also looking at warnings and the new one gets in its way.

ross (commented):
/ci bump modules again, now that NS1 should be happy

/cc https://github.com/octodns/octodns-ns1/pull/60

PR #1033, opened 2023-07-30
v1.0.0 version bump and changelog update

ross:
## v1.0.0 - 2023-07-30 - The One

1.0 marks a point at which we can formally deprecate things that will be
going away with 2.0 more than specific functionality that has been added or
having reached a notable level of stability (beyond what is normal.) It is also
long (years) overdue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## v1.0.0 - 2023-??-?? - The One
+## v1.0.0 - 2023-07-30 - The One
+
+1.0 marks a point at which we can formally deprecate things that will be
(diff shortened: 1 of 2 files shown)


yzguy (approved):

yzguy:
1.0! We did it!

ross:
Heh, yeah. I intend to a bit better job of [semantic versioning](https://semver.org/) going forward, looking at the progression https://pypi.org/project/octodns/#history is somewhat comical. That mostly means bumping the minor version on most releases and only bumping the patch version when things are a true fix/patch release. :grin:

Issue #1038, opened 2023-08-08
Updating SRV records fails with error


hellbot:
When updating SRV record having octodns_bind.Rfc2136Provider as target an "AttributeError: 'SrvValue' object has no attribute 'rdata_text'" appears.

It happens due to heavy reliance on rrs property witch calls rdata_text which doesn't exists in SRVValue in main octodns. 

I think best way to fix is to add that method to octodns 

```diff
diff --git a/octodns/record/srv.py b/octodns/record/srv.py
index af3797c..a256b94 100644
--- a/octodns/record/srv.py
+++ b/octodns/record/srv.py
@@ -132,6 +132,10 @@ class SrvValue(EqualityTupleMixin, dict):
     def data(self):
         return self
 
+    @property
+    def rdata_text(self):
+        return f"{self.priority} {self.weight} {self.port} {self.target}"
+
     def __hash__(self):
         return hash(self.__repr__())

```

PR #1039, opened 2023-08-09
rdata_text for SrvValue

hellbot:
Fixing #1038


ross (commented):
Looks good. `./script/format` should get CI happy. If you run `./script/bootstrap` it should install the pre-commit hooks that check it as well as run lint and ci.

ross (approved):

PR #1045, opened 2023-08-17
Archive code coverage data

ross:


ross:
OK this works. Tries to create the articfact regardless of whether previous steps succeeded or failed so any time things fail tests etc it'll also fail this since no coverage will have been created, but that seems fine.

Screen Shot 2023-08-17 at 12 43 46 PM

ross:
Failures are b/c this was built on top of https://github.com/octodns/octodns/pull/1044 which led to the need for it.

There's more we could do with this, e.g. upload the dists built in setup-py etc, but this solves the immediate need and we can add those things if/when they're required.

ross:
Reworked this off of main.

ross:
Replaced by https://github.com/octodns/octodns/pull/1046

Issue #962, opened 2022-12-03
Best way to share multiple records between zones


felixoi:
I'd like to share a couple of records between multiple zones. For example I'm using the same autodiscover/autoconfig records for multiple domains:
```
_autodiscover._tcp:
  type: SRV
  value:
    priority: 10
    weight: 10
    port: 443
    target: autodiscover.mailcow.de.

_autoconfig._tcp:
  type: SRV
  value:
    priority: 10
    weight: 10
    port: 443
    target: autoconfig.mailcow.de.

autodiscover:
  octodns:
    cloudflare:
      proxied: false
  type: CNAME
  value: autodiscover.mailcow.de.

autoconfig:
  octodns:
    cloudflare:
      proxied: false
  type: CNAME
  value: autoconfig.mailcow.de.
```
As the other records differ I cannot use the "alias" functionality for zones. What's the best way to achieve shared records in octodns? I don't like to copy & paste in "as code" approaches. An option would maybe be an separate yaml templating preprocessor, as I would guess that I'm not the only one sharing records between zones, it does not seem to be the perfect approach to me tho. A YAML provider which just returns the same records for every zone would maybe work? Is there something prebuild like this?

ross:
Not what you're asking for, but [Yaml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) are useful for avoiding repeating the same bits of yaml within a single zone, e.g.:

```yaml
_autodiscover._tcp: &auto
  type: SRV
  value: 
    priority: 10
    weight: 10
    port: 443
    target: autodiscover.mailcow.de.

_autoconfig._tcp: *auto
```

As for records shared across zones YAML itself doesn't support any sort of `!include` directive so there's no inbuilt way to do what you're after.  I've wanted it in the past as well and have kicked around ideas for ways to support it, but haven't come up with anything viable. 

I just had a new thought about something that might make it possible that I'll play around with in the next few days to see if it's workable. Will report back what I find.

felixoi:
Yeah I'm aware of anchors but as you mentioned they do not fit my usecase. I look forward to your feedback on whether it works.

ross:
So my first thought wasn't viable. I was looking into adding a parameter to `YamlProvider` that allowed you to define a "shared" bit of yaml that was prepended onto the zone file before parsing. That runs into problems with files starting with `---` and beyond that sorting/ordering issues. 

Still have one more thing in mind...

ross:
[...]

PR #1048, opened 2023-08-18
YamlProvider support for shared file, loaded into all zones

ross:
Add support for a "shared" yaml file in YamlProvider that is loaded for all zones using the provider. There's three main use cases I can imagine:

### Common records across all zones

This would be useful if there's a shared record(s) that should be present in all/many zones as was described in https://github.com/octodns/octodns/issues/962. 

In this case the provider config would likely look something like:

```yaml
  config:
        directory: ./config
        shared_filename: shared.yaml
```

Records for all zones would then be placed in `./config/shared.yaml` and individual zone files with the records specific to the zones would live in `./config/something.com.yaml`.  This route may even be preferable when there are no specific zones since it would allow dynamic zone config.

### Many identical zones

All of the zones configured have exactly the same records.

```yaml
  config:
        directory: ./config
        shared_filename: shared.yaml
        disable_zonefile: true
```

Only `./config/shared.yaml` would exist and all of the zone data would come from it. The list of zones would be explicitly configured in the main octoDNS config file.

### Combination

There's some shared records across some, but not all zones.

```yaml
  config:
        directory: ./config
  common:
        directory: ./config
        shared_filename: shared.yaml
        disable_zonefile: true
```

Zones with only the shared records can list `common` as their sole source. Zones with only specific records can use `config`. Finallyzones with both can list `common` and `config` as their sources. Overriding behaviors could even by achieved with the careful use of `populate_should_replace` and ordering of the sources.

P.S. the fact that this is not a 2-5 line PR shows why the refactoring was best done first :grin:

/cc https://github.com/octodns/octodns/pull/1044 which preceeds this work/change
/cc Fixes https://github.com/octodns/octodns/issues/962 @felixoi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   possible. This includes the ability to split some zones and not others and
   even to have partially split zones with some records in the primary zone YAML
   and others in a split directory. See YamlProvider documentation for more info.
+* YamlProvider now supports a `shared_filename` that can be used to add a set of
+  common records across all zones using the provider. It can be used stand-alone
+  or in combination with zone files and/or split configs to aid in DRYing up DNS
+  configs.
 
 #### Stuff
 
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -57,6 +57,11 @@ class YamlProvider(BaseProvider):
         # (optional, default False)
         split_catchall: false
 
+        # Optional filename with record data to be included in all zones
+        # populated by this provider. Has no effect when used as a target.
+        # (optional, default null)
+        shared_filename: null
+
         # Disable loading of the zone .yaml files.
         # (optional, default False)
         disable_zonefile: false
@@ -171,6 +176,7 @@ class YamlProvider(BaseProvider):
         supports_root_ns=True,
         split_extension=False,
         split_catchall=False,
+        shared_filename=False,
         disable_zonefile=False,
         *args,
         **kwargs,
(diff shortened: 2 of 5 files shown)


felixoi:
This currently doesn't support merging entries as mentioned in https://github.com/octodns/octodns/issues/962#issuecomment-1354585570 or?

ross:
> This currently doesn't support merging entries as mentioned in [#962 (comment)](https://github.com/octodns/octodns/issues/962#issuecomment-1354585570), right?

No, it doesn't. I'm not really sure how that'd be implemented in a general enough fashion. Namely how people would indicate the behavior they want? Merging would make sense for TXT records, but not for other types so it seems like it'd need fine grained control. 

Even then merging wouldn't generally be safe, e.g. if there was an SPF value in both TXT records.  Keeping both of them wouldn't make sense, it'd really need to know what the TXT values were about which is definitely beyond the scope of octoDNS.  It's something that could be implemented in a custom provider if there were specific rules/logic to be applied. 

You might take a look at: https://github.com/octodns/octodns/pull/1050, it's not going to do merging, but it does offer a way to define values once and combine them into records.

ross (commented):
/ci modules bump

stavros-k:
I suppose to use this we have to wait for the next octodns release in PyPi?
Thanks!

ross:
> I suppose to use this we have to wait for the next octodns release in PyPi?

See https://github.com/octodns/octodns#installing-a-specific-commit-sha if you're like to pick something up prior to a release.

stavros-k:
> > I suppose to use this we have to wait for the next octodns release in PyPi?
> 
> See https://github.com/octodns/octodns#installing-a-specific-commit-sha if you're like to pick something up prior to a release.

Of course RTFM :D
Thanks, been too long I set the automation up that I forgot haha!

PR #1049, opened 2023-08-19
Record's carry context, Zone exceptions make use of it to help with error messages

ross:
As a follow on to https://github.com/octodns/octodns/pull/1029 this PR extends the error messages w/more context to Zone based exceptions, e.g.:

```console
octodns.zone.DuplicateRecordException: Duplicate record relative.exxampled.com., type CNAME
  existing: ./config/exxampled.com.yaml, line 30, column 3
  new:      ./config/exxampled.com.yaml, line 30, column 3
...
octodns.zone.SubzoneRecordException: Record foo.both.exxampled.com. is under a managed subzone, ./config/exxampled.com.yaml, line 30, column 3
...
octodns.zone.InvalidNodeException: Invalid state, CNAME at relative.exxampled.com. cannot coexist with other records, ./config/exxampled.com.yaml, line 32, column 5
```

It also extends the general testing of the context bits into a few places that were missing it.

### DIFF
--- a/octodns/record/base.py
+++ b/octodns/record/base.py
@@ -5,6 +5,7 @@
 from collections import defaultdict
 from logging import getLogger
 
+from ..context import ContextDict
 from ..equality import EqualityTupleMixin
 from ..idna import IdnaError, idna_decode, idna_encode
 from .change import Update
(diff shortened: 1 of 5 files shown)


ross (commented):
/cc ci modules bump

PR #1050, opened 2023-08-19
Add support for !include YAML directive

ross:
Add support for `!include` YAML directive. This squarely lands in power user territory, but it's something that would have been super useful had it been available for GitHub's setup. Namely we have a number of dynamic records with same group of pools, edge locations, but different IP addresses/values based on the service. We'd always have to duplicate the exact same set of rules to direct to those pools across all the dynamic records. It would have be nice to be able to write those rules once and share that logic across all the records. 

A simplified version of this would look something like:

```yaml
---
api:
  dynamic:
    pools:
      iad: 
        ...
      fra:
        ...
      vpc-ap-se-1:   
    rules: !include dynamic-rules.yaml
  type: A
  value: 1.2.3.4
```

```yaml
---
- geos: [AP]
  pool: vpc-ap-se-1
- geos: [EU]
  pool: fra
- pool: iad
```

This can be used in other cases as well, e.g. if there's a TXT value that needs to be stuffed into a number of records:

```yaml
---
? ''
:
  - type: TXT
    values:
      - !include domain-claim.txt
      - !include spf.txt
```

```yaml
---
v=spf1 ip4:1.3.2.0/22 include:_spf.google.com ~all
```

```yaml
---
i-own-this=88048f84-620e-44f0-8377-0af47755d1ec
```

This allows users to define values/data a single time and then include it into multiple records. It will NOT allow including records themselves, for that see the YamlProvider's `static_filename` functionality.  The two can be mixed though to get some super flexible/powerful functionality. Though there's also the possibility of some 🦶 🔫ing if people get out of hand with it :grin:

/cc https://github.com/octodns/octodns/pull/1044 and https://github.com/octodns/octodns/pull/1048 which proceed this work/change


ross (commented):
/ci modules bump

PR #1062, opened 2023-09-13
v1.1.0 version bump & changelog update

ross:
#### Noteworthy changes

* New dynamic zone config support that allows wildcard entries in the octoDNS
  config to be expanded by the source provider(s). See
  [Dynamic Zone Config](/README.md#dynamic-zone-config) for more information.
* SplitYamlProvider has been deprecated and will be removed in 2.0. YamlProvider
  now includes the ability to process split zones when configured to do so and
  allows for more flexibility in how things are laid out than was previously
  possible. This includes the ability to split some zones and not others and
  even to have partially split zones with some records in the primary zone YAML
  and others in a split directory. See YamlProvider documentation for more info.
* YamlProvider now supports a `shared_filename` that can be used to add a set of
  common records across all zones using the provider. It can be used stand-alone
  or in combination with zone files and/or split configs to aid in DRYing up DNS
* YamlProvider now supports an `!include` directive which enables shared
  snippets of config to be reused across many records, e.g. common dynamic rules
  across a set of services with service-specific pool values or a unified SFP
  value included in TXT records at the root of all zones.
* SpfRecord is formally deprecated with an warning and will become a
  ValidationError in 2.x
* SpfDnsLookupProcessor is formally deprcated in favor of the version relocated
  into https://github.com/octodns/octodns-spf and will be removed in 2.x
* MetaProcessor added to enable some useful/cool options for debugging/tracking
  DNS changes. Specifically timestamps/uuid so you can track whether changes
  that have been pushed to providers have propogated/transferred correctly.

#### Stuff

* Add context to general configuration and Record validation, e.g.
  Some problem at filename.yaml, line 42, column 14. Our custom Yaml Loaders
  attach this context information, arbitrary string. Other providers may do so
  by creating ContextDict to pass as `data` into Record.new.
* Add --all option to octodns-validate to enable showing all record validation
  errors (as warnings) rather than exiting on the first. Exit code is non-zero
  when there are any validation errors.
* New `post_processors` manager configuration parameter to add global processors
  that run AFTER zone-specific processors. This should allow more complete
  control over when processors are run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v1.1.0 - 2023-??-?? - ???
+## v1.1.0 - 2023-09-13 - More than enough for a minor release
 
 #### Noteworthy changes
 
--- a/octodns/__init__.py
+++ b/octodns/__init__.py
@@ -1,3 +1,3 @@
 'OctoDNS: DNS as code - Tools for managing DNS across multiple providers'
 
-__VERSION__ = '1.0.0'
+__VERSION__ = '1.1.0'


ross:
This is probably the most feature-ful single release of octoDNS core since the project started. There's a bunch of things in here that I wish I'd had at GitHub and more generally had been there all along. 

None of the changes were particularly complicated code-wise, but all of them took some thinking to figure out how to go about cleanly/sanely or really more to the point to figure out how carve up the problem space into solvable pieces.

I have plans for detailed examples/descriptions for all of this in the new examples directories https://github.com/octodns/octodns/pull/1053, but for now here's some ...

## Highlights

### Dynamic Zone Config

A single (or few) wildcard Zone configs that get applied to all the zones the sources list know about

```yaml
zones:
  '*':
    sources:
      - config
    targets:
      - ns1
```

With the above config the `YamlProvider` config will be asked for a list of it's zones, which it gets by looking at all the files on disk, and dynamic entries for all those zones are created with `sources` and `targets` matching the wildcard.

### Partially Split Yaml

`SplitYamlProvider` was deprecated and the functionality is subsumed by `YamlProvider` with more flexibility over what data lives in what files. 

This is super powerful in that you can now create files with record data arbitrarily. The best Example I can think of for this would be splitting things up by owning team so that you could require [CODEOWNERS review](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) for changes. That would let each team own it's own records and anyone else would need to get approval to touch them.

More generally it will allow breaking things up logically. E.g. the big complicated dynamic records can be in their own dedicated files and then a "catchall" file can be used for all the minutia. 

### `YamlProvider` `shared_filename`

Have some boilerplate records that you'd like to include in lots of or even all zones. `shared_filename` will allow you to do just that. You can create a file with the common records defined in it and then pass the path to that file (relative to the provider's config directory) in `shared_filename`. Every zone loaded by the provider will include the common records.

The simplest use case is a single file with common records included in everything, but that's just the beginning. You could also create `YamlProvider`s with only a `shared_filename` and then add them to only the zones you want them in by including it in their `sources` list. 

### `YamlProvider` `!include`

Finally there's the new `!include <filename.yaml>` directive. This allows you to include common values across multiple records/zones. 

This would have been super helpful for our `dynamic` record `rules`. The pool details for our dynamic records were different across each record, but in general we always had the same `rules`. This functionality would have allowed us to define the rules once in a single place and then include them in all of the `dynamic` records.

#### config/my-domain.com.yaml

```yaml
...
www:
  dynamic:
    pools:
      ...
    rules: !include dynamic_rules.yaml
  ...
```

#### config/dynamic_rules.yaml
```
---
- geos: 
    - AP
    - OC
  pool: ap
- geos: 
    - AF
    - EU
  pool: eu
- pool: iad
```

yzguy (approved):

viranch (approved):
Looking forward to using some of the new features!

PR #1063, opened 2023-09-14
Fix typo when loading auto-arpa config

kkzo:
Resolved a typo that prevented the auto-arpa configuration options from loading.

### DIFF
--- a/tests/config/simple-arpa.yaml
+++ b/tests/config/simple-arpa.yaml
@@ -1,6 +1,7 @@
 manager:
   max_workers: 2
   auto_arpa:
+    populate_should_replace: True
     ttl: 1800
 
 providers:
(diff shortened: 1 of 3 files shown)


ross (approved):
Doh. Should have checked this before I moved forward with 1.1.0 in https://github.com/octodns/octodns/pull/1062. Will queue up a patch release once this is merged.

tests/config/simple-arpa.yaml
```diff
@@ -1,6 +1,7 @@
 manager:
   max_workers: 2
   auto_arpa:
+    populate_should_replace: True
```
ross, line 4: Any specific reasons for adding this and checking it in the tests. It seems fine just wondering what lead to the change as it doesn't seem to be related to coverage. Just testing that the param was passed through correctly?

kkzo: Thanks for the quick release. Yes, I added that to test if the param is actually making its way through correctly.

PR #1065, opened 2023-09-18
The record got field names from the DNSKEY record by accident, fix this

baest:
Basically changing from
https://www.rfc-editor.org/rfc/rfc4034.html#section-2.1 to https://www.rfc-editor.org/rfc/rfc4034.html#section-5.1

So:
flags -> key_tag
protocol -> algorithm
algorithm -> digest_type
public_key -> digest

### DIFF
--- a/octodns/record/ds.py
+++ b/octodns/record/ds.py
@@ -8,31 +8,31 @@ from .rr import RrParseError
 
 
 class DsValue(EqualityTupleMixin, dict):
-    # https://www.rfc-editor.org/rfc/rfc4034.html#section-2.1
+    # https://www.rfc-editor.org/rfc/rfc4034.html#section-5.1
 
     @classmethod
     def parse_rdata_text(cls, value):
         try:
-            flags, protocol, algorithm, public_key = value.split(' ')
+            key_tag, algorithm, digest_type, digest = value.split(' ')
         except ValueError:
             raise RrParseError()
         try:
-            flags = int(flags)
-        except ValueError:
-            pass
-        try:
-            protocol = int(protocol)
+            key_tag = int(key_tag)
         except ValueError:
             pass
         try:
             algorithm = int(algorithm)
         except ValueError:
             pass
+        try:
+            digest_type = int(digest_type)
+        except ValueError:
+            pass
         return {
-            'flags': flags,
-            'protocol': protocol,
+            'key_tag': key_tag,
             'algorithm': algorithm,
-            'public_key': public_key,
+            'digest_type': digest_type,
+            'digest': digest,
         }
 
     @classmethod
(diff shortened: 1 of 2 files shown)


octodns/record/ds.py
```diff
@@ -42,25 +42,25 @@ def validate(cls, data, _type):
         reasons = []
         for value in data:
             try:
-                int(value['flags'])
+                int(value['key_tag'])
```
ross, line 45: Unfortunately we'll need to fall back to looking for the previous names for backwards compatibility in case anyone has defined existing records with the original names. 

Cleanest way I can think of to update this section is something like:

```suggestion
                # key_tag is correct, but we previously released this with
                # `flags` so we need to check box
                int(value.get('flags', None) or value['key_tag'])
```

We'll need to do the same thing for all of them.

octodns/record/ds.py
```diff
@@ -70,28 +70,20 @@ def process(cls, values):
     def __init__(self, value):
         super().__init__(
             {
-                'flags': int(value['flags']),
-                'protocol': int(value['protocol']),
+                'key_tag': int(value['key_tag']),
```
ross, line 73: These will need something like:

```suggestion
                'key_tag': int(value.get('flags', None) or value['key_tag']),
```

tests/test_octodns_record_ds.py
```diff
@@ -12,64 +12,64 @@
...
             (
                 {
-                    'flags': 0,
-                    'protocol': 1,
-                    'algorithm': 2,
-                    'public_key': 'abcdef0123456',
+                    'key_tag': 0,
+                    'algorithm': 1,
+                    'digest_type': 2,
+                    'digest': 'abcdef0123456',
```
ross, line 21: We'll want to leave the original key names in at least one of these.

ross (commented):
Thanks for the find &amp; even more for the PR with the fixes. Comments inline. Once they're sorted we can move this forward.

octodns/record/ds.py
baest: Problem is that we have the algorithm field in both sets of fields (and for different fields) so I have done it a bit differently.

tests/test_octodns_record_ds.py
baest: Sure and to reach 100% test coverage I needed a bunch of them :)

octodns/record/ds.py
baest: I did the same here, it is a bit cumbersome, but I personally prefer it since it's an either or situation where I don't we want anyone mixing the fields. Also again with algorithm being present both field sets makes it difficult to handle.

ross (approved):
Added a DEPRECATION warning when the old fields are used so that we can eventually remove the backwards compat. I think this is gtg.

Thanks!

ross:
Ah. Once the integration tests run it caught an issue where the old property names are being accessed. That should be easy enough to fix.

ross:
So I was going to add back the "legacy" properties, but realized that `algorithm` is in both the old and new and that wouldn't work. Good news is that octodns-powerdns has never been released with DS support so we at least have options. I think we'll have to go about this a bit differently and reorder things a bit.  

1. Merge this change
2. Merge https://github.com/octodns/octodns-powerdns/pull/50 to update octodns-powerdns to use the new values
3. Once a release of octoDNS core has been made with DS fixes update octodns-powerdns to require >= that new release
4. Cut a release of octodns-powerdns

ross:
Scratch the last comment about process. After looking at the version checking patch I realized that the cleanest path forward would be to have octodns-powerdns just handle both ways transparently. Since there hasn't been a release with DS support yet that should be enough, https://github.com/octodns/octodns-powerdns/pull/51 does this.

Issue #1066, opened 2023-09-18
New YamlProvider crashes in hybrid mode


viranch:
My directory of yamls has a mix of split and non-split zones. When using the new YamlProvider from `octodns==1.1.1` with `split_extensions: .` enabled, it still tries to read the non-split zone file even for zones that are split, and then crashes because that non-split file doesn't exist. Vice versa for a flat zone file, which crashes when trying to list the split directory for that non-split zone.

Error snippets:

Split zone:
```
  File "/home/viranch/code/dns/env/lib/python3.11/site-packages/octodns/provider/yaml.py", line 301, in _populate_from_file
    with open(filename, 'r') as fh:
         ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './src/its-split.com.yaml'
```

Non-split zone:
```
  File "/home/viranch/code/dns/env/lib/python3.11/site-packages/octodns/provider/yaml.py", line 284, in _split_sources
    for filename in listdir(directory):
                    ^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './src/not-split.com.'
```

Fwiw I'm specifying the zone name(s) in the `octodns-sync` command, but looks like this would happen either way.

PR #1067, opened 2023-09-18
Include sources only if they exist

viranch:
Addresses #1066

### DIFF
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -278,8 +278,10 @@ class YamlProvider(BaseProvider):
                     f'Both UTF-8 "{utf8}" and IDNA "{idna}" exist for {zone.decoded_name}'
                 )
             directory = utf8
-        else:
+        elif isdir(idna):
             directory = idna
+        else:
+            return []
 
         for filename in listdir(directory):
             if filename.endswith('.yaml'):
@@ -294,8 +296,10 @@ class YamlProvider(BaseProvider):
                     f'Both UTF-8 "{utf8}" and IDNA "{idna}" exist for {zone.decoded_name}'
                 )
             return utf8
+        elif isfile(idna):
+            return idna
 
-        return idna
+        return None
 
     def _populate_from_file(self, filename, zone, lenient):
         with open(filename, 'r') as fh:
@@ -341,10 +345,17 @@ class YamlProvider(BaseProvider):
             sources.extend(self._split_sources(zone))
 
         if not self.disable_zonefile:
-            sources.append(self._zone_sources(zone))
+            source = self._zone_sources(zone)
+            if source:
+                sources.append(self._zone_sources(zone))
 
         if self.shared_filename:
             sources.append(join(self.directory, self.shared_filename))
+
+        if not sources:
+            self.log.info(
+                'populate:   no YAMLs found for %s', zone.decoded_name
+            )
 
         # determinstically order our sources
         sources.sort()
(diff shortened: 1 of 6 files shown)


ross (commented):
Not at a point where I can sit down to run this, but main question here is whether this would change the previous behavior, i.e. throwing an error when a configured zone file doesn't exist.

viranch:
The previous behavior would eventually end up throwing `FileNotFoundError` at some point when a zone file/dir doesn't exist since that case is not explicitly handled.

ross:
> The previous behavior would eventually end up throwing `FileNotFoundError` at some point when a zone file/dir doesn't exist since that case is not explicitly handled.

Was guessing that since this is checking for existance that would change and instead of the user seeing an error they'd end up with an empty zone.

viranch:
True, post-PR we'd be logging and continuing with an empty zone as opposed to exiting. I guess exiting with an error is the "right" behavior: if the zone already exists it'd be pretty dangerous to try and empty it even though octodns-sync would show sufficient deletes in the plan output.

octodns/provider/yaml.py
```diff
@@ -353,9 +353,7 @@ def populate(self, zone, target=False, lenient=False):
             sources.append(join(self.directory, self.shared_filename))
 
         if not sources:
-            self.log.info(
-                'populate:   no YAMLs found for %s', zone.decoded_name
-            )
+            raise ProviderException(f'no YAMLs found for {zone.decoded_name}')
```
ross, line 356: :cool:, That was the suggestion I came up with when thinking about it.

ross (approved):

viranch:
> Throw a changelog entry in with this if you don't mind. I assume after it (and any other YAML fixes) land we'll cut a point release.

Added changelog entry. Yep hoping to cut a release after the current batch of yaml fixes are in.

ross:
> Added changelog entry. Yep hoping to cut a release after the current batch of yaml fixes are in.

@viranch noticed you added a specific version & date in the CHANGELOG update. Confused me when I was looking at release versions just now (and when i made a changelog entry earlier. I though you'd already cut 1.1.2. We should probably do that soon and I guess put my RR fixes in and makes it 1.2.0?

viranch:
> > Added changelog entry. Yep hoping to cut a release after the current batch of yaml fixes are in.
> 
> @viranch noticed you added a specific version & date in the CHANGELOG update. Confused me when I was looking at release versions just now (and when i made a changelog entry earlier. I though you'd already cut 1.1.2. We should probably do that soon and I guess put my RR fixes in and makes it 1.2.0?

Yep sorry my bad. You want to cut 1.2.0 instead of 1.1.2 or make them both separate releases?

viranch:
> > Added changelog entry. Yep hoping to cut a release after the current batch of yaml fixes are in.
> 
> @viranch noticed you added a specific version & date in the CHANGELOG update. Confused me when I was looking at release versions just now (and when i made a changelog entry earlier. I though you'd already cut 1.1.2. We should probably do that soon and I guess put my RR fixes in and makes it 1.2.0?

Yep sorry my bad. You want to cut 1.1.2 then 1.2.0 or straight to 1.2.0?

ross:
> Yep sorry my bad. You want to cut 1.1.2 then 1.2.0 or straight to 1.2.0?

I'd say straight to 1.2.0. Seems fine to mix patch fixes with the RR improvements since they shouldn't hurt anything.

PR #1071, opened 2023-09-25
Record RR handling improvements

ross:
Support passing source into `Record.from_rrs` and handle quoted strings in all of the `parse_rdata_text` methods.

The source bit should allow bind providers to include the source info on records they originate.

This should allow us to replace a large part of the the code in octodns-powerdns that deals with parsing it's flavor of rdata. Namely it quotes a lot of the non-trivial strings. It would also use the `source` parameter.

Otherwise this should be a complete noop for existing stuff (that isn't quoted.(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v1.?.? - 2023-??-?? - ???
+
+* Record.from_rrs supports `source` parameter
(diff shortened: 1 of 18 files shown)


ross (commented):
/cc modules ci bump

Issue #1074, opened 2023-09-28
Distinguish between "fallback" pool and "backup" values in the documentation


istr:
While testing the GCore provider (in search of a cheaper alternative for ns1), I stumbled upon a concept that is worth considering for the documentation, for clarity.

Currently, there are two types of "fallback", both with the same term.

See https://github.com/octodns/octodns/blob/main/docs/dynamic_records.md

1. the fallback to the default pool if none of the rules match
2. the fallback to the values of the entry if everything else fails

Within gcore, this is implemented using two different meta flags: `fallback` for the pool fallback and `backup` for the last resort. Even though most of the other providers implement it otherwise, I like the concept, because it adds clarity.

So my proposal would be to rephrase the documentation. Instead of

>  These values become a non-healthchecked ~default pool~, generally it should be
> a superset of the catch-all pool and include enough capacity to try and
> serve all global requests (with degraded performance.) The main case they
> will come into play is if all dynamic healthchecks are failing, either on
> the service side or if the providers systems are expeiencing problems.

I would rather write

>  These values become a non-healthchecked _backup_. Generally it should be
> a superset of the catch-all pool and include enough capacity to try and
> serve all global requests (with degraded performance.) The main case they
> will come into play is if all dynamic healthchecks are failing, either on
> the service side or if the providers systems are expeiencing problems.

This could help to point out that the global entries in `values` are _not a pool_ but merely a last resort (a backup) in the sense that they don't have the features of a pool.

ross:
Naming things is hard :grin:, seem reasonable to call it backup to elucidate the difference.. I assume we'd want to update the associated visualization as well. Should I PR it or were you planning on it?

istr:
@ross I wanted to put this up for discussion first before I put a PR. Maybe @viranch has thoughts on that as well?

istr:
@ross I probably cannot make a PR quickly (next few days). So if you'd like to adopt the proposal - feel free to PR it. :smiley:

PR #1078, opened 2023-09-29
Rework dynamic documentation to call using the default pool a backup rather than fallback

ross:
/cc Fixes https://github.com/octodns/octodns/issues/1074 see it for discussion
/cc @istr


ross:
[Rendered](https://github.com/octodns/octodns/blob/backup-pool-doc/docs/dynamic_records.md)

PR #1081, opened 2023-10-05
CI pulls python versions from org secret

ross:
In order to change the versions we build against going forward it'll just be a matter of changing the ORG-level variable `PYTHON_VERSIONS_ACTIVE`. All of the repos that work with the current/up-to-date version list will be pointed at that with PRs similar to this one. 

For the couple of repos that need custom versions we can just leave the versions hard coded into the workflow file until if/when they're back in sync.

/cc @octodns/review as an FYI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         # Tested versions based on dates in https://devguide.python.org/versions/#versions
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: [${{fromJson(secrets.PYTHON_VERSIONS_ACTIVE)}}]
     steps:
       - uses: actions/checkout@v2
       - name: Setup python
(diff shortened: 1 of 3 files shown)


ross (commented):
/cc modules ci bump

ross (commented):
/cc modules ci bump again, for newer actions versions

ross:
https://github.com/octodns/terraformed/pull/22 is the final piece of this. From now all all python version changes can be made with a single PR to the terraformed repo. Once that's applied all of the repos will start using those versions and requiring the new contexts in one-shot. Should make life just a bit easier :grin:

ross:
https://github.com/octodns/terraformed/pull/23 will add Python 3.12 to the version matrix as the first use of this new functionality.

ross:
Well, fsck! Apparently forks don't get access to variables. I can understand not giving them secrets, but vars seems dumb. 

There is [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), but that runs on the target branches code and not the PRs so that's not useful.

ACAICT there's no solution which gives access to variables and runs in the context of the PR's code https://github.com/orgs/community/discussions/44322.

PR #1084, opened 2023-10-07
Try pulling python versions from a json file

ross:
Apparently variables (which are not "secrets") aren't available to 3rd party PRs so thus can't be used in CI jobs like I did in https://github.com/octodns/octodns/pull/1081#issuecomment-1751293056 ☹️. 

The approach here is to have a json file in octodns/octodns (added in this PR) that drives its matrix with a simple load, and then all the modules would fetch the raw version of the file from `main` to get their versions. It's ugly/rube-goldberg-ish, but so long as variables are treated sensitively I don't see a better option. 

/cc https://github.com/octodns/octodns/pull/1081#issuecomment-1751293056 and https://github.com/orgs/community/discussions/44322#discussioncomment-7212954 for more info/discussion


ross:
This seems to work and shouldn't have the 3rd party fork problem 😞

PR #1085, opened 2023-10-07
Add Python 3.12 to the ci test matrix and make it the current version

ross:
/cc https://github.com/octodns/octodns-docker/pull/97
/cc Replaces https://github.com/octodns/terraformed/pull/23

### DIFF
--- a/.ci-config.json
+++ b/.ci-config.json
@@ -1,4 +1,4 @@
 {
-  "python_version_current": "3.11",
-  "python_versions_active": ["3.8", "3.9", "3.10", "3.11"]
+  "python_version_current": "3.12",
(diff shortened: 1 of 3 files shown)


viranch:
Would be nice to have the CI honor the version matrix in the PR instead of `main`.

ross:
> Would be nice to have the CI honor the version matrix in the PR instead of `main`.

It will for octodns/octodns, the others will always be use whatever lives in main there.

viranch:
Ah yes missed there's a CI run for 3.12 in this PR.

ross:
I think the best flow for these PRs is to open the octodns one first, as here, make sure it's happy and then merge. Then go through and run each of the modules, either standalone or as part of PRs with needed changes (as will be the case here.) All of those PRs will be linked up.

PR #1086, opened 2023-10-12
Rename script/test-module to script/cibuild-module, update for py3.12

ross:
/cc https://github.com/octodns/octodns/pull/1085 for related work


ross (commented):
/cc ci build modules

ross (commented):
/cc ci modules bump, now with +x

Issue #1088, opened 2023-05-02
"dns.exception.SyntaxError: string too long" when updating DKIM TXT record


paulgear:
It appears that octodns-bind doesn't chunk long lines when attempting to update records.  Following is a lightly-edited excerpt of one attempt:

```
$ octodns-sync --config-file config/main.yml --doit --force
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: config_file=config/main.yml (octoDNS 0.9.21)
2023-05-01T19:27:06  [140206852075520] INFO  Manager _config_executor: max_workers=2
2023-05-01T19:27:06  [140206852075520] INFO  Manager _config_include_meta: include_meta=False
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: global_processors=[]
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=cloudflare (octodns_cloudflare 0.0.2)
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=digitalocean (octodns_digitalocean 0.0.2)
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=publicdns (octodns_bind 0.0.1)
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=writeyaml (octodns.provider.yaml 0.9.21)
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=yaml (octodns.provider.yaml 0.9.21)
2023-05-01T19:27:06  [140206852075520] INFO  Manager __init__: provider=zonefile (octodns_bind 0.0.1)
2023-05-01T19:27:06  [140206852075520] INFO  Manager sync: eligible_zones=[], eligible_targets=[], dry_run=False, force=True, plan_output_fh=<stdout>
2023-05-01T19:27:06  [140206852075520] INFO  Manager sync:   zone=gear.email.
2023-05-01T19:27:06  [140206852075520] INFO  Manager sync:   sources=['yaml'] -> targets=['digitalocean', 'publicdns']
2023-05-01T19:27:06  [140206818149952] INFO  YamlProvider[yaml] populate:   found 9 records, exists=False
2023-05-01T19:27:06  [140206818149952] INFO  DigitalOceanProvider[digitalocean] plan: desired=gear.email.
2023-05-01T19:27:07  [140206818149952] WARNING DigitalOceanProvider[digitalocean] populate: skipping unsupported SOA record
[...]

PR #1089, opened 2023-10-15
Fixed handling of chunking large TXT values for users of `rrs`

ross:
Fixed issues with handling of chunking large TXT values for providers that use the in-built `rrs` method

* `chunked_values` returns the type, not plain strings
* `ValuesMixin` adds support for getting special values when `rrs` is used, utilized by `_ChunkedValuesMixin`

/cc Fixes https://github.com/octodns/octodns/issues/1088 @paulgear @yzguy @binlan

### DIFF
--- a/octodns/record/chunked.py
+++ b/octodns/record/chunked.py
@@ -18,7 +18,7 @@ class _ChunkedValuesMixin(ValuesMixin):
             for i in range(0, len(value), self.CHUNK_SIZE)
         ]
         vs = '" "'.join(vs)
-        return f'"{vs}"'
+        return self._value_type(f'"{vs}"')
 
     @property
     def chunked_values(self):
--- a/octodns/record/chunked.py
+++ b/octodns/record/chunked.py
@@ -27,6 +27,7 @@ class _ChunkedValuesMixin(ValuesMixin):
             values.append(self.chunked_value(v))
         return values
 
+    @property
     def rr_values(self):
         return self.chunked_values
 
--- a/octodns/record/chunked.py
+++ b/octodns/record/chunked.py
@@ -27,6 +27,9 @@ class _ChunkedValuesMixin(ValuesMixin):
             values.append(self.chunked_value(v))
         return values
 
+    def rr_values(self):
+        return self.chunked_values
+
 
 class _ChunkedValue(str):
     _unescaped_semicolon_re = re.compile(r'\w;')
--- a/tests/test_octodns_record_txt.py
+++ b/tests/test_octodns_record_txt.py
@@ -142,3 +142,39 @@ class TestRecordTxt(TestCase):
         self.assertEqual(single.values, chunked.values)
         # should be chunked values, with quoting
         self.assertEqual(single.chunked_values, chunked.chunked_values)
+
+    def test_rr(self):
+        zone = Zone('unit.tests.', [])
+
+        # simple TXT
+        record = Record.new(
+            zone,
+            'txt',
+            {'ttl': 42, 'type': 'TXT', 'values': ['short 1', 'short 2']},
+        )
+        self.assertEqual(
+            ('txt.unit.tests.', 42, 'TXT', ['"short 1"', '"short 2"']),
+            record.rrs,
+        )
+
+        # long chunked text
+        record = Record.new(
+            zone,
+            'txt',
+            {
+                'ttl': 42,
+                'type': 'TXT',
+                'values': [
+                    'before',
+                    'v=DKIM1\\; h=sha256\\; k=rsa\\; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAx78E7PtJvr8vpoNgHdIAe+llFKoy8WuTXDd6Z5mm3D4AUva9MBt5fFetxg/kcRy3KMDnMw6kDybwbpS/oPw1ylk6DL1xit7Cr5xeYYSWKukxXURAlHwT2K72oUsFKRUvN1X9lVysAeo+H8H/22Z9fJ0P30sOuRIRqCaiz+OiUYicxy4xrpfH2s9a+o3yRwX3zhlp8GjRmmmyK5mf7CkQTCfjnKVsYtB7mabXXmClH9tlcymnBMoN9PeXxaS5JRRysVV8RBCC9/wmfp9y//cck8nvE/MavFpSUHvv+TfTTdVKDlsXPjKX8iZQv0nO3xhspgkqFquKjydiR8nf4meHhwIDAQAB',
+                    'z after',
+                ],
+            },
+        )
+        vals = [
+            '"before"',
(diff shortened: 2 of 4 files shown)


ross (commented):
/cc ci bump modules

ross:
This also fixes https://github.com/octodns/octodns-bind/issues/38

/cc Fixes https://github.com/octodns/octodns-bind/issues/38

octodns/record/chunked.py
```diff
@@ -18,7 +18,7 @@ def chunked_value(self, value):
             for i in range(0, len(value), self.CHUNK_SIZE)
         ]
         vs = '" "'.join(vs)
```
istr, line 20: @ross Note that not all providers use `" "` as a separator.
For example, constellix uses `""` without a space.

We should give the providers an opportunity to hook in and redefine the separator. Currently, providers need to "brute force" this, for example by replacing `" "` with `""`.

tests/test_octodns_record_txt.py
```diff
@@ -142,3 +142,39 @@ def test_long_value_chunking(self):
...
+                'values': [
+                    'before',
+                    'v=DKIM1\\; h=sha256\\; k=rsa\\; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAx78E7PtJvr8vpoNgHdIAe+llFKoy8WuTXDd6Z5mm3D4AUva9MBt5fFetxg/kcRy3KMDnMw6kDybwbpS/oPw1ylk6DL1xit7Cr5xeYYSWKukxXURAlHwT2K72oUsFKRUvN1X9lVysAeo+H8H/22Z9fJ0P30sOuRIRqCaiz+OiUYicxy4xrpfH2s9a+o3yRwX3zhlp8GjRmmmyK5mf7CkQTCfjnKVsYtB7mabXXmClH9tlcymnBMoN9PeXxaS5JRRysVV8RBCC9/wmfp9y//cck8nvE/MavFpSUHvv+TfTTdVKDlsXPjKX8iZQv0nO3xhspgkqFquKjydiR8nf4meHhwIDAQAB',
+                    'z after',
+                ],
+            },
+        )
+        vals = [
+            '"before"',
+            '"v=DKIM1\\; h=sha256\\; k=rsa\\; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAx78E7PtJvr8vpoNgHdIAe+llFKoy8WuTXDd6Z5mm3D4AUva9MBt5fFetxg/kcRy3KMDnMw6kDybwbpS/oPw1ylk6DL1xit7Cr5xeYYSWKukxXURAlHwT2K72oUsFKRUvN1X9lVysAeo+H8H/22Z9fJ0P30sOuRIRqCaiz+OiUYicxy4xrpfH" '
```
istr, line 176: As mentioned above, this depends on the provider. It could be `...Caiz+OiUYicxy4xrpfH"' (without the space) as well.

istr (commented):
I would add the ability for the providers to change the separator `" "`, for example using a function that the provider can override.

octodns/record/chunked.py
ross: Ugh, that's an annoying inconsistency. Will have to dig into/think about options for how the providers could tell the record what it needs. Since `chunked_values` is a pre-existing property there's not a clean way to pass it in there.

polarathene: The current PR is still a fix/improvement for some providers like Bind right? No harm in going forward with this and addressing a way for other providers to configure a separator?

As mentioned, those providers can apparently workaround it until then, and there is only one provider cited with the need for customizing the separator? Perhaps knowing how many other providers would need to customize the separator would be good to investigate?

ross: It looks like Constellix already uses `chunked_value` and I don't see anything there turning `" "` into `""`. It just seems like it's un-escaping the `;`. 

https://github.com/octodns/octodns-constellix/blob/00f51bf0ecf653d2e2bb75a0bc659cd709c8ccfe/octodns_constellix/__init__.py#L802-L807. 

That currently broken?

istr: Yep, it is.
```
* constellix (ConstellixProvider)
*   Update
*     <TxtRecord TXT 1800, mail._domainkey.ingo-struck.com., ['v=DKIM1\; k=rsa\; s=email\; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA0uxLHfnLxBis/ZTlMC2d4oFT0rJ7SeqZN+AAW7JRtx+yZ8Fu09exnmttYpKaCSupKj3ftcvHrlu1ji4NYPlhX9qlyNwYentcFBfepp75GUTj8mtcMak72409IIkLP2fr3kHBBaDk4JR5FwMZFZsgL1IVss4WUQybKppACvEUcxeOHrftUH7Rsr2X""rCCRJopM1QV1AaQum/i3dw0ZBZgSmY9/oTtMeOco8HX/7as2M57KO8Mw+tYyx6+RyaeKOC/X9r1Da8Ayvz2P0LEojaRh4PX3LKYZiTntviO+Ed8lGrxmBXN5Jr+uqeJ+MQozoTGU+395zoAoTHpILxLr650ijQIDAQAB']> ->
*     <TxtRecord TXT 1800, mail._domainkey.ingo-struck.com., ['v=DKIM1\; k=rsa\; s=email\; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA0uxLHfnLxBis/ZTlMC2d4oFT0rJ7SeqZN+AAW7JRtx+yZ8Fu09exnmttYpKaCSupKj3ftcvHrlu1ji4NYPlhX9qlyNwYentcFBfepp75GUTj8mtcMak72409IIkLP2fr3kHBBaDk4JR5FwMZFZsgL1IVss4WUQybKppACvEUcxeOHrftUH7Rsr2XrCCRJopM1QV1AaQum/i3dw0ZBZgSmY9/oTtMeOco8HX/7as2M57KO8Mw+tYyx6+RyaeKOC/X9r1Da8Ayvz2P0LEojaRh4PX3LKYZiTntviO+Ed8lGrxmBXN5Jr+uqeJ+MQozoTGU+395zoAoTHpILxLr650ijQIDAQAB']> (config)
```

However, the fix is/would be trivial. (replacing `" "` with `""`)

ross: > So yes, if this is seemingly the only provider, let's move forward with this one and care for the "outliers" individually as @polarathene pointed out.

Sounds good. I tend to consider bind standard compliant so hopefully it is just a quirk w/Constellix.

istr: I will provide a patch for constellix when this one is merged.

ross:
Thanks for the discussion/digging/thinking.

binlan:
works like a charm. cant wait to see it in the next release. @ross do you have an idea when it will arrive?

ross:
Not atm, few things in flight that should probably land first. Won't be too long, but no specific date. In the meantime you can install w/a SHA https://github.com/octodns/octodns#installing-a-specific-commit-sha, that's actually what I do more often than not.

binlan:
tested it already with the sha install method. thats why i was asking. ok, will use it in that way in a dedicated python env until the release. thx for your quick response.

PR #1092, opened 2023-11-01
ExcludeRootNsChanges processor that will error (or warn) if plan includes a   change to root NS records

ross:
/cc https://github.com/octodns/octodns-route53/issues/75
/cc @lbrachet


istr:
@ross looks good to me.

Issue #1093, opened 2023-11-08
Processor to filter out records based on value


Solvik:
Hi there!

I'm opening this issue to discuss the ability to add a processor based on record's value

We use octodns to synchronize an internal PowerDNS instance to route53. We have the case where internal records are then synchronized to a public DNS zone and we want to avoid that.

I've seen that in the latest versions of octodns, you've introduced an amazing feature such as processors.

Would you be open to a contribution ? 

I see two ways:
- create a dedicated ExcludePrivateIpFilter
- create a generic ValueRejectlistFilter that would use a regex list `/ ^127\./` etc.

What do you think is the best ? I can do the contribution

ross:
> I see two ways:
> ...

For IP addresses i think filtering by a list of CIDRs would be the appropriate tool for the job. It could potentially have a boolean flag parameter that auto added the private ranges: `127.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`. Potentially some other stuff off the full reserved IP address list, https://en.wikipedia.org/wiki/Reserved_IP_addresses. Obviously should support IPv6 as well.

It'd be possible to filter on the value of other record types like TXT too, but without a specific use case I'd be hesitant to implement it. Value based filtering would get really complicated messy if it tried to extend to non-trivial types.

Solvik:
Ok thanks for your feedback, I'll propose a MR :)

thystips:
Hello!

I suggest implementing Allow and Reject filters options for specific PTR records too.

ross:
> I suggest implementing Allow and Reject filters options for specific PTR records too.

Not sure I follow. What PTRs are included would be driven off the zones configured since the zone is the subnet. To avoid pushing private PTR zones up they would just be left out of the (public) targets.

If you're talking about filtering on the value side of the PTR that'd be a totally different tool, and would apply equally to other record types. That'd be doable, but definitely not part of this PTR.

thystips:
I read issue to fast and only keep in mind filtering and record's value. I will open another issue !

ross:
Closed by https://github.com/octodns/octodns/pull/1094 and https://github.com/octodns/octodns/pull/1098 adds some extra functionality.

PR #1094, opened 2023-11-08
add NetworkValueRejectlistFilter and NetworkValueAllowlistFilter processors

Solvik:
Hi

Here's my proposal for two new processor filters to allow or reject A and AAAA based on a list of CIDR

As discussed in #1093

### DIFF
--- a/octodns/processor/filter.py
+++ b/octodns/processor/filter.py
@@ -127,35 +127,6 @@ class _NameBaseFilter(BaseProcessor):
     process_target_zone = _process
 
 
-class _NetworkValueBaseFilter(BaseProcessor):
-    def __init__(self, name, _list):
-        super().__init__(name)
-        self.networks = []
-        for value in _list:
-            try:
-                self.networks.append(ip_network(value))
-            except ValueError:
-                raise ValueError(f'{value} is not a valid CIDR to use')
-
-    def _process(self, zone, *args, **kwargs):
-        for record in zone.records:
-            if record._type not in ['A', 'AAAA']:
-                continue
-
-            ips = [ip_address(value) for value in record.values]
-            if any(
-                ip in network for ip, network in product(ips, self.networks)
-            ):
-                self.matches(zone, record)
-            else:
-                self.doesnt_match(zone, record)
-
-        return zone
-
-    process_source_zone = _process
-    process_target_zone = _process
-
-
 class NameAllowlistFilter(_NameBaseFilter, AllowsMixin):
     '''Only manage records with names that match the provider patterns
 
@@ -218,6 +189,35 @@ class NameRejectlistFilter(_NameBaseFilter, RejectsMixin):
 
     def __init__(self, name, rejectlist):
         super().__init__(name, rejectlist)
+
+
+class _NetworkValueBaseFilter(BaseProcessor):
+    def __init__(self, name, _list):
+        super().__init__(name)
+        self.networks = []
+        for value in _list:
+            try:
+                self.networks.append(ip_network(value))
+            except ValueError:
+                raise ValueError(f'{value} is not a valid CIDR to use')
+
+    def _process(self, zone, *args, **kwargs):
+        for record in zone.records:
+            if record._type not in ['A', 'AAAA']:
+                continue
+
+            ips = [ip_address(value) for value in record.values]
+            if any(
+                ip in network for ip, network in product(ips, self.networks)
+            ):
+                self.matches(zone, record)
+            else:
+                self.doesnt_match(zone, record)
+
+        return zone
+
+    process_source_zone = _process
+    process_target_zone = _process
 
 
 class NetworkValueAllowlistFilter(_NetworkValueBaseFilter, AllowsMixin):
--- a/octodns/processor/filter.py
+++ b/octodns/processor/filter.py
@@ -144,8 +144,7 @@ class _NetworkValueBaseFilter(BaseProcessor):
 
             ips = [ip_address(value) for value in record.values]
             if any(
-                ip in network
-                for ip, network in product(ips, self.networks)
+                ip in network for ip, network in product(ips, self.networks)
             ):
                 self.matches(zone, record)
             else:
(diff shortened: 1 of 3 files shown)


octodns/processor/filter.py
```diff
@@ -125,6 +127,35 @@ def _process(self, zone, *args, **kwargs):
...
+
+    def _process(self, zone, *args, **kwargs):
+        for record in zone.records:
+            if record._type not in ['A', 'AAAA']:
+                continue
+
+            if any(
+                ip_address(value) in network
+                for value, network in product(record.values, self.networks)
+            ):
```
ross, line 148: Nice. Like the use of `any` here and the `product` bit cleans things up nicely. Small tweak would avoid recreating the IP addresses for each network test.

```suggestion
            ips = [ip_address(value) for value in record.values]
            if any(
                ip in network
                for ip, network in product(ips, self.networks)
            ):
```

octodns/processor/filter.py
```diff
@@ -125,6 +127,35 @@ def _process(self, zone, *args, **kwargs):
...
+        for value in _list:
+            try:
+                self.networks.append(ip_network(value))
+            except ValueError:
+                raise ValueError(f'{value} is not a valid CIDR to use')
+
+    def _process(self, zone, *args, **kwargs):
+        for record in zone.records:
+            if record._type not in ['A', 'AAAA']:
+                continue
```
ross, line 143: Kind of funky case here since these records technically don't match the filter, but if this called `doesnt_match` then they'd all get removed. I think this is the best path, to not even consider them, but it probably needs to be documented in the docstrings below that the filters only operate on `A` and `AAAA` records and that all other types will be left as-is.

IF someone actually wanted to get rid of the other types they could use the type based filters for that.

tests/test_octodns_processor_filter.py
```diff
@@ -161,6 +163,38 @@ def test_regex(self):
...
+    )
+    zone.add_record(doesnt)
+    matchable1 = Record.new(
+        zone, 'private-ipv6', {'type': 'AAAA', 'ttl': 42, 'value': 'fd12:3456:789a:1::1'}
+    )
+    zone.add_record(matchable1)
+    matchable2 = Record.new(
+        zone, 'public-ipv6', {'type': 'AAAA', 'ttl': 42, 'value': 'dead:beef:cafe::1'}
+    )
+    zone.add_record(matchable2)
```
ross, line 183: Probablby best to add a non A/AAAA record here to test things with and verify that it's left alone and remains the the zone. Not really testing the new provider here, but more making sure that the Accepts/RejectsMixin behave as expected in case they get modified at some point in the future.

ross (commented):
Looks good. Few minor comments inline for thought/discussion.

ross:
Looks like ci failures are formatting related. `./script/format` should get that addressed. If you want to effectively do everytihng CI does locally you can run `./script/cibuild`.

octodns/processor/filter.py
Solvik: thanks for this :)

tests/test_octodns_processor_filter.py
Solvik: you're absolutely right, will do

Solvik: I've added a simple TXT record to handle that case

octodns/processor/filter.py
Solvik: I've added this comment

ross (approved):
Looks good. Thanks!

ross:
Failing coverage for the invalid ip address except case. Looking at the A/AAA record types I don't believe it's possible to create them with invalid values so it's probably safe to drop the try/except wrapper.

Screenshot 2023-11-09 at 11 44 14 AM

Solvik:
It's made to validate the CIDR supplied by an user in the configuration file, I think it's necessary, I should add a coverage test

Solvik:
After some testing, I'm seeing an issue which may be related to the octodns-route53 module

With the following configuration:
```
processors:
  exclude-private-ip:
    class: octodns.processor.filter.NetworkValueRejectlistFilter
    rejectlist:
      - 10.0.0.0/8

zones:
  foo.com.:
    sources:
    - pdns
    processors:
    - exclude-private-ip
    targets:
    - route53
    - yaml
```

The pdns source has records such as:
```
example-record:
- ttl: 300
  type: A
  value: 10.1.0.1
- ttl: 300
  type: TXT
  value: some value
```

I can see the A records being filtered out from the yaml output but not from the route53 zone with those currently existing records, what do you think?

ross:
> It's made to validate the CIDR supplied by an user in the configuration file, I think it's necessary, I should add a coverage test

Ah. I wasn't paying enough attention. Obviously that's the CIDR part of things and not the IP. 

> I can see the A records being filtered out from the yaml plan but not from the route53 plan, which currently has these records, what do you think?

That's surprising. I would expect both providers to be getting the same plan. I'll have to poke around and test things out myself.

ross:
> I can see the A records being filtered out from the yaml plan but not from the route53 plan, which currently has these records, what do you think?

Do the records already exist in Route53? If so this may be a side effect of how filtering works. It effectively makes those records invisible to octodns. If they already exist it will leave them alone/make them unmanaged. 

It won't create new ones though. 

That's b/c the filtering happens on both the source and target:

```python
    process_source_zone = _process
    process_target_zone = _process
```

It might be interesting to explore adding an option to all the filters to optionally disable one or both of these.

If target zone is disabled you could get a plan to delete them and never be able to add new ones. 

It's harder to imagine a use case for source zone. I think disabling it would result in trying to add the records every time, likely failing since one would already exist. Maybe it'd just make sense to disable the target side.

Will explore that, but if that's what's happening here I don't believe it's related to the PR/blocking it moving forward.

ross:
> Will explore that, but if that's what's happening here I don't believe it's related to the PR/blocking it moving forward.

Implemented over in https://github.com/octodns/octodns/pull/1098. Seems to work well/as intended. Will land it after this change has, once it's been updated for these new filter types.

Solvik:
> > Do the records already exist in Route53? If so this may be a side effect of how filtering works. It effectively makes those records invisible to octodns. If they already exist it will leave them alone/make them unmanaged.

Thanks for digging, that's exactly my case ; I had existing records

I figured I could delete it manually and see that OctoDNS ignore them for the next run as intended.

The feature in #1098 would make this behavior indempotent so that's really nice ! Thanks again

ross (approved):
/cc ci modules bump

PR #1098, opened 2023-11-11
Add include_target option to most processor.filter processors

ross:
Most processor.filter processors now support an include_target flag that can be set to False to leave the target zone data untouched, thus remove any existing filtered records. Default behavior is unchanged and filtered records will be completely invisible to octoDNS

/cc https://github.com/octodns/octodns/pull/1094#issuecomment-1806595582 which spawned the idea


ross (commented):
/cc ci modules bump

Issue #1096, opened 2023-11-10
"pip install --no-binary" fails


viranch:
I have a use-case to not use binary wheel for installing octodns, but the pip install command fails:

```
(env) → /tmp/testenv » python -m pip install --no-binary octodns octodns==1.0.0
Collecting octodns==1.0.0
  Downloading octodns-1.0.0.tar.gz (532 kB)
     |████████████████████████████████| 532 kB 5.2 MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
WARNING: Discarding https://files.pythonhosted.org/packages/6a/a9/fccc8c6120d51ea98d67a455980094a18eacf8e581c68a1e78ca0044c8cf/octodns-1.0.0.tar.gz#sha256=ccfc9c17789f71879bb10d13cba6bfe85e8eb2ba4b7a04bd3d13f10760afd502 (from https://pypi.org/simple/octodns/) (requires-python:>=3.8). Requested octodns==1.0.0 from https://files.pythonhosted.org/packages/6a/a9/fccc8c6120d51ea98d67a455980094a18eacf8e581c68a1e78ca0044c8cf/octodns-1.0.0.tar.gz#sha256=ccfc9c17789f71879bb10d13cba6bfe85e8eb2ba4b7a04bd3d13f10760afd502 has inconsistent version: filename has '1.0.0', but metadata has '1.0.0+unknown'
ERROR: Could not find a version that satisfies the requirement octodns==1.0.0 (from versions: 0.8.0, 0.8.4, 0.8.5, 0.8.6, 0.8.7, 0.8.8, 0.9.0, 0.9.1, 0.9.2, 0.9.3, 0.9.4, 0.9.5, 0.9.6, 0.9.8, 0.9.9, 0.9.10, 0.9.11, 0.9.12, 0.9.13, 0.9.14, 0.9.15, 0.9.16, 0.9.17, 0.9.18, 0.9.19, 0.9.20, 0.9.21, 1.0.0rc0, 1.0.0rc1, 1.0.0, 1.1.0, 1.1.1, 1.2.0, 1.2.1)
ERROR: No matching distribution found for octodns==1.0.0
(env) → /tmp/testenv »
```

This seems to be the culprit, but can't figure out why the metadata has this weird version:
```
inconsistent version: filename has '1.0.0', but metadata has '1.0.0+unknown'
```

This is happening for all versions of octodns as well as octodns-* packages.

[...]

PR #1099, opened 2023-11-11
Simplify versioning

ross:
setup.py had non-trivial logic in it to try and stick a portion of the commit SHA onto the end of the version number when things were installed from the repo. This was fragile and had issues in cases like https://github.com/octodns/octodns/issues/1096 so this PR drops that logic and just uses the current version number. 

It also add `__version__` which in my research/testing for https://github.com/octodns/octodns/issues/1096 I ran into as being a (more) standard/recommended name for the version, per pep8.

/cc https://peps.python.org/pep-0008/#module-level-dunder-names which mentions __version__
/cc Fixes https://github.com/octodns/octodns/issues/1096 @viranch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   of a standard, per pep-8. __VERSION__ is deprecated and will go away in 2.x
 * Fixed issues with handling of chunking large TXT values for providers that use
   the in-built `rrs` method
+* Removed code that included sha in module version number when installing from
+  repo a it caused problems with non-binary installs.
 
 #### Stuff
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Fixed issues with handling of chunking large TXT values for providers that use
   the in-built `rrs` method
 * Removed code that included sha in module version number when installing from
-  repo a it caused problems with non-binary installs.
+  repo as it caused problems with non-binary installs.
 
 #### Stuff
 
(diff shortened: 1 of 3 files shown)


istr:
Looks good and makes sense. Confirmed that this fixes #1096.

CHANGELOG.md
```diff
@@ -1,9 +1,18 @@
...
-* Added ZoneNameFilter processor to enable ignoring/alerting on type-os like
-  octodns.com.octodns.com
+#### Noteworthy changes
+
+* Added octodns.__version__ to replace octodns.__VERSION__ as the former is more
+  of a standard, per pep-8. __VERSION__ is deprecated and will go away in 2.x
 * Fixed issues with handling of chunking large TXT values for providers that use
   the in-built `rrs` method
+* Removed code that included sha in module version number when installing from
+  repo a it caused problems with non-binary installs.
```
viranch, line 10: Typo: _as_ it caused...

setup.py
```diff
@@ -50,16 +48,7 @@ def long_description():
 
 
 def version():
```
viranch, line 50: Method no longer needed?

viranch (approved):
LGTM overall, minor comments. thanks for the fix!

setup.py
ross: Yeah. Just missed deleting it when I decided to just have the __version__ down below.

ross (commented):
/cc bump ci modules

PR #1101, opened 2023-11-12
__VERSION__ -> __version__ in script/release and internal refs

ross:
😦. Should have been done in the previous round (ideally.) 

This prefers __version__ when referencing the module's own version and prefers __version__ falling back to __VERSION__ when referring to modules. 

/cc https://github.com/octodns/octodns/pull/1099


ross (commented):
/cc ci modules bump

PR #1102, opened 2023-11-13
Fixed Record.data, now includes the octodns special section data

ross:
Fixed a bug/shortcoming in Record.data where it didn't include the `octodns` portion of the record's data. Record.copy also omitted it since it relies on .data for the heavy lifting. There are no known actual problems due to this bug, but it's possible (non-public) third party providers are susceptible. The most likely place to hit issues in is tests where data and/or copy are abused.

/cc https://github.com/octodns/octodns-cloudflare/pull/65 work on which lead to finding this issue


ross (commented):
/cc ci modules bump

ross:
* [cloudflare failure](https://github.com/octodns/octodns/actions/runs/6853225902/job/18633594351?pr=1102) is the one that's expected, will be fixed as part of https://github.com/octodns/octodns-cloudflare/pull/65

Will have to dig into [mythicbeasts](https://github.com/octodns/octodns/actions/runs/6853225902/job/18633598132?pr=1102) and [transip](https://github.com/octodns/octodns/actions/runs/6853225902/job/18633600248?pr=1102), but a quick skim leads me to suspect that it's the same sort of hit in testing, but not in practice issue, but we'll see...

viranch:
Does this mean populating from remote will now need to include the healthcheck settings for dynamic records? That's not happening ATM (IIRC) in any providers and it sounds like the plan will show all dynamic record updates due to mismatching octodns section (due to it missing in populated records but present in yaml).

viranch:
Looks like comparison still doesn't include octodns/healthcheck section, so my concern above is probably not applicable, but something to think about in future perhaps.

ross:
> Does this mean populating from remote will now need to include the healthcheck settings for dynamic records? 

I don't think that's directly tied to this change. It just preserves the data that is on the record the way it always should have. If you used `.copy()` it would have the data, but not if you called `.data` it didn't.
  
It's possible `octo-dump` will include stuff that it previously didn't, but I'll have to look over the YamlProvider and see if it's using `data` and if this changed things. 

> That's not happening ATM (IIRC) in any providers and it sounds like the plan will show all dynamic record updates due to mismatching octodns section (due to it missing in populated records but present in yaml).

Are you seeing this somewhere? I would have thought the problem would show up as false diffs before this change as I don't offhand know where the normal sync process would be using `.data`.

Will definitely do more testing with this before moving forward with it Any specific config/flow you want to point me at that shows problems?

ross:
> Looks like comparison still doesn't include octodns/healthcheck section, so my concern above is probably not applicable, but something to think about in future perhaps.

Yeah. As I dug into things based on your first comment, I realized that `Record.changes` doesn't consider `octodns`. That would be a bigger deal/more potential for issues since that area is fairly freeform and it wouldn't be able to cope with "defaults", i.e. something that isn't specified and getting a default being equivalent having the value specified with that default. 

It may never be possible to include it and we may always rely on providers to do whatever they need w/include_change and/or extra changes.



> It's possible `octo-dump` will include stuff that it previously didn't, but I'll have to look over the YamlProvider and see if it's using `data` and if this changed things.

As part of looking into this I found that `YamlProvider` is already manually copying in _octodns after using .data.

https://github.com/octodns/octodns/blob/65e87d0a465b0b1196cf6476245d609078420eff/octodns/provider/yaml.py#L389-L390

That can now go away.

viranch:
> I realized that `Record.changes` doesn't consider `octodns`.

Ok yeah I assumed this was also changing in this PR. All good since it's not, but to expand on my comment for future consideration:

I was referring to the fact that providers like NS1/Azure that support dynamic records, don't include `octodns.healthcheck` during the populate phase and rely on `_extra_changes` to detect changes in that area. I think since `octodns.healthcheck` is more first-class than rest of the provider-specific sections inside `octodns`, `octodns.healthcheck` should be considered in `Record.changes` (even provider-specific section should be considered in the long-term). I guess it will need to be done in a safe way as the providers will need to start populating healthcheck settings to avoid unwarranted `Update`s.

ross (commented):
/cc ci modules bump

ross:
Flawed test assumptions in modules addressed. This should be good to go now.

PR #1103, opened 2023-11-13
Update the modules action job's module list

ross:
Been a while, few changes. Notably removing dyn and adding a few new ones. Giving octodns-netbox a chance.


ross (commented):
/cc ci modules bump

ross:
octodns-netbox is still failing, removed.

PR #1117, opened 2023-12-04
Version 1.4.0 bump & changelog update

ross:
## v1.4.0 - 2023-12-04 - Minor Meta

* Record.lenient property added similar to other common/standard _octodns data
* Processor.process_source_and_target_zones added to support modifying both the
  desired and/or existing zones just prior to computing changes.
* Fix bug with Record.copy when values is an empty list []
* Fix an issue in MetaProcessor/Manager.include_meta where include_provider
  wasn't correctly taking effect


ross (commented):
/cc ci modules build

Issue #1119, opened 2023-12-06
Odd processor query


quistian:
I have a need to compress or strip off the outermost subdomain...  It does sound odd. 
Given the name of the fqdn is unique, I was wanting a processor that would in effect translate:

q123.123.bozo.net -> q123.bozo.net
n456.456.bozo.net -> n456.bozo.net
y222_test.NNN.bozo.net -> y222_test.bozo.net
etc.. 

Looking at the current filter.py processor code, is this something a customization one could do?
At the present time I am filtering the local yaml data. 

Thanks.

ross:
Not something that would currently be possible with octoDNS code as is. Processors are focused on record data & plans. There's no plug-in arch for modifying the zone data atm. 

You could create a custom provider that inherits from `YamlProvider` and does whatever mucking you want/need to do with the zone data. Should be fairly simple & straightforward to do that. 

Otherwise I'll have to look around in the code and see if there's a clean/clear pattern for how processing zones would fit into things.

quistian:
Many thanks. That's a start. And by "inherits" from YamlProvider, do you mean start with the current YamlProvider code as a template, and manage the file step of writing the data to the local YAML files?

ross:
So I was assuming you had sub zones configured and needed to muck with their names to push the data to the parent zone. E.g. you have `config/123.bozo.net.yaml` on disk and configured in the octodns yaml file and want `bozo.net.` to actually have the data pushed to it.

If instead you have `config/bozo.net.yaml` and have `q123.123:` defined as a record in the zone you should be able to do that modification in a custom processor (it would probably be too unique of a use case to be included in octoDNS core.)

quistian:
Yes.. What I want is to have the data at the leaf subdomain to be pushed up one level.  In effect take off the outermost subdomain combining and pushing all the RRs up one level. I have a case where by design none of the names will collide. I shall look at #1120. Thanks.

ross:
If you're merging records between two zones (pushing trimming a childs records to fit in its parents) #1120 probably won't be of use. If all of the records are already in a single octoDNS/provider zone and some just have the extra subdomain that you want to remove a custom processor should be able to do that.

Feel like we're to the point where it'll probably require knowing the full setup for me to be able to make suggestions which may or may not make sense/be needed :grin:

quistian:
[...]

PR #1120, opened 2023-12-06
Processors.process_zone_config, dynamic zone config converted

ross:
* `Processors.process_zone_config` method added to allow processors that work with the zone config data. Configured with `manager.zone-processors: []`, default is ['dynamic-zone-config']
* Converted dynamic zone config to be a processors ^, if zone-processors are explicitely configured and dynamic zone config is desired `dyanmic-zone-config` must be included in the list as the desired position
* Loads of additional testin/vetting for the above functionality

/cc https://github.com/octodns/octodns/issues/1119 which lead to this exploration/work which I'm fairly happy with, but may or may not be what it's looking for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 * Support added for config env variable expansion on nested levels, not just
   top-level provider/processor keys
+* `Processors.process_zone_config` method added to allow processors that work
+  with the zone config data. Configured with `manager.zone-processors: []`,
+  default is ['dynamic-zone-config']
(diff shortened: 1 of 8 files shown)


ross:
Not fully sure about this one. I really like the shape of it and it generalizes the dynamic zone config into something that would be pretty powerful, but I'm not sure if anyone would make use of it. 

/cc @octodns/review for 💭s

ross:
Going to park this since there's not a clear need for it and it touches some central stuff non-trivially. I generally like the idea so if anyone comes up with a need for it we can revisit.

Issue #1121, opened 2023-12-07
Meta processor breaks octodns-sync, when CNAME records are being created or updated


fjaeckel:
We're seeing, that when a octodns-sync process would create or update a CNAME, that the sync process fails with below stack trace. The issue can be resolved by setting `include_meta: False` temporarily, deploying tthat code with the CNAME changes and rolling back with `include_meta: True`.

In this specific instance a CNAME record was _deleted_.

I suppose this is because a CNAME can only have 1 value and meta's TXT records could handle multiple records. As this would indicated:

```
File "/data/dns/env/lib/python3.11/site-packages/octodns/processor/meta.py", line 137, in _up_to_date
return existing is not None and _keys(existing.values) == _keys(
^^^^^^^^^^^^^^^
AttributeError: 'CnameRecord' object has no attribute 'values'. Did you mean: 'value'?
```

Full traceback:
```
Traceback (most recent call last):
File "/data/dns/env/bin/octodns-sync", line 8, in <module>
sys.exit(main())
^^^^^^
File "/data/dns/env/lib/python3.11/site-packages/octodns/cmds/sync.py", line 57, in main
manager.sync(
File "/data/dns/env/lib/python3.11/site-packages/octodns/manager.py", line 697, in sync
ps, d = future.result()
^^^^^^^^^^^^^^^
File "/usr/share/pyenv/versions/3.11.6/lib/python3.11/concurrent/futures/_base.py", line 449, in result
return self.__get_result()
^^^^^^^^^^^^^^^^^^^
File "/usr/share/pyenv/versions/3.11.6/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
raise self._exception
File "/usr/share/pyenv/versions/3.11.6/lib/python3.11/concurrent/futures/thread.py", line 58, in run
result = self.fn(*self.args, **self.kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/data/dns/env/lib/python3.11/site-packages/octodns/manager.py", line 471, in _populate_and_plan
plan = processor.process_plan(
^^^^^^^^^^^^^^^^^^^^^^^
File "/data/dns/env/lib/python3.11/site-packages/octodns/processor/meta.py", line 145, in process_plan
and self._up_to_date(plan.changes[0])
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1122, opened 2023-12-07
Fix bug in MetaProcessor up to date record check

ross:
Fix for bug in MetaProcessor _up_to_date check that was failing when there was a plan with a single change type with a single value, e.g. CNAME.

/cc Fixes https://github.com/octodns/octodns/issues/1121 @fjaeckel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## v1.?.0 - 2023-??-?? -
 
+* Fix for bug in MetaProcessor _up_to_date check that was failing when there was
+  a plan with a single change type with a single value, e.g. CNAME.
 * Support added for config env variable expansion on nested levels, not just
   top-level provider/processor keys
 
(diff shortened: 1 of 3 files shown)


fjaeckel (approved):

ross:
Found another level of issues with this, that actually may have always been present 🙃, anyway, reworking/cleaning up to simplify that code and doing a major pass through testing things to try and make sure everything is thought out and tested. Will push the results once that's done.

ross:
No more add a record in one method and update it in another which is where a lot of the complications/problems arose. This now generates and adds the record in `process_source_and_target_zones` with all of its values in one go at that time.

It still has to do the plan logic to make sure that the meta record doesn't cause plans/changes (unless the meta config changed.) 

I __think__ this should be in a lot better shape than it was previously and at the very least it should be less confusion to figure out how it works.

ross (commented):
/cc ci modules builds

Issue #1126, opened 2024-01-03
OctoDNS dry-run not failing on em dash in txt record


freekdegreef:
When trying to set this text record `v=spf1 –all` using OctoDNS without `--do-it` (which we use as a validate step in our CI pipeline) OctoDNS fails to pick up the em dash posing as a hyphen. When actually making the changes this results in an error:
```
Invalid value for 'entity.change.additions[example.com.][TXT].rrdata[0]': '"v=spf1 –all"'
```
In our case this is with the Google DNS provider but this might also apply to other providers.

ross:
It looks like utf-8 chars aren't legal in TXT values, but atm i'm having a hard time finding something that definitively says that's the case. 

https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14 doesn't seem to specify what's legal in a `character-string`, though it's age would lead me to believe it's assuming ASCII.

freekdegreef:
Thanks @ross ✌🏻

PR #1127, opened 2024-01-03
Add ASCII validation to ChunkedValue, TXT/SPF

ross:
It appears that TXT values can only contain ASCII characters. This PR adds a validation ensuring that's the case. See https://github.com/octodns/octodns/issues/1126 for more info.

/cc Fixes https://github.com/octodns/octodns/issues/1126


ross:
Tried a couple providers and both reject non-ASCII TXT values which i guess is evidence towards confirming the conclusion.

viranch (approved):
Straight forward enough, LGTM!

PR #1128, opened 2024-01-05
Add list_zones to AutoArpa

baest:
- This makes it possible to use dynamic zone config with AutoArpa
- Document what gotchas that entails
- Add tests both with another source (for generating zones to populate) and without (no zones to populate)


ross (approved):
Looks good. Thanks for including the tests

PR #1131, opened 2024-01-11
Record.values/value can be used interchangeably 

ross:
A record like:

```yaml
something:
  type: TXT
  values: one value
```

~~Will now throw a validation error~~ will be treated as if it used `value`:

```console
...
octodns.record.exception.ValidationError: Invalid record "txt.exxampled.com.", ./config/exxampled.com.yaml, line 248, column 3
  - single value provided under values key, "one value"
```

~~single value provided under values key~~

/cc https://github.com/octodns/octodns-cloudflare/issues/73 which this is a partial fix for

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@
 * Support added for config env variable expansion on nested levels, not just
   top-level provider/processor keys
 * _ChunkedValue ASCII validation added, SPF & TXT
-* Minor validation improvement to catch cases where Record values is provided a
-  single a-single-value by accident instead of a list, a common type-o that
-  can results in each character being a value in the resulting Record data
+* Re-work value/values handling to always try and do the "right" thing based on
+  the content, so both singular values and lists will be handled identically
+  regardless of whether the key is value or values. This may result in
+  changes/fixes on the first sync after updating IFF you currently have
+  `values: a-single-thing`, which would have previously been pushed up as bunch
+  of single character values.
 
 ## v1.4.0 - 2023-12-04 - Minor Meta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Support added for config env variable expansion on nested levels, not just
   top-level provider/processor keys
 * _ChunkedValue ASCII validation added, SPF & TXT
+* Minor validation improvement to catch cases where Record values is provided a
+  single a-single-value by accident instead of a list, a common type-o that
+  can results in each character being a value in the resulting Record data
 
 ## v1.4.0 - 2023-12-04 - Minor Meta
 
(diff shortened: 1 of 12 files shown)


viranch:
Would it be better to allow the flexibility and just coerce on our side?

```
values = values if isinstance(values, list) else [values]
```

ross:
> Would it be better to allow the flexibility and just coerce on our side?

I thought about that for a bit and was on the fence. The thing that ended up tipping me towards the validation was that if we auto-handle it people would see a change on version update, i.e. if they had an existing broken record it would go from having a ton of one char values to the corrected single value and they wouldn't necessarily know why. It was pretty weak though since if they (hopefully) thought about it they'd figure out what was going on. Don't feel very strongly about it and it was a pretty weak lean.

@yzguy @istr any opinions/thoughts?

istr:
For what it is be worth: I made a design decision a long time ago to _never_ use plural forms in variable names, whatsoever, for anything I start. You almost always come to a point where you want to turn a key/variable that once referred to a single value to an array or a map, or vice versa. This leads to problems/considerations like this. If you stick to "no plurals" from the start, you will have consistency from the beginning to the end of your project's lifecycle.

And it is semantically correct in most cases, at least for arrays / lists.

For example, read out loud `value[2] = "foo"`. This could be read 'set the second value to the string "foo"' for 1-based indices or 'set the value at index 2 to the string "foo" for 0-based indices. Much harder to read the same fluently for "values".

When refactoring existing stuff, I also always refactor towards "no plurals", keeping the old plural form for a while and then eventually dropping it.

istr:
As an addition: Note that the "no plurals" principle is just a subset of the more general "no type information in variable or key names" rule. Plurals are an indicator that the variables/keys are _not a scalar_, so that's a kind of type information - which is a more general design problem, if you ask me.

ross:
OK. Will pick this back up soon and rework things to treat `values` and `value` identically/indifferently. Will call out the potential for (problematic) values to be fixed on the first sync post upgrade in the CHANGELOG.

ross:
Started in on this, but ran into hiccups with the PTR type. IIRC it originally only supported a single value and was later updated to support values. Seems like there's something funky in how that happened that's breaking trying to auto-handle value/values. Will pick it back up and dig in further when I get a chance.

ross:
Updates made, `value` and `values` can now be used interchangeably and the CHANGELOG has info about the possibility of a fixing change post-update.

The biggest diff here is that all of the value type `validate` methods used to have to handle data being a single value or a list and they no longer do, it'll always be a list now.

viranch (approved):
Code wise LGTM!

istr:
> Updates made, value and values can now be used interchangeably

This is a good start. The implementation looks good.

istr (approved):
Looks good to me.
Randomly changed `values` with `value` and works as expected: no changes.

PR #1132, opened 2024-01-15
Add filters for checking record values

Nullreff:
This adds two new filters, `ValueAllowlistFilter` and `ValueRejectlistFilter` that allow checking the value(s) of records to include or exclude, similar to the name filters that already exist.  This was prompted by needing to filter CNAME records that point to a specific domain but is general enough to be used for any record.  Documentation has been updated, tests added and has full coverage.

### DIFF
--- a/octodns/processor/filter.py
+++ b/octodns/processor/filter.py
@@ -232,9 +232,9 @@ class _ValueBaseFilter(_FilterProcessor):
         for record in zone.records:
             values = []
             if hasattr(record, 'values'):
-                values = [str(value) for value in record.values]
+                values = [value.rdata_text for value in record.values]
             else:
-                values = [str(record.value)]
+                values = [record.value.rdata_text]
 
             if any(value in self.exact for value in values):
                 self.matches(zone, record)
@@ -280,9 +280,6 @@ class ValueAllowlistFilter(_ValueBaseFilter, AllowsMixin):
           - route53
     '''
 
-    def __init__(self, name, allowlist):
-        super().__init__(name, allowlist)
-
 
 class ValueRejectlistFilter(_ValueBaseFilter, RejectsMixin):
     '''Reject managing records with names that match the provider patterns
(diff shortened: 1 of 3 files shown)


octodns/processor/filter.py
```diff
@@ -215,6 +215,111 @@ def __init__(self, name, rejectlist):
...
+        sources:
+          - config
+        processors:
+          - only-these
+        targets:
+          - route53
+    '''
+
+    def __init__(self, name, allowlist):
+        super().__init__(name, allowlist)
```
ross, line 284: Don't believe this is necessary, should be fine to allow the parent to handle things. 

```suggestion
```

octodns/processor/filter.py
```diff
@@ -215,6 +215,111 @@ def __init__(self, name, rejectlist):
...
+        sources:
+          - config
+        processors:
+          - not-these
+        targets:
+          - route53
+    '''
+
+    def __init__(self, name, rejectlist):
+        super().__init__(name, rejectlist)
```
ross, line 320: Don't believe this is necessary, should be fine to allow the parent to handle things. 

```suggestion
```

octodns/processor/filter.py
```diff
@@ -215,6 +215,111 @@ def __init__(self, name, rejectlist):
...
+            else:
+                exact.add(pattern)
+        self.exact = exact
+        self.regex = regex
+
+    def _process(self, zone, *args, **kwargs):
+        for record in zone.records:
+            values = []
+            if hasattr(record, 'values'):
+                values = [str(value) for value in record.values]
```
ross, line 235: It seems a bit funky to string-ify complex values here, e.g. a `SRV` or something like that and it's not 100% clear what this would do. After looking at the code it seems like most (maybe all) the value types have `__repr__` methods that this would trigger and thus this would hit those. 

It may be preferable to be explicit here and use the `rdata_text` property/method and call out that's what happens in the documentation. That way people could write matchers (regex or exact) to match on them and we'd be explicit in what's going to be used for the matching. 

https://github.com/octodns/octodns/blob/30f8b4f5f5e9b613c1282615d645e0e2b3f71015/octodns/record/srv.py#L137-L138

octodns/processor/filter.py
Nullreff: Copied these from `NameAllowlistFilter` and `NameRejectlistFilter`, might want to remove them there too.

ross (approved):

ross:
Thanks

Issue #1130, opened 2024-01-07
CloudFlare Specific Proxied Type / Value


mhosker:
Hi,

I am currently looking to implement OctoDNS using Cloudflare & NS1, however a number of my records are required to be proxied via CloudFlare even when the DNS query is served by NS1.

I can make this work using a CloudFlare partial CNAME zone, where proxied records that exist in NS1 can be created as CNAMEs pointing back to Cloudflare e.g `www.example.com.cdn.cloudflare.net.`.

However, herein lies my problem... without using two zone files (one for NS1 & one for Cloudflare) is it possible to either specify a per provider value or add an override for CloudFlare specifically. Something like...

```
www:
  octodns:
    cloudflare:
      proxied: true
      type: A
      value: 8.8.8.8
  ttl: 300
  type: CNAME
  value: www.example.com.cdn.cloudflare.net.
```
Where `8.8.8.8` is the value of your secret origin server sitting behind Cloudflare.

I think this feature would be really beneficial (if it doesn't already exist and I have missed it!) as in a lot of cases the reason for proxying behind Cloudflare is to mask the origin IP, so I can see few use cases where you would have split DNS setup with Cloudflare returning a proxied record some of the time and the other provider returning the origin IP the rest of the time.

Additionally, if I were to use two non Cloudflare DNS providers and use Cloudflare as a CDN only, I would run into the same issue whereby the DNS providers would need to return the `.cdn.cloudflare.net.` CNAME record, however Cloudflare would need the true origin A / AAAA / CNAME configured.

Hopefully this makes sense, at the moment its making my deployment a bit tricky as I both don't want duplicate zone files to maintain and additional config OR my origin server IP exposed for non Cloudflare queries (plus it wouldn't work as my origins are all locked down to Cloudflare IPs only)!

As an aside, I love the OctoDNS project and have it successfully working with another zone that is solely Cloudflare. I plan to write some blog posts / produce a video on my rather unorthodox method of running OctoDNS using an Azure DevOps pipeline!

Thanks in advance! :)

ross:
> However, herein lies my problem... without using two zone files (one for NS1 & one for Cloudflare) is it possible to either specify a per provider value or add an override for CloudFlare specifically. Something like...

There's no built in way to accomplish this out of the box as it's a fairly unique setup, at least I haven't run across anyone doing it over the past ~6 years.

I can think of two main routes to manage it w/o having to completely duplicate record data:

[...]

PR #1133, opened 2024-01-15
Read only type

ross:
Over in https://github.com/octodns/octodns/issues/1130#issuecomment-1892721006 `_type` was being set with the expectation that it would alter the type of the record object, but it will not and it could potentially mess things up by changing the class-static `_type` vars on things. This PR adds a `readonly` decorator` and uses it on the `_type` (class) property for all of the record subtypes. Any 3rd party types would be free to start using the readonly bit once they require a new enough octoDNS, but it shouldn't hurt much of anything if they don't.

/cc https://github.com/octodns/octodns/issues/1130#issuecomment-1892721006 which lead to this PR

### DIFF
--- a/octodns/record/a.py
+++ b/octodns/record/a.py
@@ -4,7 +4,7 @@
 
 from ipaddress import IPv4Address as _IPv4Address
 
-from .base import Record
+from .base import Record, readonly
 from .dynamic import _DynamicMixin
 from .geo import _GeoMixin
 from .ip import _IpValue
@@ -19,7 +19,7 @@ Ipv4Address = Ipv4Value
 
 
 class ARecord(_DynamicMixin, _GeoMixin, Record):
-    _type = 'A'
+    _type = readonly('A')
     _value_type = Ipv4Value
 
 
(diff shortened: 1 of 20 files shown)


viranch (approved):
Isn't there a more native python way to make a class variable static/read-only? If not (or yes but not viable), this implementation LGTM.

ross:
> Isn't there a more native python way to make a class variable static/read-only? If not (or yes but not viable), this implementation LGTM.

I couldn't find a way to do it for class-static/method, only for object attributes. It looked like it worked using both `@classmethod` and `@property` 3.9 <= python <= 3.11, but was deprecated/declared unsupported 

* https://docs.python.org/3.11/library/functions.html#classmethod (see last point about the deprecation) 
* https://stackoverflow.com/q/76249636 (talks about the "right" way to do it, consensus seems to be custom "property")
* https://stackoverflow.com/a/1800999 (result that lead me through a rabbit hole of answers)

It would be possible to do a more traditional decorator here, but since all the existing stuff was just assigning to variables this getter class was slightly simpler/easier to understand.

ross:
After sitting on this for a couple days I think it's not really worth it. The problem it solves has only come up once or twice and it's a non-obvious/non-trivial so going to close.

PR #1135, opened 2024-01-16
Fall back to repr() in value filters and keep list names

Nullreff:
Follow up to #1132

I figured out why the seemly redundant constructors are included in filters.  The yaml files use `allowlist` and `rejectlist`, without the constructor it wants to use `_list` instead.  This doesn't match the behavior of other filters and would break backwards compatibility if changed.

Also found that `URLFWD` and a few other records don't actually have an `rdata_text` field and I couldn't find anything in the DNS spec about how those should be formatted.  Changed it to fall back to calling `repr()` if it can't find `rdata_text`, added a test case and updated the documentation.

### DIFF
--- a/octodns/processor/filter.py
+++ b/octodns/processor/filter.py
@@ -280,6 +280,9 @@ class ValueAllowlistFilter(_ValueBaseFilter, AllowsMixin):
           - route53
     '''
 
+    def __init__(self, name, allowlist):
+        super().__init__(name, allowlist)
+
 
 class ValueRejectlistFilter(_ValueBaseFilter, RejectsMixin):
     '''Reject managing records with names that match the provider patterns
@@ -312,6 +315,9 @@ class ValueRejectlistFilter(_ValueBaseFilter, RejectsMixin):
         targets:
           - route53
     '''
+
+    def __init__(self, name, rejectlist):
+        super().__init__(name, rejectlist)
 
 
 class _NetworkValueBaseFilter(BaseProcessor):


ross:
> I figured out why the seemly redundant constructors are included in filters. The yaml files use `allowlist` and `rejectlist`, without the constructor it wants to use `_list` instead. This doesn't match the behavior of other filters and would break backwards compatibility if changed.

Sorry about that. Totally my bad I should have remembered that from the other ones. 

> Also found that `URLFWD` and a few other records don't actually have an `rdata_text` field and I couldn't find anything in the DNS spec about how those should be formatted. Changed it to fall back to calling `repr()` if it can't find `rdata_text`, added a test case and updated the documentation.

Hrm. Let me look into./think about this. I thought everything implemented the rdata stuff, but maybe my recollection is off.

ross:
> Hrm. Let me look into./think about this. I thought everything implemented the rdata stuff, but maybe my recollection is off.

Looks like all of the base/rfc types (that bind supports) have `rdata_*`, but some of the ancillary types like `URLFWD` don't. 

I think the right fix here is to add support to those since it'll work here and for bind & tinydns. Absent a spec we can just make something up/figure out what would make sense. I'll start in on a PR for that in the next couple of days.

ross:
> I think the right fix here is to add support to those since it'll work here and for bind & tinydns

Finally got a chance to sit down in front of a computer for long enough to look into this. Implementation over in https://github.com/octodns/octodns/pull/1137, `UrlfwdRecord` was the only core record type missing the methods and there's now a unit test making sure that everything has them going forward.

I think this is good to go once things are reverted to just use `rdata_text` which should be safer to rely on.

Nullreff:
I removed the commit that falls back to `repr()`.  Also tested against my own files and everything appears to be working.

ross (approved):

PR #1139, opened 2024-02-11
Record.octodns formalized, Record._octodns deprecated

ross:
Both point to the same thing via properties. `_octodns` is in use in quite a lot of places and it deserves to be official API and not private `_`.


ross (commented):
/ci modules bump

Issue #1138, opened 2024-02-09
Arguments passed to the provider


avachette:
Hi @ross,

I wanted to know if arguments passed to the provider and especially the one with environment variable can be checked at provider level, the following code portion of code seems related to my issue:
- https://github.com/octodns/octodns/blob/main/octodns/manager.py#L380

In long term I want to add Hashicorp vault support and in many case environment variable needs to be filled when secret are retrieved using hvac python implementation.

Here is my current implementation:
```
    def _build_kwargs(self, source):
        # Build up the arguments we need to pass to the provider
        kwargs = {}
        for k, v in source.items():
            kwargs[k] = v

        return kwargs
```

Everything else is done at provider level for PowerDNS as an example:
```
#
#
#

import logging
import os
import hvac
from operator import itemgetter

from requests import HTTPError, Session

from octodns import __VERSION__ as octodns_version
from octodns.provider import ProviderException
from octodns.provider.base import BaseProvider
from octodns.record import Record

from .record import PowerDnsLuaRecord

__VERSION__ = '0.0.5'


def _escape_unescaped_semicolons(value):
    pieces = value.split(';')
    if len(pieces) == 1:
        return value
    last = pieces.pop()
    joined = ';'.join([p if p and p[-1] == '\\' else f'{p}\\' for p in pieces])
    ret = f'{joined};{last}'
    return ret


class PowerDnsBaseProvider(BaseProvider):
    SUPPORTS_GEO = False
    SUPPORTS_DYNAMIC = False
    SUPPORTS_ROOT_NS = True
    SUPPORTS = set(
        (
            'A',
            'AAAA',
            'ALIAS',
            'CAA',
            'CNAME',
            'LOC',
            'MX',
            'NAPTR',
            'NS',
            'PTR',
            'SPF',
            'SSHFP',
            'SRV',
            'TLSA',
            'TXT',
            PowerDnsLuaRecord._type,
        )
    )
    TIMEOUT = 5

    POWERDNS_MODES_OF_OPERATION = {
        'native',
        'primary',
        'secondary',
        'master',
        'slave',
    }
    POWERDNS_LEGACY_MODES_OF_OPERATION = {'native', 'master', 'slave'}

    def __init__(
        self,
        id,
        host,
        api_key,
        port=8081,
        scheme="http",
        ssl_verify=True,
        timeout=TIMEOUT,
        soa_edit_api='default',
        mode_of_operation='master',
        vault_enabled=False,
        vault_address=None,
[...]

PR #1140, opened 2024-02-11
Secrets handlers support

ross:
Add functionality that enables configurable secrets sources, with a hard-coded `env` that provides the existing environmental variable support.

/cc Fixes https://github.com/octodns/octodns/issues/1138


ross (commented):
/ci modules bump

ross:
Decided to move forward with this as I though of a really good option for another source of secrets, `keyring`. Will put together an external module that provides the functionality.

ross (commented):
/ci modules bump

ross:
🤞 this is a fairly safe change. The only bit of "production" code it touches is the env variable lookup code and it should be in line with the previous setups and is tested fairly well. Good news is that problems should result in failures to configure things and bail out early more than invalid data.

PR #1141, opened 2024-02-26
Version 1.5.0 changelog update & version bump

ross:
## v1.5.0 - 2024-02-26 - Checksums, nested expansion, & flexable values

* Beta support for Manager.enable_checksum and octodns-sync --checksum Allows a
  safer plan & apply workflow where the apply only moves forward if the apply
  phase plan exactly matches the previous round's planning.
* Fix for bug in MetaProcessor _up_to_date check that was failing when there was
  a plan with a single change type with a single value, e.g. CNAME.
* Support added for config env variable expansion on nested levels, not just
  top-level provider/processor keys
* _ChunkedValue ASCII validation added, SPF & TXT
* Re-work value/values handling to always try and do the "right" thing based on
  the content, so both singular values and lists will be handled identically
  regardless of whether the key is value or values. This may result in
  changes/fixes on the first sync after updating IFF you currently have
  `values: a-single-thing`, which would have previously been pushed up as bunch
  of single character values.
* Record.octodns added to formally make that data "API" and Record._octodns is
  deprecated. The latter has been converted to properties that return the
  former and emit deprecation warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v1.?.0 - 2023-??-?? -
+## v1.5.0 - 2024-02-26 - Checksums, nested expansion, & flexable values
 
 * Beta support for Manager.enable_checksum and octodns-sync --checksum Allows a
   safer plan & apply workflow where the apply only moves forward if the apply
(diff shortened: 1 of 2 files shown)


ross (commented):
/ci modules bump

Issue #1148, opened 2024-03-07
Config values set using environment variables are casted to a float


wjgauthier:
After upgrading to OctoDNS version 1.5.0, I started getting syncing errors with the PowerDNS provider. I get an error like this when I run octodns-sync:
`urllib3.exceptions.LocationParseError: Failed to parse: http://127.0.0.1:10000.0/api/v1/servers/localhost/zones/example.com`

I set the value of the PowerDNS provider options using environment variables like this: 
```
providers:
  powerdns:
    api_key: env/POWERDNS_API_KEY
    class: octodns_powerdns.PowerDnsProvider
    host: env/POWERDNS_HOST
    port: env/POWERDNS_PORT
    scheme: env/POWERDNS_SCHEME
```

Based on the error message above, it appears that the port is converted to a float. The breaking change was introduced in commit 4c09f4b, where the value is casted to a float. Changing this to an int fixes the problem.

https://github.com/octodns/octodns/blob/3acfd174b61ff5352294a0fbb5f9fb706d76bf14/octodns/manager.py#L393

ross:
> Based on the error message above, it appears that the port is converted to a float. The breaking change was introduced in commit [4c09f4b](https://github.com/octodns/octodns/commit/4c09f4bc71df659e698eff7ea95b7056c012a35f), where the value is casted to a float. Changing this to an int fixes the problem.

Thanks for the digging/link, super helpful. 

I could have sworn that change happened in 1.4.0, but looks like it landed the day after the release and then was updated further after that before 1.5.0.

I'd also kind of consider it a bug in octodns-powerdns that it just string-ifies the value w/o a format specifier 

https://github.com/octodns/octodns-powerdns/blob/80af99402747b63f07308c14f821edd4ea4dfb5f/octodns_powerdns/__init__.py#L132-L135

I'll probably get a PR in for that too, but do still consider this a regression so will fix it on both ends. I will try and sit down either this evening or tomorrow morning to write them up, looking more likely tomorrow.

ross:
https://github.com/octodns/octodns/pull/1151 / https://pypi.org/project/octodns/1.5.1/ has the fix

PR #1150, opened 2024-03-08
_build_kwargs fix, don't convert int-y things to floats

ross:
Only try converting things to float when there's a `.` in them to avoid int-y things having being a float (even if it's with .0)

/cc Fixes https://github.com/octodns/octodns/issues/1148#issuecomment-1984919619 @wjgauthier
/cc https://github.com/octodns/octodns/pull/1118 which introduced the issue
/cc https://github.com/octodns/octodns-powerdns/pull/63 which was turned up b/c of this


ross (commented):
/ci modules build

Issue #1149, opened 2024-03-08
Account float conversion breaks requests


LuminatiHD:
New in version 1.5.0 , the function `_build_kwargs` of the class `Manager` in the file `octodns/manager.py` converts the attributes to floats before passing them on. So later on in the Module octodns-simple, an instance of the class `DnsimpleClient` receives its account number as a float. So instead of its account number being, say, 12345, it's now 12345.0. Which means, its requests look like:
`https://api.dnsimple.com/v2/12345.0/endpoint` 
instead of
`https://api.dnsimple.com/v2/12345/endpoint`,
which obviously doesn't work.

ross:
Duplicate of https://github.com/octodns/octodns/issues/1148. fix in progress in https://github.com/octodns/octodns/pull/1150 & https://github.com/octodns/octodns-powerdns/pull/63.

Closing in favor of https://github.com/octodns/octodns/issues/1148, any further discussion can happen there.

LuminatiHD:
1.5.1 solved this issue, but 1.6.0 broke it again. While there is a fix now for octodns-dnsimple, which automatically converts it to an int, octodns still passes the account number along as a float.

ross:
Hrm. Nothing changed about the float/int logic between those two versions. 

https://github.com/octodns/octodns/blob/v1.5.1/octodns/manager.py#L390-L399
https://github.com/octodns/octodns/blob/v1.6.0/octodns/manager.py#L438-L447

The only thing that changed was a check to only apply that if the type was a string coming in, since the new secrets handlers can return things already typed. https://github.com/octodns/octodns/blob/v1.6.0/octodns/manager.py#L437

I'll see if I can reproduce it, but at the moment not seeing how.

ross:
I'm not able to reproduce this. Anything that comes from an env variable should be a string from `os.environ` and the code in both 1.5.1 and 1.6.0 should check for a `.` and if one is not found use `int`. 

Do you have any details on now to reproduce it?

PR #1154, opened 2024-03-17
Fix env var int handling regression

ross:
The `Manager` version should have gone away once the secret handler was extracted.

/cc Fixes https://github.com/octodns/octodns/issues/1149


ross (commented):
/ci modules build

PR #1155, opened 2024-03-17
Version 1.6.1 bump & changelog update

ross:
## v1.6.1 - 2024-03-17 - Didn't we do this already

* Fix env var type handling that was previously fixed in 1.5.1 and then
  regressed in 1.6.0.


ross (commented):
/ci modules bump

PR #1159, opened 2024-04-10
Add Zone Specific Threshold Config

cholick:
As users of OctoDNS, we have quite a complex configuration that's being both partially populated by automation. We'd like to add an additional level of protection for zones that we do not expect to change, so that folks deploying OctoDNS have an additional guard rail in place.

This PR adds a Zone specific threshold that:
* Has no defaults
* Takes precedence over the provider defaults or explicitly specified provider configuration

In building this PR, the one thing that seemed strange to me was having to guard against `existing` not being `None`, since the `Manager` looks like it would be constructing any existing `Zone` from configuration. As in this line:
```
if existing and existing.update_pcent_threshold is not None:
```

I expected to be simply
```
if existing.update_pcent_threshold is not None:
```

So I might be missing some flows about how `Plan` is being used.

### DIFF
--- a/octodns/provider/plan.py
+++ b/octodns/provider/plan.py
@@ -57,15 +57,16 @@ class Plan(object):
         # them and/or is as safe as possible.
         self.changes = sorted(changes)
         self.exists = exists
-        self.update_pcent_threshold = update_pcent_threshold
-        self.delete_pcent_threshold = delete_pcent_threshold
 
         # Zone thresholds take precedence over provider
         if existing and existing.update_pcent_threshold is not None:
             self.update_pcent_threshold = existing.update_pcent_threshold
+        else:
+            self.update_pcent_threshold = update_pcent_threshold
         if existing and existing.delete_pcent_threshold is not None:
             self.delete_pcent_threshold = existing.delete_pcent_threshold
-
+        else:
+            self.delete_pcent_threshold = delete_pcent_threshold
         change_counts = {'Create': 0, 'Delete': 0, 'Update': 0}
         for change in changes:
             change_counts[change.__class__.__name__] += 1
--- a/octodns/provider/plan.py
+++ b/octodns/provider/plan.py
@@ -60,6 +60,12 @@ class Plan(object):
         self.update_pcent_threshold = update_pcent_threshold
         self.delete_pcent_threshold = delete_pcent_threshold
 
+        # Zone thresholds take precedence over provider
+        if existing and existing.update_pcent_threshold is not None:
+            self.update_pcent_threshold = existing.update_pcent_threshold
+        if existing and existing.delete_pcent_threshold is not None:
+            self.delete_pcent_threshold = existing.delete_pcent_threshold
+
         change_counts = {'Create': 0, 'Delete': 0, 'Update': 0}
         for change in changes:
             change_counts[change.__class__.__name__] += 1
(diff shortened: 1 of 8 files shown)


octodns/provider/plan.py
```diff
@@ -60,6 +60,12 @@ def __init__(
         self.update_pcent_threshold = update_pcent_threshold
         self.delete_pcent_threshold = delete_pcent_threshold
 
+        # Zone thresholds take precedence over provider
+        if existing and existing.update_pcent_threshold is not None:
+            self.update_pcent_threshold = existing.update_pcent_threshold
+        if existing and existing.delete_pcent_threshold is not None:
+            self.delete_pcent_threshold = existing.delete_pcent_threshold
+
```
ross, line 68: I think combining the two ways the pcents get set into one section makes the logic a slightly clearer, comment explained it well, this is just belt & suspenders:

```suggestion

        # Zone thresholds take precedence over provider
        if existing and existing.update_pcent_threshold is not None:
            self.update_pcent_threshold = existing.update_pcent_threshold
        else:
            self.update_pcent_threshold = update_pcent_threshold
        if existing and existing.delete_pcent_threshold is not None:
            self.delete_pcent_threshold = existing.delete_pcent_threshold
        else:
            self.delete_pcent_threshold = delete_pcent_threshold
```

tests/test_octodns_manager.py
```diff
@@ -1294,6 +1294,23 @@ def test_config_secret_handlers(self):
...
+
+            self.assertEqual(0.2, zone.update_pcent_threshold)
+            self.assertEqual(0.1, zone.delete_pcent_threshold)
+
+            # subzone has different threshold
+            subzone = manager.get_zone('subzone.unit.tests.')
+
+            self.assertEqual(0.02, subzone.update_pcent_threshold)
+            self.assertEqual(0.01, subzone.delete_pcent_threshold)
+
```
ross, line 1313: Probably good to get a test in here to make sure `zone.*pcent*` are `None` for things where it's not specified. Looks like that should be as simple as adding a 3rd zone to `zone-threshold.yaml` that doesn't specify a value.

ross (commented):
Looks good. Couple minor things inline.

I did wonder about how this works with `alias` zones. They not super common and just allow a zone to be populated with the record data from another. As is the alias zone doesn't copy over the `pcent` information from where it's pulling it's data. I think I lean toward not copying over the `pcnet` info as it doesn't seem super obvious that it would/should and people can always define it on the alias directly if they're already going to be putting it all over the other/origin zones.

> In building this PR, the one thing that seemed strange to me was having to guard against existing not being None, since the Manager looks like it would be constructing any existing Zone from configuration. As in this line:

Without digging in to see for sure I imagine this is just test cases taking short cuts and passing in `None` for `existing` when they "know" it's not needed.

octodns/provider/plan.py
cholick: Good suggestion, that reads better.

tests/test_octodns_manager.py
cholick: Makes sense; pushed that additional test.

ross (approved):

ross:
Thanks!

cholick:
@ross thanks for the quick review and merge!

Issue #1158, opened 2024-04-03
Feature request: Make it possible to choose record for auto ARPA


wjgauthier:
In a setup where multiple A/AAAA records point to the same IP, it would be great if you can choose which record should take preference when using the auto_arpa provider, for example by having an optional use_for_ptr option on those types of records.

Currently, I set `populate_should_replace` to true on the auto_arpa provider and `strict_supports` to false on the PowerDNS provider, which causes the first match to be used as the value of the PTR record. This is not always what I want.

If you think this feature makes sense, I can probably help out with a pull request.

ross:
Something like this probably make sense. It may be preferable to have it be a priority rather than a flag to use this one, that way if the one with the flag set goes away the behavior would still make sense/be predictable. 

I think the main bits that would have to change are around here:

https://github.com/octodns/octodns/blob/e594f227c1d376d42d1b16d4f976dd7d575b0986/octodns/processor/arpa.py#L37-L40

It'd probably need something like:

```python
auto_arpa_priority = record.octodns.get('auto_arpa_priority', 0)
```

And then rather than have `self._records` be `set`s then could be a `list`s that a tuple is appended to:

```python
                    self._records[f'{ptr}.'].append((auto_arpa_priority, record.fqdn))
```

The `sorted` down here would do the right thing 

https://github.com/octodns/octodns/blob/e594f227c1d376d42d1b16d4f976dd7d575b0986/octodns/processor/arpa.py#L59

It'd just need to turn the list of tuples into a straight list of values

```python
fqdns = [d[1] for d in fqdns]
```

I guess the last question would be how it knows when to use just one (or more.) Maybe another parameter to `AutoArpa` tells it how many of `fqdn`s to use defaulting to a large number, `99999` or whatever.

Anyway, that's my 5m quick skim thoughts. I could be flipping how it all works in which case the detail of what I say above won't work, but I think the general ideas of priority and N should be doable regardless. 

Happy to 👁️ a PR. If not I'll get thrown onto the TODO stack and looked at sooner or later.

ross:
Thanks

PR #1161, opened 2024-04-17
Add a priority option to AutoArpa

wjgauthier:
I have made the changes we talked about in https://github.com/octodns/octodns/issues/1158. I hope this makes sense. Just two things:
- I had to change all the `assertEqual` statements in `test_octodns_processor_arpa.py` because `process_source_zone` now returns a priority
- Because `self._records` is a list now, it can contain duplicate values for the same IP, so I convert it to a set to ensure unique values (the test_dynamic test would fail because of duplicate values)


octodns/processor/arpa.py
```diff
@@ -37,7 +39,10 @@ def process_source_zone(self, desired, sources):
 
                 for ip in ips:
                     ptr = ip_address(ip).reverse_pointer
-                    self._records[f'{ptr}.'].add(record.fqdn)
+                    auto_arpa_priority = record.octodns.get('auto_arpa_priority', 999)
+                    self._records[f'{ptr}.'].append((auto_arpa_priority, record.fqdn))
+                    unique_list = list(set(self._records[f'{ptr}.']))
```
ross, line 44: I believe this will only keep things unique when they have the same priority since tuples of (priority, fqdn) are being placed into the list. so if the same fqdn showed up 

It may be preferable to just `append` everything here and then deal with making things unique down below when it's pulling the fqdn's out of the tuples. That'd probably result in a slightly more predictable behavior where the lowest priority seen for a fqdn would win.

Another option that comes to mind would be a `dict` with the fqdn a the key and priority as the value. That would result in last one winning though which is at least consistent, but probably harder to predict. Sorting it would require a couple of extra steps, something like `.items()`, flipping the k & v, and then sorting.

ross (commented):
Mostly there. One comment/discussion bit inline. 

Thanks.

octodns/processor/arpa.py
wjgauthier: Yes, I think you are right. This will only keep things unique if they have the same priority. I have moved it to populate instead.

octodns/processor/arpa.py
```diff
@@ -56,7 +59,11 @@ def populate(self, zone, target=False, lenient=False):
         for arpa, fqdns in self._records.items():
             if arpa.endswith(f'.{zone_name}'):
                 name = arpa[:-n]
+                fqdns = list(set(fqdns))
```
ross, line 62: I think this will still have the problem of including the priority in the uniqueness. I believe the uniq-ifying would need to happen after line 64 where the `d[1]` bit is extracted, but that would potentially lose the sort order so this one is kind of interesting. 

May be easiest/cleanest to switch lines 63-65 to a more explicit setup like:

```python
seen = set()
ordered = sorted(fqdns)
fqdns = []
for _, fqdn in ordered:
    if fqdn in seen:
        continue
    fqdns.append(fqdn)
    seen.add(fqdn)
    if len(seen) >= self.max_auto_arpa:
        break
```

The trickiness of this definitely points towards the need for some specific testing around the behavior with duplicates fqdns & priorities, fqdns w/different priorities, etc. 

Might make sense to pull the logic out into a helper function:

```python
# Note: this takes a list of (priority, fqdn) tuples and returns the ordered and uniqified list of fqdns.
fqdns = self._order_and_unique_fqdns(fqdns, self.max_auto_arpa)
```

It would then be easy to call this method directly on an `AutoArpa` instance with all the various permutations without having to deal with creating records and all of the unrelated scaffolding.

ross (commented):
Comment/suggestion inline.

octodns/processor/arpa.py
wjgauthier: Great, thank you. I have added the suggestion.

ross (approved):

ross:
googlecloud failure is unrelated, just dep problems in all of their modules.

PR #1162, opened 2024-04-18
Build wheel and install that for cibuild-module

ross:
Pip doesn't seems to be deprecating the ability to work when things are installed into it with `python setup.py install`. It currently prints deprecation warnings and in some cases (google cloud) fails.

/cc https://github.com/octodns/octodns-googlecloud/pull/48


ross (commented):
/ci modules build

PR #1165, opened 2024-04-26
AutoArpa: Add an optional inherit_ttl option

wjgauthier:
Hi Ross,

It would be great with some more flexibility for setting the TTL value on AutoArpa PTR records. Right now, it's only possible to set a default value on the auto_arpa provider configuration, which is then used by all records. I would like to be able to set a different value on some of my records.

This pull request adds an optional option, `inherit_ttl`, which takes the TTL value of the corresponding A/AAAA and uses that instead of the default value.

Another possibility could also be to add an `auto_arpa_ttl` option directly on the A/AAAA records, like what we did with `auto_arpa_priority`. But I think it would make more sense to use the same value as the A/AAAA records for the most part.

I hope this feature request makes sense.

Thanks,
William

### DIFF
--- a/octodns/processor/arpa.py
+++ b/octodns/processor/arpa.py
@@ -12,7 +12,12 @@ from .base import BaseProcessor
 
 class AutoArpa(BaseProcessor):
     def __init__(
-        self, name, ttl=3600, populate_should_replace=False, max_auto_arpa=999, inherit_ttl=False
+        self,
+        name,
+        ttl=3600,
+        populate_should_replace=False,
+        max_auto_arpa=999,
+        inherit_ttl=False,
     ):
         super().__init__(name)
         self.log = getLogger(f'AutoArpa[{name}]')
@@ -61,10 +66,10 @@ class AutoArpa(BaseProcessor):
         # order the fqdns making a copy so we can reset the list below
         ordered = sorted(fqdns)
         fqdns = []
-        for _, _, fqdn in ordered:
+        for _, record_ttl, fqdn in ordered:
             if fqdn in seen:
                 continue
-            fqdns.append(fqdn)
+            fqdns.append((record_ttl, fqdn))
             seen.add(fqdn)
             if len(seen) >= max_auto_arpa:
                 break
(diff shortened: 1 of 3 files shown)


octodns/processor/arpa.py
```diff
@@ -79,12 +85,13 @@ def populate(self, zone, target=False, lenient=False):
         for arpa, fqdns in self._records.items():
             if arpa.endswith(f'.{zone_name}'):
                 name = arpa[:-n]
+                _, record_ttl, _ = fqdns[0]
```
ross, line 88: If the ttls across duplicate fqdns don't match this will pick the first one it sees which may be a little confusing to users since they won't necessarily be able to predict what will get picked without knowing the order things end up in the list. Worst case it could even varry which one was first in the list from run to run and cause ttl churn.

octodns/processor/arpa.py
```diff
@@ -79,12 +85,13 @@ def populate(self, zone, target=False, lenient=False):
...
                 # Note: this takes a list of (priority, fqdn) tuples and returns the ordered and uniqified list of fqdns.
```
ross, line 89: It appears there's an extra bit in the tuple being passed now so the comment/doc needs to be updated:

```suggestion
                # Note: this takes a list of (priority, ttl, fqdn) tuples and returns the ordered and uniqified list of fqdns.
```

Related to the `[0]` comment above perhaps it makes sense to have `_order_and_unique_fqdns` pick the ttl and return a tuple of `(ttl, fqdn)` so that it can use the ttl of the post ordered "first" occurance of each fqdn.

Relatedly we'll want some tests around the behavior to make sure it works as defined/expected and doesn't regress in the future.

ross (commented):
Good idea. Mostly there, couple comments/discussion points inline.

octodns/processor/arpa.py
wjgauthier: Yes, true. I didn't think of the scenario where multiple PTR values exist. I use max_auto_arpa=1 in my setup. Would the other option (auto_arpa_ttl directly on the A/AAAA) make more sense then?

ross: Not opposed to a way to do an explicit `max_auto_arpa`, but i like the copy-over option as well. I'd probably use/prefer the copy option myself, but would accept a another PR with the explicit option if other people would use it.

octodns/processor/arpa.py
wjgauthier: I'm not sure I understand this correctly, but it now takes the TTL value of the first ordered FQDN (the one with the lowest priority). I have also added a test for this.

ross (approved):
Seems like you covered al the bits in my comment so we're good here.

Issue #1164, opened 2024-04-25
Domain duplication [route53 -> bind9]


paxlo:
Hi @ross. I am using the configuration from my [previous ticket](https://github.com/octodns/octodns/issues/1146) to synchronise a large number of records from route53 to bind9. The latest version of octodns 1.6.1 is installed from pip. The records synchronised successfully, but many of them contain a duplicated domain like example.com.example.com.

```
$ORIGIN .
$TTL 900        ; 15 minutes
example.internal     IN SOA  ns-5.example.internal. hostmaster.example.internal. (
                                4          ; serial
                                604800     ; refresh (1 week)
                                86400      ; retry (1 day)
                                2419200    ; expire (4 weeks)
                                604800     ; minimum (1 week)
                                )
$TTL 86400      ; 1 day
                        NS      ns-5.example.internal.
$ORIGIN example.internal.
$TTL 300        ; 5 minutes
api                     CNAME   k8s-kaasingr-ingressn-5849e410fe-f568f9ed4e6a7d81.elb.us-east-1.amazonaws.com
argocd                  CNAME   k8s-kaasingr-ingressn-b0b7ae0801-03420a89ebc14a6b.elb.us-east-1.amazonaws.com
auth                    CNAME   k8s-kaasingr-ingressn-b0b7ae0801-03420a89ebc14a6b.elb.us-east-1.amazonaws.com
$ORIGIN dalaran.eks.aws.example.internal.
hubble                  CNAME   k8s-kaasingr-ingressn-1da6582ece-2ab191a55f9cee42.elb.us-east-1.amazonaws.com.example.internal.
jaeger                  CNAME   k8s-kaasingr-ingressn-1da6582ece-2ab191a55f9cee42.elb.us-east-1.amazonaws.com.example.internal.
kiali                   CNAME   k8s-kaasingr-ingressn-1da6582ece-2ab191a55f9cee42.elb.us-east-1.amazonaws.com.example.internal.
policy-reporter         CNAME   k8s-kaasingr-ingressn-1da6582ece-2ab191a55f9cee42.elb.us-east-1.amazonaws.com.example.internal.
prometheus              CNAME   k8s-kaasingr-ingressn-1da6582ece-2ab191a55f9cee42.elb.us-east-1.amazonaws.com.example.internal.
$ORIGIN eks.aws.example.internal.
$TTL 60 ; 1 minute
[...]

PR #1167, opened 2024-04-26
Fix for EnsureTrailingDots reverting value types back to strings

ross:
The processor was previously altering the types of string-y values by assigning strings over them. Once it was a `str` it no longer supported the rdata methods that providers like bind utilize. This fixes that issue by preserving the value types.

/cc Fixes https://github.com/octodns/octodns/issues/1164 @paxlo


ross:
Verified this fixes the reproduction of https://github.com/octodns/octodns/issues/1164 in octodns-bind.

ross (commented):
/ci modules build

PR #1169, opened 2024-04-29
Version 1.7.0 bump & changelog update

ross:
## v1.7.0 - 2024-04-29 - All the knobs and dials

* Support for specifying per-zone change thresholds, to allow for zones
  where lots of changes are expected frequently to live along side zones
  where little or no churn is expected.
* AutoArpa gained support for prioritizing values
* Fix for EnsureTrailingDots reverting value types back to strings which then
  failed when rr methods were used on them (e.g. w/octodns-bind)


ross (commented):
/ci modules build

PR #1171, opened 2024-05-05
Fix CAA value handling to support tags

ross:
Fix CAA rdata parsing to allow values with tags/properties. 

/cc https://github.com/octodns/octodns-route53/pull/92 @jmittermair

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.?.? - 2024-??-?? - ???
+
+* Fix CAA rdata parsing to allow values with tags
+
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci modules build

ross:
Just did a quick scan of the other providers that deal with parsing CAA record rdata and afaict they're already correct, e.g. https://github.com/octodns/octodns-powerdns/blob/25a72bb4decb3ff135e237c67ec2f6167732951e/octodns_powerdns/__init__.py#L212

istr (approved):
@ross looks good to me.

PR #1173, opened 2024-05-10
Fix add_record() lenient check to allow managed sub-zone DS records

rasturic:
Like NS records, it's valid to have a sub-zone record name as zone being managed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## v1.?.? - 2024-??-?? - ???
 
+* Allow DS records to be specified for managed sub-zones, same as NS
 * Fix CAA rdata parsing to allow values with tags
 
 ## v1.7.0 - 2024-04-29 - All the knobs and dials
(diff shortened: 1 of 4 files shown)


ross (commented):
👍 to the change. Needs a bit of testing to go along with it. Looks like the existing tests of NS behavior are https://github.com/octodns/octodns/blob/fae406af212034b67d9d9bb21196e684ca4a9e15/tests/test_octodns_zone.py#L311-L337

rasturic:
> 👍 to the change. Needs a bit of testing to go along with it. Looks like the existing tests of NS behavior are
> 
> https://github.com/octodns/octodns/blob/fae406af212034b67d9d9bb21196e684ca4a9e15/tests/test_octodns_zone.py#L311-L337

I was wondering about that. Hopefully the ones I just pushed up are good.

ross (approved):
Testing looks good. Thanks!

PR #1176, opened 2024-05-30
Add support for SVCB and HTTPS records

pieterlexis-tomtom:
These records are defined in [RFC 9460](https://datatracker.ietf.org/doc/html/rfc9460) and are used for "Service Binding" and provide clients with all connection information for network services.

Modern browsers already query for these records and several larger websites have [adopted usage of HTTPS records](https://blog.apnic.net/2023/12/18/use-of-https-resource-records/) already.

I have added a few `XXX` comments that indicate some questions I had regarding the internal data structure.
But by-and-large, parsing and generating these records works well.

### DIFF
--- a/octodns/record/svcb.py
+++ b/octodns/record/svcb.py
@@ -29,18 +29,26 @@ def validate_svcparam_port(svcparamvalue):
     return reasons
 
 
+def validate_list(svcparamkey, svcparamvalue):
+    if not isinstance(svcparamvalue, list):
+        return [f'{svcparamkey} is not a list']
+    return list()
+
+
 def validate_svcparam_alpn(svcparamvalue):
-    reasons = []
-    alpns = svcparamvalue.split(',')
-    for alpn in alpns:
+    reasons = validate_list('alpn', svcparamvalue)
+    if len(reasons) != 0:
+        return reasons
+    for alpn in svcparamvalue:
         reasons += _ChunkedValue.validate(alpn, 'SVCB')
     return reasons
 
 
 def validate_svcparam_iphint(ip_version, svcparamvalue):
-    reasons = []
-    addresses = svcparamvalue.split(',')
-    for address in addresses:
+    reasons = validate_list(f'ipv{ip_version}hint', svcparamvalue)
+    if len(reasons) != 0:
+        return reasons
+    for address in svcparamvalue:
         try:
             if ip_version == 4:
                 IPv4Address(address)
@@ -48,23 +56,24 @@ def validate_svcparam_iphint(ip_version, svcparamvalue):
                 IPv6Address(address)
         except AddressValueError:
             reasons.append(
-                f'ip{ip_version}hint "{address}" is not a valid IPv{ip_version} address'
+                f'ipv{ip_version}hint "{address}" is not a valid IPv{ip_version} address'
             )
     return reasons
 
 
-def validate_svcparam_ip4hint(svcparamvalue):
+def validate_svcparam_ipv4hint(svcparamvalue):
     return validate_svcparam_iphint(4, svcparamvalue)
 
 
-def validate_svcparam_ip6hint(svcparamvalue):
+def validate_svcparam_ipv6hint(svcparamvalue):
     return validate_svcparam_iphint(6, svcparamvalue)
 
 
 def validate_svcparam_mandatory(svcparamvalue):
-    reasons = []
-    mandatories = svcparamvalue.split(',')
-    for mandatory in mandatories:
+    reasons = validate_list('mandatory', svcparamvalue)
+    if len(reasons) != 0:
+        return reasons
+    for mandatory in svcparamvalue:
         if (
             mandatory not in SUPPORTED_PARAMS.keys()
             and not mandatory.startswith('key')
@@ -92,14 +101,30 @@ def validate_svckey_number(paramkey):
     return []
 
 
+def parse_rdata_text_svcparamvalue_list(svcparamvalue):
+    return svcparamvalue.split(',')
+
+
 # cc https://datatracker.ietf.org/doc/html/rfc9460#keys
 SUPPORTED_PARAMS = {
     'no-default-alpn': {'has_arg': False},
-    'alpn': {'validate': validate_svcparam_alpn},
+    'alpn': {
+        'validate': validate_svcparam_alpn,
+        'parse_rdata_text': parse_rdata_text_svcparamvalue_list,
+    },
     'port': {'validate': validate_svcparam_port},
-    'ipv4hint': {'validate': validate_svcparam_ip4hint},
-    'ipv6hint': {'validate': validate_svcparam_ip6hint},
-    'mandatory': {'validate': validate_svcparam_mandatory},
+    'ipv4hint': {
+        'validate': validate_svcparam_ipv4hint,
+        'parse_rdata_text': parse_rdata_text_svcparamvalue_list,
+    },
+    'ipv6hint': {
+        'validate': validate_svcparam_ipv6hint,
+        'parse_rdata_text': parse_rdata_text_svcparamvalue_list,
+    },
+    'mandatory': {
+        'validate': validate_svcparam_mandatory,
+        'parse_rdata_text': parse_rdata_text_svcparamvalue_list,
+    },
     'ech': {'validate': validate_svcparam_ech},
 }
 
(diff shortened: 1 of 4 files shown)


ross:
Will probably take me a couple days to read up on the new record types and develop an opinion on them.

Looks like the `dig` on OSX is not new enough for the types, but you can use a dockerized version:

```console
coho:~ ross$ docker run --rm -ti toolbelt/dig +short https google.com.
1 . alpn="h2,h3"
coho:~ ross$ docker run --rm -ti toolbelt/dig +short https cloudflare.com.
1 . alpn="h3,h2" ipv4hint=104.16.132.229,104.16.133.229 ipv6hint=2606:4700::6810:84e5,2606:4700::6810:85e5
coho:~ ross$ docker run --rm -ti toolbelt/dig +short https facebook.com.
coho:~ ross$
```

Seems to be about 50/50 for the big "name brand" sites I've tested. 

Once I'm caught up I'll take a look at the code/XXXs

octodns/record/svcb.py
```diff
@@ -0,0 +1,251 @@
...
+}
+
+
+class SvcbValue(EqualityTupleMixin, dict):
+
+    @classmethod
+    def parse_rdata_text(cls, value):
+        try:
+            # XXX: these are called SvcPriority, TargetName, and SvcParams in RFC 9460 section 2.
+            #       Should we mirror these names, or are priority, target and params good enough?
```
ross, line 112: Generally we've matched the names used in RFCs, e.g. MX does have `priority` & `target`. IMO It's annoying that they're prefixed/compound, but we should probably continue to match them. Everything else in octoDNS "should" be doing so and anything that doesn't is a "bug"

octodns/record/svcb.py
```diff
@@ -0,0 +1,251 @@
...
+            # XXX: Should we split params into the specific ParamKeys and ParamValues?
+            (priority, target, *params) = value.split(' ')
```
ross, line 114: This one is interesting. Generally octoDNS has tended to leave "compound" params like this alone and just use the strings verbatim, but for the most part those fields didn't explicitly say they're a list of key & value pairs and here that is the case and there's even further limiting things to only appearing once and laying out the details of the k & v.

https://datatracker.ietf.org/doc/html/rfc9460#presentation

> Each SvcParamKey SHALL appear at most once in the SvcParams

https://datatracker.ietf.org/doc/html/rfc9460#keys

I'd lean toward parsing them and having them as a dict in octoDNS's model of things, and thus in the YAML. I think that'd mean this bit stays as-is and something down below near the `unquote(target)` would convert the list of strings into a dict (with `None` for the value of anything is valueless) unless I've missunderstood the details of the params. 

I guess there's a valid question of whether we further parse things like ipv4/6hint and turn the `,` separated IPs into lists.

octodns/record/svcb.py
```diff
@@ -0,0 +1,251 @@
...
+    try:
+        paramkeynum = int(paramkey[3:])
+        if 7 < paramkeynum > 65535:
+            return [f'SvcParam key "{paramkey}" has wrong key number']
+    except ValueError:
+        return [f'SvcParam key "{paramkey}" has wrong format']
+    return []
+
+
+SUPPORTED_PARAMS = {
```
ross, line 95: Would be good to link this to https://datatracker.ietf.org/doc/html/rfc9460#keys since that's the source of the keys. Al

```suggestion
# cc https://datatracker.ietf.org/doc/html/rfc9460#keys
SUPPORTED_PARAMS = {
```

tests/zones/unit.tests.tst
```diff
@@ -55,3 +55,14 @@ aaaa        600   IN  AAAA  2601:644:500:e210:62f8:1dff:feb8:947a
...
+; SVCB and HTTPS records
+svcb-alias 300 IN SVCB 0 alias-target.unit.test.
+
+svcb-service 300 IN SVCB 1 . ipv4hint=192.0.2.4 port=9000
+svcb-service 300 IN SVCB 2 svcb-target.unit.test. port=9001
+
+https-alias 300 IN HTTPS 0 alias-target.unit.test.
+
+https-service 300 IN HTTPS 1 . ipv4hint=192.0.2.4 port=9000
+https-service 300 IN HTTPS 2 svcb-target.unit.test. port=9001
```
ross, line 68: Looks like dnspython 2.1.0 added support for these record types https://dnspython.readthedocs.io/en/latest/whatsnew.html#id10 and we already require

https://github.com/octodns/octodns/blob/45753744a66982dd977de3bbe1ddceda07176e42/setup.py#L79

So this should always be supported.

ross (commented):
Overall solid & thorough PR. Made a pass through answering the XXX bits and a couple other minor things. 

Thanks.

octodns/record/svcb.py
pieterlexis-tomtom: There them is the question of having keys like this:

```
svcparams:
  alpn: h2,h3
  ipv4hint: 192.0.2.0
```

or with the actual titles:

```
svcparams:
  - svcparamkey: alpn
    svcparamvalue: h2,h3
  - svcparamkey: ipv4hint
    svcparamvalue: 192.0.2.1
```

I'm leaning towards the first  to make it easy on the users.

pieterlexis-tomtom: > I guess there's a valid question of whether we further parse things like ipv4/6hint and turn the , separated IPs into lists.

The real question is, what does octodns gain from that?

tests/zones/unit.tests.tst
pieterlexis-tomtom: What does this mean then, should these lines be removed?

octodns/record/svcb.py
pieterlexis-tomtom: alright, I'll update that. I wonder how big the mental strain on users will be though. But with having the yaml under version control, changes are very easy to see.

```yaml
svcparams:
  alpn:
    - h2
    - h3
  ipv4hint:
    - 192.0.2.0
```

Or would you want accept this as well:

```yaml
svcparams:
  alpn:
    - h2
    - h3
  ipv4hint: 192.0.2.0
```

ross: > But with having the yaml under version control, changes are very easy to see.

Yeah, that's a main argument for breaking things out.



> Or would you want accept this as well:

No strong opinion. If it's easy to support the single value option is probably nice to have and something people will accidentally do from time to time.

pieterlexis-tomtom: it might be nice to have, but I realize that would mean even more code to prevent users from not just stuffing a comma-separated string in the yaml. Explicitly requiring a list is cleaner.

pieterlexis-tomtom:
I think this is ready for review

ross (approved):

octodns/record/svcb.py
```diff
@@ -0,0 +1,318 @@
...
+                params[paramkey] = paramvalue[0]
+                if (
+                    SUPPORTED_PARAMS.get(paramkey, {}).get(
+                        'parse_rdata_text', None
+                    )
+                    is not None
+                ):
+                    params[paramkey] = SUPPORTED_PARAMS[paramkey][
+                        'parse_rdata_text'
+                    ](paramvalue[0])
```
ross, line 170: Screenshot 2024-06-05 at 2 33 12 PM

I think the failures have to do with a ~bug in older versions of python <-> coverage. I vagulely remembered running into something similar in the past and github's `involves:ross` search in the coverage repo turned it up:

https://github.com/nedbat/coveragepy/issues/1672

I believe the following will keep the same logic/intent, but hopefully break things up in a way that avoids the optimization/coverage problem. It should also avoid double looking things up in the dicts. 

```suggestion
                parse_rdata_text = SUPPORTED_PARAMS.get(paramkey, {}).get(
                    'parse_rdata_text', None
                )
                if parse_rdata_text is not None:
                    params[paramkey] = parse_rdata_text
```

pieterlexis-tomtom: Fixed!

ross:
Interesting, guess that wasn't enough. I'll poke around and see what I can find:

Screenshot 2024-06-06 at 3 33 55 PM

ross:
Pushed a slight reworking to get rid of the continue and instead use an else which afaict is logically equivilent and maybe a little more readable. Will hold off merging this until @pieterlexis-tomtom gets a change to 👀 and give a 👍, but I think it's good to go now.

https://github.com/octodns/octodns/blob/df85637d300262f53625396b56889b2cda45edfb/octodns/record/svcb.py#L160-L168

pieterlexis-tomtom:
> Pushed a slight reworking to get rid of the continue and instead use an else which afaict is logically equivilent and maybe a little more readable. Will hold off merging this until @pieterlexis-tomtom gets a change to 👀 and give a 👍, but I think it's good to go now.

LGTM!

ross:
Thanks

Issue #1177, opened 2024-06-04
Print plan in machine consumable format


olivergondza:
First and foremost, thanks for providing this awesome tool!

While working on automating zone management in ansible, I would benefit from a way to load the plan (executed or from a dry-run), and work with the results.

This will enable customized reporting of what have changed, and customized logic to suspend `--doit` invocation. IOW, I would like the automation to never delete records, and never change more than 10 records without explicit ack from an administrator. Access to the summary and listing will enable users to implement all that.

To do that:

- It would be beneficial to send to stdout/file (not stderr), so it is easy to separate from logging messages of varying severity,
- and format it as json/yaml.

Thanks!

ross:
There are currently a handful of `_PlanOutput`s available, located in https://github.com/octodns/octodns/blob/main/octodns/provider/plan.py. I could have sworn there was a JSON one, but apparently not. Doesn't look like any of them are particularly machine readable.  Would be fairly simple to create one.

Outputs are configured in the same way as processors, and there's a default `PlanLogger`:

https://github.com/octodns/octodns/blob/45753744a66982dd977de3bbe1ddceda07176e42/octodns/manager.py#L171-L180

PRs welcome, otherwise I'll stick it on the TODO list.

ross:
Threw something together quickly while waiting on something else https://github.com/octodns/octodns/pull/1178

mrwillrobbins:
Hi @ross, 

I was testing this and noticed the record.name was missing from the json output.  I needed this for pre/post checks so was perfect timing. 

I tested the below and seems to work.

```
class PlanJson(_PlanOutput):
    def __init__(self, name, indent=None, sort_keys=True):
        super().__init__(name)
        self.indent = indent
        self.sort_keys = sort_keys

    def run(self, plans, fh=stdout, *args, **kwargs):
        data = defaultdict(dict)
        for target, plan in plans:
            plan_data = {'changes': []}
            for change in plan.changes:
                record = change.record
                change_data = change.data
                change_data['name'] = record.name
                plan_data['changes'].append(change_data)
            data[target.id][plan.desired.name] = plan_data

        fh.write(dumps(data, indent=self.indent, sort_keys=self.sort_keys))
        fh.write('\n')
```
Thank you,
Will

ross:
> was testing this and noticed the record.name was missing from the json output

Doh! That makes it pretty much unless. What I get for not having a use case for the results and not poking around at it. 

[...]

PR #1178, opened 2024-06-04
Quick pass at implementing PlanJson

ross:
```yaml
manager:
  plan_outputs:
    json:
      class: octodns.provider.plan.PlanJson
      indent: '  '
```

Plan output schema is `"<target>": "<zone>": "changes"`, default is `indent=None` which would be more easily machine parsable. 

```json
{
  "out": {
    "exxampled.com.": {
      "changes": [
        {
          "new": {
            "ttl": 3600,
            "value": "1.2.3.4"
          },
          "type": "create"
        },
        {
          "new": {
            "ttl": 60,
            "values": [
              "octodns-version=1.7.0",
              "time=2024-06-04T20:59:50.929216+00:00",
              "uuid=c6613a81-2ba9-44ac-b378-9b4c59ffb04c"
            ]
          },
          "type": "create"
        },
        {
          "new": {
            "ttl": 3600,
            "value": "www.foo2."
          },
          "type": "create"
        }
      ]
    }
  }
}
```

/cc Fixes https://github.com/octodns/octodns/issues/1177 @olivergondza

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Allow DS records to be specified for managed sub-zones, same as NS
 * Fix CAA rdata parsing to allow values with tags
+* Include `record_type` in Change data
+* Add PlanJson plan_output support
 
 ## v1.7.0 - 2024-04-29 - All the knobs and dials
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 * Improved handling of present, but empty/None config file values.
 * Add PlanJson plan_output support
-
+* Include `record_type` in Change data
+ 
 ## v1.8.0 - 2024-06-10 - Set the records straight
 
 * Add support for SVCB and HTTPS records
 * Allow DS records to be specified for managed sub-zones, same as NS
 * Fix CAA rdata parsing to allow values with tags
-* Include `record_type` in Change data
 
 ## v1.7.0 - 2024-04-29 - All the knobs and dials
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
+## v1.?.? - 2024-??-?? - ???
+
+* Add PlanJson plan_output support
+
 ## v1.8.0 - 2024-06-10 - Set the records straight
 
 * Add support for SVCB and HTTPS records
 * Allow DS records to be specified for managed sub-zones, same as NS
 * Fix CAA rdata parsing to allow values with tags
 * Include `record_type` in Change data
-* Add PlanJson plan_output support
 
 ## v1.7.0 - 2024-04-29 - All the knobs and dials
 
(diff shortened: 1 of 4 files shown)


olivergondza:
Thanks for the prompt grooming, @ross!

My pair-review, provided with zero familiarity with octodns implementation considerations:

- I would default to indented variant—I have not seen a parser having issues with whitespace or indent. The plus is it is much easier to debug/investigate by human when things blow up.
- When presenting the plan data, it would be great to have it keyed by "modification type", so it will be easier to filter out for "Create", "Update", "Delete" operations present.
- I might be missing something, but is there a way to detect the record type?

Thanks!

ross:
> * I would default to indented variant—I have not seen a parser having issues with whitespace or indent. The plus is it is much easier to debug/investigate by human when things blow up.

You mentioned machine readable as the main use case. It's WAY easier to grab everything on the single line/up to the newline and feed it to a json parser than to feed multiple lines. 

The more human readable format are the other plan types and there's nothing preventing configuring multiple plan outputs, one of the machine readable and the other for humans, `PlanLogger`. 

> * When presenting the plan data, it would be great to have it keyed by "modification type", so it will be easier to filter out for "Create", "Update", "Delete" operations present.

The change ordering is intention/very specific and is implemented by the ordering of change objects, https://github.com/octodns/octodns/blob/45753744a66982dd977de3bbe1ddceda07176e42/octodns/record/change.py, and Plan https://github.com/octodns/octodns/blob/main/octodns/provider/plan.py#L54-L58

A lot of providers/tooling require that things happen in that order to avoid conflicting operations, e.g. trying to create a CNAME where an A currently exists or vice versa. Everything else in octoDNS works in this ordering and I wouldn't want an built-in plan output that didn't also use it. 

> * I might be missing something, but is there a way to detect the record type?

You are correct. The record type doesn't appear to be included. Will look into that. I'm guessing the other uses of plan.data didn't need it. Should be able to add it.

ross (commented):
/ci modules build

ross:
Added `record_type`:

```js
{
  "out": {
    "exxampled.com.": {
      "changes": [
        {
          "new": {
            "ttl": 3600,
            "value": "1.2.3.4"
          },
          "record_type": "A",
          "type": "create"
        },
        {
          "new": {
            "ttl": 60,
            "values": [
              "octodns-version=1.8.0",
              "time=2024-06-10T14:46:04.848047+00:00",
              "uuid=430142f9-4b54-4055-99af-9b03c3369c24"
            ]
          },
          "record_type": "TXT",
          "type": "create"
        },
        {
          "new": {
            "ttl": 3600,
            "value": "www.foo2."
          },
          "record_type": "CNAME",
          "type": "create"
        },
        {
          "new": {
            "ttl": 3600,
            "value": "This is a TXT record with \"quotes\" in it to ensure they are handled correctly"
          },
          "record_type": "TXT",
          "type": "create"
        }
      ]
    }
  }
}
```

olivergondza:
@ross, thanks a bunch for implementing this, and responding!

PR #1179, opened 2024-06-07
Move HTTPS to its own file, import them in octodns.record

ross:
The types need to be imoprted into `octodns.record` so that they're registered and supported by default. Ran into this problem when adding support in https://github.com/octodns/octodns-powerdns/pull/70.

Splitting the HTTPS type out into it's own file is just a minor tweak to make it work/look a little more like all the rest of the types. It also adds a no-op HttpsValue just to make things more consistent.

/cc https://github.com/octodns/octodns-powerdns/pull/70 which lead me to hit part of this.
/cc https://github.com/octodns/octodns/pull/1176 which this builds on/tweaks slightly

### DIFF
--- a/octodns/record/__init__.py
+++ b/octodns/record/__init__.py
@@ -14,6 +14,7 @@ from .dname import DnameRecord, DnameValue
 from .ds import DsRecord, DsValue
 from .exception import RecordException, ValidationError
 from .geo import GeoCodes, GeoValue
+from .https import HttpsRecord, HttpsValue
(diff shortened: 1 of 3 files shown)


pieterlexis-tomtom:
:+1:

sorry for missing this :)

ross:
> sorry for missing this :)

You had no way to know, I was the one who missed it. Figured implemeting things in octodns-powerdns would shake out things.

PR #1180, opened 2024-06-10
Version 1.8.0 bump & changelog update

ross:
## v1.8.0 - 2024-06-10 - Set the records straight

* Add support for SVCB and HTTPS records
* Allow DS records to be specified for managed sub-zones, same as NS
* Fix CAA rdata parsing to allow values with tags


ross (commented):
/ci modules bump

PR #1186, opened 2024-06-21
1.9.1 bump & changelog update

ross:
## v1.9.1 - 2024-06-21 - What's in a name

* Add `name` to Change.data, and thus PlanJson output


ross (commented):
/ci modules build

Issue #1189, opened 2024-07-03
Support adding additional fields to MetaProcessor


nogweii:
Add a configuration block to MetaProcessor, something like this:

```yaml
processors:
  meta:
    class: octodns.processor.meta.MetaProcessor
    record_name: meta
    extra:
      ci: env/CI_JOB_URL
      team: marketing
```

That way additional information can be added while taking advantage of the logic to only change the meta values when there has been a change elsewhere in the plan.

When there are conflicts, like if someone specified `include_uuid: true` and `extra: uuid: 1234`, I'm not sure what the best behavior is. Throw a warning, for sure, but which value wins?

PR #1190, opened 2024-07-03
Add MetaProcessor.include_extra

ross:
/cc Fixes https://github.com/octodns/octodns/issues/1189
/cc @nogweii

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v1.?.? - 2024-??-?? - ???
+
+* MetaProcessor.include_extra to add support for arbitrary extra values to be
+  set on the meta record.
+
 ## v1.9.1 - 2024-06-21 - What's in a name
 
 * Add `name` to Change.data, and thus PlanJson output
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci modules bump

nogweii:
Wow, that was fast! I was hoping to ensure the continued pattern of `key=value` through the configuration, rather than completely arbitrary strings. In the documentation's example, how would one create a `foo` key with an environment variable for the value (rather than `bar`)?

ross:
> I was hoping to ensure the continued pattern of `key=value` through the configuration

Hrm. That exact thing wouldn't be possible unless the `key=` was also in the env variable. Guess I was thinking arbitrary strings would be more flexible than forcing the `key=val`, but maybe thats more consistent with what it already does...

nogweii:
Indeed, it's more flexible. I don't know which is more important, that flexibility or consistency with the other keys.

Appending `foo=` to any external strings is not that difficult. 😆

ross:
I'll go ahead and switch it to a dict of keys and values. That's probalby more inline with what it already does.

PR #1191, opened 2024-07-03
Switch MetaProcessor.include_extra to k=v

ross:
/cc https://github.com/octodns/octodns/pull/1190#issuecomment-2206567161 @nogweii

### DIFF
--- a/octodns/processor/meta.py
+++ b/octodns/processor/meta.py
@@ -62,9 +62,8 @@ class MetaProcessor(BaseProcessor):
         # Extra values to set on the records
         # (default: None)
         #include_extra:
-        #  - key=val
-        #  - foo=bar
-        #  - arbitrary string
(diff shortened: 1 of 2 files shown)


nogweii:
Nice. This results in `include_extra` overriding the other includes. Is that intentional? (e.g. `uuid`)

ross:
> Nice. This results in `include_extra` overriding the other includes. Is that intentional? (e.g. `uuid`)

Hadn't put much thought into it, but I'd lean towards it being this way as it would allow people to do funky stuff if they have good reason to, e.g. get time to be something coming form somewhere else. I'm stretching, but maybe they want to use the same time value across multiple octodns runs/configs or something. 

Definitely a potential 🦶 🔫 , but I'm OK with it :grin:

nogweii (approved):

PR #1204, opened 2024-10-07
v1.10.0 version bump & changelog update

ross:
## v1.10.0 - 2024-10-06 - Lots of little stuff

* Zone name validation checking for double dots, and throwing InvalidNameError
  rather than base Exception
* Record validation checks for double dots in names
* MetaProcessor.include_extra to add support for arbitrary extra values to be
  set on the meta record.
* Correctly handled quoted svcparams when parsing SVCB/HTTPS rdata text
* Fix handling of chunked values (TXT, SPF) when escaped characters land at the
  split boundaries, don't split escapes from their following chars

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v1.?.? - 2024-??-?? - ???
+## v1.10.0 - 2024-10-06 - Lots of little stuff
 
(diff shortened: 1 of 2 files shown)


ross (commented):
/ci modules bump

ross (commented):
/ci modules bump

PR #1245, opened 2025-04-18
Add python 3.14 to the ci matrix

ross:


ross:
Looks like 3.14 isn't available in actions yet. Will let this sit until then.

ross:
Mis-read, this is months out. Closing.

PR #1246, opened 2025-04-28
Ns1 filter chain patch

mtgagnon:
In OctoDNS NS1, filter chain records are created as lenient null records with no values associated with them. When also using a value filter processor this caused OctoDNS to crash as there was no value to parse, even when the record is lenient. I added a warning for when a value filter tries to process a null record, and then ignores the record. If the record is not lenient then OctoDNS will not allow a blank record so this does not cause issues with creating invalid records.

### DIFF
--- a/octodns/processor/filter.py
+++ b/octodns/processor/filter.py
@@ -234,8 +234,10 @@ class _ValueBaseFilter(_FilterProcessor):
             if hasattr(record, 'values'):
                 values = [value.rdata_text for value in record.values]
             else:
-                values = [record.value.rdata_text]
-
+                try:
+                    values = [record.value.rdata_text]
+                except AttributeError:
+                    self.log.warning(f"Record value is NoneType: {record.fqdn}")
             if any(value in self.exact for value in values):
                 self.matches(zone, record)
                 continue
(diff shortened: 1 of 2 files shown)


octodns/processor/filter.py
```diff
@@ -234,8 +234,12 @@ def _process(self, zone, *args, **kwargs):
...
                 values = [value.rdata_text for value in record.values]
             else:
-                values = [record.value.rdata_text]
-
+                try:
+                    values = [record.value.rdata_text]
+                except AttributeError:
+                    self.log.warning(
+                        f"Value processor ignoring record, value is NoneType: {record.fqdn}"
+                    )
```
ross, line 242: Since this is already doing `if hasattr(record, 'values'):` up above I think it'd make sense to use an elif to check for a non-None value too rather than add a 2nd level of branching with the try/except:

```python
elif  record.value is not None:
    values = [record.value.rdata_text]
else:
    self.log.warning('value for %s is NoneType, ignoring', record.fqdn)
```

Also reworked the log line. No need to include "Value processor" since the class name and ID will be the logger name. Also using placeholders rather than f-string as is the standard practice for log lines (in octoDNS.)

octodns/processor/filter.py
```diff
@@ -281,6 +285,7 @@ class ValueAllowlistFilter(_ValueBaseFilter, AllowsMixin):
     '''
 
     def __init__(self, name, allowlist):
+        self.log = getLogger(f'ValueAllowlistFilter[{name}]')
```
ross, line 288: Good catch.

ross (commented):
Looks good. One comment inline.

ross (approved):
Thanks

PR #1249, opened 2025-05-16
feat: Add NetworkInterfaceSource

provokateurin:
I've been using https://github.com/octodns/octodns-ddns, but the external service used for getting the IP addresses sometimes fails. In reality my setup can be much simpler and octoDNS should just use the IPs from the network interface.

Therefore I created this small source which does exactly that and also allows filtering which IPs are used (with a sane default for only global addresses), as for example loopback addresses usually make no sense to publish via DNS.

I was not sure if this should rather go into a separate repository like many other providers, but I felt like this source is quite neutral and somewhat comparable to the envvar one as it just reads some value from the system.

### DIFF
--- a/octodns/source/networkinterface.py
+++ b/octodns/source/networkinterface.py
@@ -1,0 +1,110 @@
+import logging
+from ipaddress import ip_address
+
+import ifaddr
+
+from ..record import Record
+from .base import BaseSource
+
+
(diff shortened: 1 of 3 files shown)


provokateurin:
Thank you for your feedback, I'll check with octoddns if this functionality makes sense in there or if I create a completely new source instead.

PR #1251, opened 2025-05-24
Fix typo in AcmeManagingProcessor

travisty-:
There appears to be a typo in the name of the ACME processor.

Renames "AcmeMangingProcessor" to "AcmeManagingProcessor".

### DIFF
--- a/octodns/processor/acme.py
+++ b/octodns/processor/acme.py
@@ -59,3 +59,6 @@ class AcmeManagingProcessor(BaseProcessor):
                 existing.remove_record(record)
 
         return existing
+
+
+AcmeMangingProcessor = AcmeManagingProcessor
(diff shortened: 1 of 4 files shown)


octodns/processor/acme.py
```diff
@@ -7,14 +7,14 @@
 from .base import BaseProcessor
 
 
-class AcmeMangingProcessor(BaseProcessor):
-    log = getLogger('AcmeMangingProcessor')
+class AcmeManagingProcessor(BaseProcessor):
```
ross, line 10: Wow. That's a dumb one. Given that no one has noticed I wonder if it's even getting used by anybody. Just in case we should throw the following down after the class to alias the old name. 

```python
AcmeMangingProcessor = AcmeManagingProcessor
```

travisty-: This was the first processor I reached for when I was setting up OctoDNS at work today, so I'm surprised that this one flew under the radar lol. Oh good idea, that way we can avoid a breaking change. I pushed another commit adding an alias.

Issue #1258, opened 2025-06-10
Get current zone name as a variable in YAML files


jleroy:
I'm working to deploy DKIM records across a number of domains. I'm using a dedicated provider in conjunction with `shared_filename` and `disable_zonefile` to add the records in questions only to the zones that need them:

```yaml
providers:
  ...
  fastmail:
    class: octodns.provider.yaml.YamlProvider
    directory: ./snippets/Fastmail
    shared_filename: Fastmail.yaml
    disable_zonefile: true
  ...

zones:
  ...
  example.com.:
    sources:
      - fastmail
      - config_example
    targets:
      - bunny
    ...
```

But the DKIM records themselves are handled by our provider, who asks to add the following CNAME records to each zone:

```yaml
fm1:
  type: CNAME
  value: fm1.[domain.tld].dkim.fmhosted.com.

fm2:
  type: CNAME
  value: fm2.[domain.tld].dkim.fmhosted.com.

fm3:
  type: CNAME
  value: fm3.[domain.tld].dkim.fmhosted.com.
```

How can I set the zone name dynamically in the YAML file?
A solution can be to have something like a "$zone_name" variable who will be replaced by YAMLProvider when reading records names / values?

ross:
Have a look at https://github.com/octodns/octodns/pull/1259 for what I have in mind. I've run out of time for it today and it may be a while before I get back to finish it as we have guests for the next week...

jleroy:
Now I encounter this error, as if the processor was called to late? 

```python
Traceback (most recent call last):
  File "/Users/jleroy/Developments/octodns-bunny/env/bin/octodns-sync", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/jleroy/Developments/octodns-bunny/env/lib/python3.13/site-packages/octodns/cmds/sync.py", line 62, in main
    manager.sync(
    ~~~~~~~~~~~~^
        eligible_zones=args.zone,
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        checksum=args.checksum,
        ^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/jleroy/Developments/octodns-bunny/env/lib/python3.13/site-packages/octodns/manager.py", line 777, in sync
    ps, d = future.result()
            ~~~~~~~~~~~~~^^
[...]

PR #1259, opened 2025-06-10
Implment Templating processor and prototype Value.template methods

ross:
```yaml
...

processors:
  templater:
    class: octodns.processor.templating.Templating

...
zones:
  example.com:
    sources: ...
    processors:
      - templater
    targets: ...
```

```yaml
foo:
  type: TXT
  values:
    - There are {zone_num_records} record(s) in {zone_name}
    - Hello World!
```

/cc Fixes https://github.com/octodns/octodns/issues/1258 
/cc @jleroy


ross:
OK. I think this is good to ship as a first pass. It's mostly standalone and care has been taken (along with testing) to make sure that custom record types w/o the `template` method should be unaffected. :crossed_fingers:

PR #1261, opened 2025-06-16
DRAFT: Add Sphinx for Documentation

olofvndrhr:
Here is the first iteration of the added Documentation via Sphinx.
I played with it a bit and it looks promising. This is of course not "final" but a prototype.
Themes, settings etc can be tweaked.

### Live preview
To see how everything looks, check:  https://olofvndrhr.github.io/octodns/
It will be rebuilt on all changes of this PR, so everyone can see how it then looks.

### Automatic Documentation via [Sphinx Autodocs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)

**Documentation file structure**
```txt
octodns/
└── docs/
    ├── _build/      -> generated html (are git ignored)
    ├── modules/     -> automatically generated module files (are git ignored)
    ├── _static/     -> static content like files, icons etc.
    ├── infos/       -> special pages with project infos (license, changelog etc.)
    └── pages/       -> manual documentation about topics/guides
```

**There are also two important Sphinx extensions in use:**
- [sphinx.ext.napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) -> which makes it possible to use the NumPy oder Google Docstring style
- [myst_parser](https://myst-parser.readthedocs.io/en/latest/) -> which makes it possible to write documentation in Markdown format instead of ReStructuredText

#### Manual
1. Run the script `script/get-docs` to gather all modules from octodns and create the `.rst` files which Sphinx needs.
2. Run the script `script/build-docs` to build the documentation. Output will be on the path `docs/_build/html`
3. Check the generated docs -> open `docs/_build/html/index.html`

> The generated modules are in the `.gitignore `as they would just spam the repo with files

#### GitHub Action
A GitHub actions workflow was also created to build on PR, Tag and manual trigger.
1. Installs python and all dependencies from requirements-dev.txt
2. Installs the project (octodns) locally
3. Gathers the modules and builds the documentation
4. Pushes all built documentation (the html) to the branch `gh-pages`
5. _(if configured)_ The GitHub bot deploys the html to GitHub Pages

### Automatic Markdown formatting

As all new documentation will be written in Markdown, there needs to be a standard to follow.
As i already saw on my previous PR, auto-formatters will change the formatting if not careful.

So i added [mdformat](https://github.com/hukkin/mdformat) to the workflow.
Why mdformat? Because it has no dependencies and is also written in python (instead of something like [prettier](https://prettier.io/)).

I adapted the scripts `script/format` to include markdown files and ran it also. Thats why so many markdown files are changed in the PR.

### Other

- The requirements were updated for the new dependencies
- Some Docstrings were updated for the new format
- The ACME processor was updated with new documentation (as a prototype to see what's possible and how it looks)

*Best regards*


octodns/processor/acme.py
```diff
@@ -8,36 +8,61 @@
 
 
 class AcmeManagingProcessor(BaseProcessor):
-    log = getLogger('AcmeManagingProcessor')
+    log = getLogger("AcmeManagingProcessor")
```
ross, line 11: What made all the quote changes from ' to "?

octoDNS's pyproject.toml tells black to leave string quoting alone as the single quote variant is used as the standard across all of the octoDNS modules (predating black's usage.) 

https://github.com/octodns/octodns/blob/c2be2bc0e2c6f4aff461ecbba2ed2b5370c4673f/pyproject.toml#L3

It's a lot of churn here and would be across all the modules. Wouldn't be the end of the world to change it, but I think I'd prefer to do it in a dedicated PR if/when that happens since it isn't directly related to the documentation.  We'd also need to make sure that the black settings change too.

ross (commented):
That looks really good. My first question was whether the examples stuff would be included and I'm glad to see that it is. 

> I adapted the scripts `script/format` to include markdown files and ran it also. Thats why so many markdown files are changed in the PR.

See https://github.com/octodns/octodns/blob/main/.git-blame-ignore-revs for previous times when large formatting-specific changes were made. That file allows us to tell git/GitHub to ignore those revs when looking at the history of files so that don't show up as noisy changes. If all the formatting-specific changes happen in a standalone commit we can add its SHA to the list w/a comment about it.

> The requirements were updated for the new dependencies

Would it make sense to have a `docs` specific `extras_require` and then modify `./script/update-requirements` to create `requirements-docs.txt`? Don't have an opinion really, just offering it up as a option.

> The ACME processor was updated with new documentation (as a prototype to see what's possible and how it looks)

Nice. Having a destination for the docs will definitely make me more likely to go through and clean up and add new stuff.

octodns/processor/acme.py
olofvndrhr: I run [ruff](https://github.com/astral-sh/ruff) as an auto-formatter in my IDE, and the default there is double-quotes. That's why its changed on all files where i modified the Docstring.
Will change it back.
And before the real PR is opened i will also squash the commits, so its not a mess of commits.

ross: Shame it doesn't use the black settings in pyproject.toml since it's aiming to replace it and other related tooling.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
+name: OctoDNS Documentation
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+  pull_request:
+    branches: [main, master]
```
ross, line 9: all the octoDNS repos use `main`

```suggestion
    branches: [main]
```

requirements-dev.txt
```diff
@@ -1,48 +1,65 @@
...
+requests==2.32.4
 rfc3986==2.0.0
-rich==13.9.2
-twine==5.1.1
-urllib3==2.2.3
-wheel==0.44.0
-zipp==3.20.2
+rich==14.0.0
+ruamel.yaml.clib==0.2.12
+ruamel.yaml==0.18.14
```
ross, line 60: What OS and python are you working on? Seeing a bit more change here than I'd expect from an update and wondering if it's different requirements on different envs.

script/update-requirements
```diff
@@ -25,34 +27,28 @@ with open('setup.py') as fh:
...
-
-with open('requirements-dev.txt', 'w') as fh:
-    fh.write(f'# DO NOT EDIT THIS FILE DIRECTLY - use {script} to update\n')
-    fh.write('\n'.join(dev_frozen))
-    fh.write('\n')
+    for target in TARGETS:
+        to_install = f'.[{target}]' if target else '.'
+
+        check_call([join(tmpdir, 'bin', 'pip'), 'install', to_install])
+        frozen = check_output([join(tmpdir, 'bin', 'pip'), 'freeze'])
```
ross, line 34: There's a subtle change in behavior here. Previously -dev built on top of the base requirements and only included the additional packages. Main reason for that was to avoid the possibility of having different versions of the same package across the files which would result churn when doing a script/bootstrap. I believe that could happen here if one of the packages in -dev or -docs restricted a package's versions more than the base.

olofvndrhr: Good catch, i did actually miss that.

I wrote now a new version which follows the old behaviour.
But as we now are installing multiple extras and the cross dependencies could be hard to filter out i tried to do it for every extra "section" in a new venv. So we are sure only the correct extra dependencies are saved.

Let me know what you think. To be honest, its a completely different script now. So feel free to decline.

ross: I'll take another :eyes: at it when I get a chance. Not opposed to the changes and DRYing it up now that there's a 3rd path. Perhaps it'd be workable to just have it continue to be cummlative and install the base requirements, then have dev be additional on top of that, then the doc on top of dev. I don't imagine anyone would be installing doc w/o dev.

olofvndrhr: Thats true, but if there is another extra section in the future it would break again :)
My latest commit simplifies it more, and from my tests it runs like the old one, just now with the possibility for unlimited extra sections.

olofvndrhr:
> Yeah. black only offered a "leave quotes alone" option so I went with that. There have definitely been things slip through PR review and stuff around quoting.

And for that i would like to know what route you want to continue on, either:
- still use `black` and `isort`, but add formatter settings to `pyproject.toml`
- switch to `ruff` as it already includes both and has some nice stuff that "fixes issues"

For both routes you would need to decide some points like:
- quote settings for strings
- quote settings for docstrings
- line length (still 80?)
- docstring style (google/numpy)

olofvndrhr:
> Not 100% sure what you mean by "formatter" settings here. There's already some black settings in there.

For ruff there are more settings, but as black works fine, only the option `skip-string-normalization=true` would need to  be removed to unify all quotes.
But thats something for another PR, you're right about that

ross:
> script/update-requirements

Had to make some fairly minor changes to the logic of the update-requirement scripts across all the repos to deal with `click` deprecating support for python 3.9 early. Should be fairly straightforward to add that logic to the update version in here. LMK if you want me to take a poke at it.

olofvndrhr:
> > script/update-requirements
> 
> Had to make some fairly minor changes to the logic of the update-requirement scripts across all the repos to deal with `click` deprecating support for python 3.9 early. Should be fairly straightforward to add that logic to the update version in here. LMK if you want me to take a poke at it.

Updated the script. Now also works on my version as it doesn't always "find" click.
```txt
    i = [i for i, r in enumerate(output) if r.startswith('click==')][0]
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

PR #1266, opened 2025-06-19
Fixes for geo_data.py action

ross:
/cc https://github.com/octodns/octodns/pull/1265 which introduced this script
/cc https://github.com/octodns/octodns/actions/runs/15762916330/job/44433463900 which was the first manual test run


ross:
Seems to be good. Will have to wait until something changes naturally to know for sure.

PR #1271, opened 2025-06-27
Add basic documentation on configuring/utilising octoDNS processors

g-a-c:
I was unable to find _any_ solid references on processors online, apart from a [blog article](https://jpmens.net/2025/02/22/notes-to-self-on-octodns-and-its-providers/) complaining about the documentation...

Because there are multiple `processors` keys in different places, with different expected data types, it took me some trial and error to get this working for my own `MetaProcessor` use case so I decided to write a quick document on which key does what.

This could definitely be _more_ complete, but it feels like a reasonable start to me and should hopefully be enough to get the kind of user who is like to use octoDNS up and running.

Feedback appreciated!

### DIFF
--- a/.changelog/c4f025d1c23c40dd98380e6d3496364d.md
+++ b/.changelog/c4f025d1c23c40dd98380e6d3496364d.md
@@ -1,0 +1,4 @@
+---
+type: none
+---
+Documentation for processors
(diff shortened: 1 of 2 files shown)


ross (approved):
Looks good. When I started working through the stuff in the `examples` directory I actually started on one focused on `processors`, but didn't get the time to finish it. The other thing i wanted to do with it was demo how easy it is to write completely custom processors suited to whatever specific needs people come up with.

PR #1272, opened 2025-06-28
expirimentation around using jsonschema for validation

ross:
Ran across [jsonschema](https://pypi.org/project/jsonschema/) in a talk by @gizm00 at a [python meetup](https://www.meetup.com/pdxpython/events/308298313/?eventOrigin=group_past_events) and made a mental note to look into it. I thought it'd might be interesting to explore shifting record validation over to use it and ideally do a schema for the main config file and processors/providers/sources.

Look like it'd take some effort to get there, but in theory it should be able to do (most) everything we're currently doing, but in a more systemic way. 

If this direction is followed I think the biggest hurdles would be:
* Making sure that the source context is preserved and presented in the error message. The validate method takes a python dict which has no info about where in a yaml file or other source provider that data came from so it can't give the user line numbers or anything to help them track down the problem. This would be a show-stopper
* Making sure the error messages make sense to end uses. As-is they dump the relivant section of jsonschema which is going to be jibberish. The actual error message seems reasonable, there's just a lot of less useful noise accompanying it. 
* Making sure to preserve the `lenient` behavior where errors warn 

Anyway, all of this makes me feel like if this happens then it needs to be part of a major release, with an opt-in mode for several releases/a good while beforehand.

Links and notes:
* Validation of simple string formats, e.g. ipv4/ipv5, fqdns, etc. https://python-jsonschema.readthedocs.io/en/stable/validate/#validating-formats &amp; https://www.learnjsonschema.com/2020-12/format-annotation/format/
* Did a lot of searching and faffing about trying to figure out how to get different validations to apply based on the record `type` as well as get errors for unexpected keys etc. I think I got there, some links involved in figuring that out:
   * https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse
   * https://github.com/json-schema-org/json-schema-spec/issues/556 "A simple example"
   * https://stackoverflow.com/questions/49784760/jsonschema-validate-type-based-on-value-of-another-property
* TBD how this would hook into the flow.
   * There's the obvious case where `Record.validate` would use it
   * Less clean is how YamlProvider might wrap this schema in it's own enclosure/objext to validate it's specific schema on top of the individual records

Parking this for now. It seems pretty interesting so I may play around with it, but tbd if I get into it enough to get it from here to a usable state.

/cc https://pypi.org/project/jsonschema/

### DIFF
--- a/octodns/record/a.py
+++ b/octodns/record/a.py
@@ -13,6 +13,7 @@ from .ip import _IpValue
 class Ipv4Value(_IpValue):
     _address_type = _IPv4Address
     _address_name = 'IPv4'
+    _format = 'ipv4'
 
 
 Ipv4Address = Ipv4Value
--- a/octodns/record/aaaa.py
+++ b/octodns/record/aaaa.py
@@ -13,6 +13,7 @@ from .ip import _IpValue
 class Ipv6Value(_IpValue):
     _address_type = _IPv6Address
     _address_name = 'IPv6'
+    _format = 'ipv6'
 
 
 Ipv6Address = Ipv6Value
--- a/octodns/record/base.py
+++ b/octodns/record/base.py
@@ -42,6 +42,54 @@ class Record(EqualityTupleMixin):
         return cls._CLASSES
 
     @classmethod
+    def jsonschema(cls):
+        schema = {
+            # base Record requirements
+            'title': 'Record',
+            'type': 'object',
+            'properties': {
+                'type': {'type': 'string', 'enum': list(cls._CLASSES.keys())}
+            },
+        }
+
+        class_schemas = []
+        for _type, _class in cls._CLASSES.items():
+            _value_type = _class._value_type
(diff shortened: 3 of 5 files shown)


ross:
I'm really liking the idea of this, but I think it'd be a net-loss for octoDNS users as the error messages are much less human-readable/actionable. I've come up with ways to translate them (and will push up a bit more of that) but it's getting pretty clunky and w/o something like t-strings I don't think I'd be 100% happy with the results. Going to park this and let it auto-close. I may revisit in the future as i really like the "spec" nature of this, but I'm not happy with the lack of context/detail in the error messages.

ross:
Recent set of linked PRs got A jsonschema up and going for use in editors/linters. It seems to be working well, but I'd call it Beta at this point. 

It is not used internally for validation for the reasons outlined above, namely the error messages aren't great and seem to be more targeted at live editing rather than runtiume consumption. I'll revisit this in the future to see if anything has changed, but for now there's progress, but not completion.

PR #1274, opened 2025-07-06
Fix include target in some Processors

olofvndrhr:
Fixes the use of `include_target` in some filters where the kwargs were not passed to its parent class.

Error before:
`TypeError: NameAllowlistFilter.__init__() got an unexpected keyword argument 'include_target'`

> (also added a missing property of `ZoneNameFilter`)

### DIFF
--- a/.changelog/4f9d1e4e41834834a21cd99ef46af928.md
+++ b/.changelog/4f9d1e4e41834834a21cd99ef46af928.md
@@ -1,4 +1,4 @@
 ---
-type: none
+type: minor
 ---
-fix include-target in filter processors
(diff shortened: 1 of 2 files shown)


.changelog/4f9d1e4e41834834a21cd99ef46af928.md
```diff
@@ -0,0 +1,4 @@
+---
+type: none
+---
+fix include-target in filter processors
```
ross, line 4:
```suggestion
Add kwargs to filter processors init params
```

ross:
Thanks

PR #1278, opened 2025-07-28
Templating processor: remove trailing dot on FQDN variables

jleroy:
Thank you again for your work on the templating feature @ross.
As promised in #1258 and #1259, I finally had the time to test it. This PR fix two issues in the templating processor.

`zone_name`, `zone_decoded_name`, `zone_encoded_name`, `record_fqdn`, `record_decoded_fqdn` and `record_encoded_fqdn` variables contain a trailing dot, which seems consistent with OctoDNS internals but makes these variables basically unusable for record with target values like `CNAME`, `MX`, `SRV`...

For example, a CNAME record with value `{zone_name}-in.provider.com.` in zone `example.com` will be translated to `example.com.-in.provider.com.` and not `example.com-in.provider.com.` as expected by most users. So I think it’s better to remove the trailing dots, even if it means adding it after the variable if needed. Also the feature is pretty new so it shouldn't impact a lot of users (if any).

The processor is broken on alias zones. The variables are replaced by the values of the parent zone instead of the current zone. In this example, `example.net` is an alias of `example.com`:
```yaml
z2éuu:
  type: TXT
  value: record_name={record_name}; record_decoded_name={record_decoded_name}; record_encoded_name={record_encoded_name}; record_fqdn={record_fqdn}; record_decoded_fqdn={record_decoded_fqdn}; record_encoded_fqdn={record_encoded_fqdn}; record_type={record_type}; record_ttl={record_ttl}; record_source_id={record_source_id}
```

```
* example.net.
* googlecloud (GoogleCloudProvider)
*   Create <TxtRecord TXT 3600, z2éuu.example.net., ['record_name=z2éuu; record_decoded_name=z2éuu; record_encoded_name=xn--z2uu-cpa; record_fqdn=z2éuu.example.com; record_decoded_fqdn=z2éuu.example.com; record_encoded_fqdn=xn--z2uu-cpa.example.com; record_type=TXT; record_ttl=3600; record_source_id=config']> (config)
*   Summary: Creates=1, Updates=0, Deletes=0, Existing=0, Meta=False
```

I tried to add unit tests to check the value of records handled by templating processor, without success. I don’t see how to do that cleanly. Any idea ?

### DIFF
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -4,6 +4,7 @@
 
 from collections import deque
 from concurrent.futures import ThreadPoolExecutor
+from copy import deepcopy
 from hashlib import sha256
 from importlib import import_module
 from importlib.metadata import PackageNotFoundError
@@ -500,7 +501,7 @@ class Manager(object):
         if desired:
             # This is an alias zone, rather than populate it we'll copy the
             # records over from `desired`.
-            for _, records in desired._records.items():
+            for _, records in desired._pre_processing_records.items():
                 for record in records:
                     zone.add_record(record.copy(zone=zone), lenient=lenient)
         else:
@@ -519,6 +520,10 @@ class Manager(object):
                         source.__class__.__name__,
                     )
                     source.populate(zone)
+
+            # Frozen copy of zone records *before* going through processors so
+            # alias zone can copy them if needed.
+            zone._pre_processing_records = deepcopy(zone._records)
 
         for processor in processors:
             zone = processor.process_source_zone(zone, sources=sources)
@@ -690,6 +695,7 @@ class Manager(object):
                     raise ManagerException(msg)
 
                 aliased_zones[zone_name] = source_zone
+                self.log.info('sync:     alias_of=%s', source_zone)
                 continue
 
             lenient = config.get('lenient', False)
--- a/octodns/manager.py
+++ b/octodns/manager.py
@@ -4,7 +4,6 @@
 
 from collections import deque
 from concurrent.futures import ThreadPoolExecutor
-from copy import deepcopy
 from hashlib import sha256
 from importlib import import_module
 from importlib.metadata import PackageNotFoundError
@@ -501,7 +500,7 @@ class Manager(object):
         if desired:
             # This is an alias zone, rather than populate it we'll copy the
             # records over from `desired`.
-            for _, records in desired._pre_processing_records.items():
+            for _, records in desired._records.items():
                 for record in records:
                     zone.add_record(record.copy(zone=zone), lenient=lenient)
         else:
@@ -520,10 +519,6 @@ class Manager(object):
                         source.__class__.__name__,
                     )
                     source.populate(zone)
-
-            # Frozen copy of zone records *before* going through processors so
-            # alias zone can copy them if needed.
-            zone._pre_processing_records = deepcopy(zone._records)
 
         for processor in processors:
             zone = processor.process_source_zone(zone, sources=sources)
@@ -695,7 +690,6 @@ class Manager(object):
                     raise ManagerException(msg)
 
                 aliased_zones[zone_name] = source_zone
-                self.log.info('sync:     alias_of=%s', source_zone)
                 continue
 
             lenient = config.get('lenient', False)
--- a/octodns/processor/templating.py
+++ b/octodns/processor/templating.py
@@ -66,10 +66,17 @@ class Templating(BaseProcessor):
 
     def process_source_zone(self, desired, sources):
         sources = sources or []
+        zone_name = desired.decoded_name
+        zone_decoded_name = desired.decoded_name
+        zone_encoded_name = desired.name
+        if not self.trailing_dots:
+            zone_name = zone_name[:-1]
+            zone_decoded_name = zone_decoded_name[:-1]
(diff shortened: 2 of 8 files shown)


octodns/manager.py
```diff
@@ -520,6 +521,10 @@ def _populate_and_plan(
                     )
                     source.populate(zone)
 
+            # Frozen copy of zone records *before* going through processors so
+            # alias zone can copy them if needed.
+            zone._pre_processing_records = deepcopy(zone._records)
```
ross, line 526: Does `deepcopy` actually know how to make identical records? I would have thought it would be able to safely do that w/objects.

There is shallow copy support in Zone, but I'd have to page the details back into memory to know if it was useful here or not.

ross, line 526: Does `deepcopy` actually know how to make identical records? I would have thought it would be able to safely do that w/objects.

There is shallow copy support in Zone, but I'd have to page the details back into memory to know if it was useful here or not.

octodns/processor/templating.py
```diff
@@ -62,9 +62,9 @@ def __init__(self, id, *args, context={}, **kwargs):
     def process_source_zone(self, desired, sources):
         sources = sources or []
         zone_params = {
-            'zone_name': desired.decoded_name,
-            'zone_decoded_name': desired.decoded_name,
-            'zone_encoded_name': desired.name,
+            'zone_name': desired.decoded_name.rstrip('.'),
+            'zone_decoded_name': desired.decoded_name.rstrip('.'),
+            'zone_encoded_name': desired.name.rstrip('.'),
```
ross, line 67: Generally octoDNS removes the trailing `.` w/`[:-1]`, probably best to do that, based on a param, for consistency's sake.

octodns/manager.py
```diff
@@ -500,7 +501,7 @@ def _populate_and_plan(
         if desired:
             # This is an alias zone, rather than populate it we'll copy the
             # records over from `desired`.
-            for _, records in desired._records.items():
+            for _, records in desired._pre_processing_records.items():
```
ross, line 504: I believe this would break things that depend on/expect the processors to have run. E.g. if you're using filtering processors to remove/ignore things the alias zone would have them and the parent wouldn't.

ross (commented):
> For example, a CNAME record with value {zone_name}-in.provider.com. in zone example.com will be translated to example.com.-in.provider.com. and not example.com-in.provider.com. as expected by most users. So I think it’s better to remove the trailing dots, even if it means adding it after the variable if needed. Also the feature is pretty new so it shouldn't impact a lot of users (if any).

Will have to think about this for a bit. Part of me thinks it'd be best to have seperate versions of things w/o the dots so you can get whichever you want. I see the logic of always being able to add them though so not having them kinda makes sense. I think the two options are:

1. Separate variables `zone_name` and `zone_name_no_dot` or whatever 
2. A processor parameter `trailing_dots` that defaults to True for backwards compat and can be set to False to get the no dot behavior.

After thinking about it for a bit I think I like option 2.

> The processor is broken on alias zones. The variables are replaced by the values of the parent zone instead of the current zone. In this example, `example.net` is an alias of `example.com`:

This seems like a separate issue so I'd prefer to tackle it in a different PR. As mentioned in the inline comments, I think there's actually going to be a bunch of hiccups/details to trying to get this right so I'll probably have to dig in and see what's possible. TBH the best option, for now anyway, might actually be to use symlinks to files rather than alias zones.

ross:
> > The processor is broken on alias zones. The variables are replaced by the values of the parent zone instead of the current zone. In this example, `example.net` is an alias of `example.com`:
> 
> This seems like a separate issue so I'd prefer to tackle it in a different PR. As mentioned in the inline comments, I think there's actually going to be a bunch of hiccups/details to trying to get this right so I'll probably have to dig in and see what's possible. TBH the best option, for now anyway, might actually be to use symlinks to files rather than alias zones.

Once I poked around and thought about it a bit I realized that `Processor.process_source_and_target_zones` was the place that should have been doing the templating. That happens later just before the plan between the desired and existing zones is created and is specific to the target provider and the EXACT zone since its state in the two zones can be different and require different changes. This leaves an un-templated copy of desired in Manager and everything just works.

/cc https://github.com/octodns/octodns/pull/1279

octodns/processor/templating.py
jleroy: Fixed.

jleroy:
I reverted the commit regarding the handling of alias zone as you fixed it in #1279.


> 2. A processor parameter `trailing_dots` that defaults to True for backwards compat and can be set to False to get the no dot behavior.
> 
> After thinking about it for a bit I think I like option 2.

Updated my PR accordingly.

octodns/processor/templating.py
```diff
@@ -75,9 +80,15 @@ def process_source_zone(self, desired, sources):
...
+        if not self.trailing_dots:
+            zone_params = zone_params | {
+                'zone_name': desired.decoded_name[:-1],
+                'zone_decoded_name': desired.decoded_name[:-1],
+                'zone_encoded_name': desired.name[:-1],
+            }
 
         def params(record):
-            return {
+            record_params = {
```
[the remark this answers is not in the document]
jleroy: Done

jleroy:
It should be good now :)

PR #1279, opened 2025-07-29
Fix issue with using Templating processor on aliased zones

ross:
Switches the processor to use process_source_and_target_zones which
happens later, during planning changes and thus happens separately for
the source and aliased zones. This leaves the templates in the copy of
desired that Manager uses to start the alias process. The updated copy
is in the plan so externally everything will still make sense and
behave as expected.

```
* primary.com.
* out (YamlProvider)
*   Create Zone<primary.com.>
*   Create <TxtRecord TXT 3600, primary.com., ['This is not a template']> (config)
*   Create <TxtRecord TXT 3600, templated.primary.com., ['This is a template, zone_name=primary.com.']> (config)
*   Summary: Creates=2, Updates=0, Deletes=0, Existing=0, Meta=False
* alias.com.
* out (YamlProvider)
*   Create Zone<alias.com.>
*   Create <TxtRecord TXT 3600, alias.com., ['This is not a template']> (config)
*   Create <TxtRecord TXT 3600, templated.alias.com., ['This is a template, zone_name=alias.com.']> (config)
*   Summary: Creates=2, Updates=0, Deletes=0, Existing=0, Meta=False
```

/cc @jleroy https://github.com/octodns/octodns/pull/1278

### DIFF
--- a/.changelog/519279fa6fa94b2fb61bd3552084dbd6.md
+++ b/.changelog/519279fa6fa94b2fb61bd3552084dbd6.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Fix issues with using Templating processor on alias zones
--- a/octodns/processor/templating.py
+++ b/octodns/processor/templating.py
@@ -59,19 +59,17 @@ class Templating(BaseProcessor):
         super().__init__(id, *args, **kwargs)
         self.context = context
 
-    def process_source_zone(self, desired, sources):
-        sources = sources or []
+    def process_source_and_target_zones(self, desired, existing, provider):
         zone_params = {
             'zone_name': desired.decoded_name,
             'zone_decoded_name': desired.decoded_name,
             'zone_encoded_name': desired.name,
             'zone_num_records': len(desired.records),
-            'zone_source_ids': ', '.join(s.id for s in sources),
             # add any extra context provided to us, if the value is a callable
             # object call it passing our params so that arbitrary dynamic
             # context can be added for use in formatting
             **{
-                k: (v(desired, sources) if callable(v) else v)
+                k: (v(desired, provider) if callable(v) else v)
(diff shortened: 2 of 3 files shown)


jleroy:
Thank you @ross, everything seems to be working fine.

Just one small issue. When a record is using an invalid variable, a KeyError is raised without any context:
```python
[...]
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/octodns/processor/templating.py", line 96, in process_source_and_target_zones
    new_values = [v.template(params(record)) for v in record.values]
                  ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/octodns/record/chunked.py", line 89, in template
    return self.__class__(self.format(**params))
                          ~~~~~~~~~~~^^^^^^^^^^
KeyError: 'zone_source_ids'
```

I’ve made a small [patch](https://github.com/user-attachments/files/21492643/RaiseValidationError.patch) to raise a ValidationError instead:
```python
[...]
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/octodns/processor/templating.py", line 102, in process_source_and_target_zones
    raise ValidationError(
    ...<4 lines>...
    )
octodns.record.exception.ValidationError: Invalid record "xazz.test.tld."
  - Invalid template parameter "zone_source_ids" in record value
```

ross:
> I’ve made a small [patch](https://github.com/user-attachments/files/21492643/RaiseValidationError.patch) to raise a ValidationError instead:

Mostly looks good. I'll probably tweak it slightly to use a different exception rather than ValidationError which is more about the adherence to DNS specs and best practices. PR in the works.

PR #1282, opened 2025-07-30
Add new provider parameter "root_ns_warnings" to disable root NS related warnings

jleroy:
Some people (like me) just don’t want to manage zones root NS records in OctoDNS for a number of reasons.
In this case, having multiple warnings displayed per zone can really pollute `octodns-sync` output and create a "warning fatigue" which can lead to missing important warnings.

This PR add support for a new `root_ns_warnings` parameter for providers, which default to `True`. When set to `False`, root NS related warnings aren’t logged.

Configuration example:
```yaml
providers:
  googlecloud:
    class: octodns_googlecloud.GoogleCloudProvider
    root_ns_warnings: False
```

### DIFF
--- a/.changelog/fb7aa3d7570d4ee1ae8afb93408bfff4.md
+++ b/.changelog/fb7aa3d7570d4ee1ae8afb93408bfff4.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add new provider parameter root_ns_warnings to disable root NS related warnings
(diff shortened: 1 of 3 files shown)


ross (commented):
Have seen`IgnoreRootNsFilter`, https://github.com/octodns/octodns/blob/main/octodns/processor/filter.py#L414-L431, it will filter them out of both the desired and existing which IIRCC short-circuits the warnings/exceptions. I could be mis-remembering though.

The desired case that should eval to false is 
https://github.com/octodns/octodns/blob/a89f6352068a3c735b4615a364c7b1166bbda0dd/octodns/provider/base.py#L152-L154

And the expected case it:
https://github.com/octodns/octodns/blob/a89f6352068a3c735b4615a364c7b1166bbda0dd/octodns/provider/base.py#L190-L191

If that works as i remember./expect this is probably more a case of missing documentation.

ross (approved):
> In this one a warning is logged if _no_ root NS record exists in desired zone.

Guess I'm mis-remembering and didn't catch the `not` in there.

Issue #1281, opened 2025-07-30
InvalidCodepoint error raised when ValidationError is called on an IDN FQDN including an underscore character


jleroy:
In my OctoDNS config file, I have the following zone:

```yaml
zones:
  ...
  dédé.fr.:
    sources:
      - config
    targets:
      - googlecloud
 ```

...who contains a CNAME record with an invalid value (missing trailing dot):

```yaml
---
test._domainkey:
  type: CNAME
  value: target.without.trailing.dot
```

When I run `octodns-sync`, I get the following traceback:

```python
Traceback (most recent call last):
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/octodns/idna.py", line 51, in idna_decode
    return _decode(name).replace('.pl-a-ce--hold-e--r.', '..')
           ~~~~~~~^^^^^^
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/idna/core.py", line 430, in decode
    s = ulabel(label)
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/idna/core.py", line 321, in ulabel
    check_label(label_bytes)
    ~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/idna/core.py", line 277, in check_label
    raise InvalidCodepoint(
        "Codepoint {} at position {} of {} not allowed".format(_unot(cp_value), pos + 1, repr(label))
    )
idna.core.InvalidCodepoint: Codepoint U+005F at position 1 of '_domainkey' not allowed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jleroy/.pyvenv/dns-configs/bin/octodns-sync", line 33, in <module>
    sys.exit(load_entry_point('octodns==1.12.0', 'console_scripts', 'octodns-sync')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/jleroy/.pyvenv/dns-configs/lib/python3.13/site-packages/octodns/cmds/sync.py", line 62, in main
    manager.sync(
    ~~~~~~~~~~~~^
        eligible_zones=args.zone,
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
[...]

PR #1285, opened 2025-08-05
Fix encoding and decoding of mixed idna fqdns

ross:
I'd previously looked at the `requests` library for its handling of idna, but landed on code that isn't actually what it uses for the core case. Dug into that deeper and found out that its using urllib3 for that. This round implements things based on how that works, though it only has an encode and I had to translate the process to decode.

```yaml
---
'*':
  type: A
  value: 1.2.3.4
foo._bar:
  type: A
  value: foo
zajęzyk:
  type: A
  value: 2.3.4.6
```

```console
octodns.record.exception.ValidationError: Invalid record "foo._bar.zajęzyk.pl.", ./config/zajęzyk.pl.yaml, line 6, column 3
  - invalid IPv4 address "foo"
```

/cc Fixes https://github.com/octodns/octodns/issues/1281
/cc @jleroy

### DIFF
--- a/.changelog/0452fdb43ba64ae99bffeef152a6b5f7.md
+++ b/.changelog/0452fdb43ba64ae99bffeef152a6b5f7.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Fix encoding and decoding of mixed idna fqdns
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci bump

jleroy:
Everything seems to works fine on my side. 👍

PR #1296, opened 2025-08-25
fix: links

blackdotraven:
some links where broken.
Removed `)` for all of them.


ross (approved):
Doh, errant `s///`

Thanks for the fix.

Issue #1303, opened 2025-09-25
A need for clarity of wildcards in the config file


quistian:
I am trying to use the following wildcard syntax, in order to catch certain zone names and have them sent to a different target than a default one.

e.g.

zones:
  privatelink.east.*:
    sources:
       - local-yaml
     targets:
       -  azure-east

   '*':
     sources:
        - local-yaml
     targets:
        - azure-west
 

However the wild card syntax e.g. does not sent the zone:

privatelink.east.openai.azure.net to the target: azure-east, rather the default azure-west.

In order to get a zone to go to the azure-eat target I have to explicitly name the zone. e.g.

zones:
   privatelink.east.openai.azure.net.:
       sources:
           - local-yaml
        targets:
           - azure-east

quistian:
Many thanks for the clarity. In the meantime the fqdn including the dot does the effective "closest match" which is what I want.
There are not too many fqdns in my case, so it works.

And clarification on the serial list of checks in the processors section. I assume the name,type,value passes as long as all of the tests yield True?

ross:
WIP PR to add advaced regex based wildcard support is over in https://github.com/octodns/octodns/pull/1304.

PR #1304, opened 2025-09-27
Backwards compatible _preprocess_zones that also supports regexes, needs more testing

ross:
* [ ] Add lots more testing
* [ ] Document the details of the new (and old) behavior

/cc https://github.com/octodns/octodns/issues/1303#issuecomment-3335029453 which lead to this :rabbit: :hole:


ross:
After sleeping on it I decided that rather than hack detection of names that are regexes it'd be cleaner to add a regex config key, this also allows adding glob style matching support. Anyway, will be clearer when the doc is written, but you can probably get the gist from the tests as things currently stand.

PR #1309, opened 2025-10-08
Add processor support to octodns-dump

### LINKED ISSUE #1308: Feature Request - octodns-dump command to honour processors' filters
## Environment

OS: Mac OS
Python: 3.12.11
octodns: 1.13.0
octodns-cloudflare: 1.0.0

## Feature Request

Currently, the `octodns-dump` command does not honour `filters` under [processors](https://octodns.readthedocs.io/en/latest/api/processors.html#processors).

```yaml
processors:
  type-filter:
    class: octodns.processor.filter.TypeAllowlistFilter
    allowlist:
      - CNAME

zones:
  '*':
    sources:
      - config
    targets:
      - cloudflare
    processors:
      - type-filter
```

## Motivation

It would be great to be able to dump only filtered records. This would help users work on only _relevant_ records.

ross:
The `octodns-dump` command now honors processors configured for zones, allowing users to filter and transform records during dump operations.

## Changes

- Added `_get_processors()` helper method to build and translate processor lists
- Updated `dump()` to apply processors after zone population, before dumping
- Updated `sync()` to use the new `_get_processors()` helper for consistency
- Added test coverage for processor application and error handling

## Example Usage

```yaml
processors:
  type-filter:
    class: octodns.processor.filter.TypeAllowlistFilter
    allowlist:
      - CNAME

zones:
  example.com.:
    sources:
      - config
    processors:
      - type-filter
    targets:
      - cloudflare
```

Users can now dump only filtered records, making it easier to work with specific subsets of DNS data.

Fixes #1308


ross:
First experiment with using Claude Code to address an issue, which is why I put the detailed comment in https://github.com/octodns/octodns/issues/1308#issuecomment-3381528717.

Mostly a success. It was able to get a working solution on its first try, but needed quite a bit of "code review" to get to something clean and simple that did things in a way that looked and felt like the code around it and avoided duplicating logic. Had to tell it that it could extract a helper method to avoid duplicating the processor lookup and translation code and otherwise give it suggested variable names to avoid some chunkiness. Was pretty impressed with the idea it came up with for testing the functionality. That bit didn't need any suggestions or help and it managed to find tests that broke some assumptions in the code that it then needed to fix.

Probably could have done it faster w/o Claude. Maybe some room for improvement with more details in the prompt/memory, but jury's still out for whether it makes sense for this sort of use-case.

PR #1312, opened 2025-10-12
Manager zone config handling refactoring and cleanup

ross:
## Summary

This PR refactors and cleans up the Manager's zone configuration handling to improve code organization and enable safer workflows for API integration:

* Moved `active_zones`, `active_sources`, and `active_targets` from method parameters to Manager constructor parameters, allowing filtering to be configured once at initialization rather than per-method-call
* Implemented `Manager._zones` property caching to eliminate redundant zone configuration processing while preserving validation
* Refactored zone preprocessing logic to be closer to where it's used rather than modifying `manager.config['zones']` in-place
* Added test coverage for dynamic zone config handling with subzones

These changes establish a cleaner API surface and more predictable state management, which are prerequisites for octodns-api integration.

## Test plan

* All existing tests pass
* New test coverage added for dynamic zone config with subzones scenario

/cc https://github.com/ross/octodns-api/issues/1


ross:
Nothing inherently wrong with this, but it's a bigger change that I'm looking to make atm. Will pick it or something like it back up if/when octodns-api moves forward.

PR #1318, opened 2025-10-17
processors: Add a simple TTL clamping processor

muelli:
This will come in handy for APIs not supporting TTL outside a certain range. The Spaceship API, for example, only allows TTLs in the range of 5..3600 but rewriting a whole zone for Spaceship only seems not as convenient as clamping the values as they flow through OctoDNS.

The code is coming from Claude. My prompt was:

    Write a simple OctoDNS processor that clamps TTL values.

I followed up with:

    AttributeError: 'TtlClampProcessor' object has no attribute 'log'
    Where should self.log come from?

And finally:

    Hm. I suspect the code is not running, somehow. It doesn't seem to be getting control. I
    put a raise in the process_record method but it doesn't fail. And I don't see any clamped
    values nor log output.

So it took three attempts to make it produce something useful.

### DIFF
--- a/octodns/processor/clamp.py
+++ b/octodns/processor/clamp.py
@@ -41,11 +41,7 @@ class TtlClampProcessor(BaseProcessor):
             )
         self.min_ttl = min_ttl
         self.max_ttl = max_ttl
-        self.log.info(
-            'TtlClampProcessor initialized: min=%ds, max=%ds',
-            self.min_ttl,
-            self.max_ttl,
-        )
+        self.log.info('__init__: min=%ds, max=%ds', self.min_ttl, self.max_ttl)
 
     def process_source_zone(self, desired, sources):
         """
@@ -58,7 +54,7 @@ class TtlClampProcessor(BaseProcessor):
         Returns:
             The modified zone
         """
-        self.log.debug('Processing source zone: %s', desired.name)
+        self.log.debug('process_source_zone: desired=%s', desired.name)
 
         for record in desired.records:
             original_ttl = record.ttl
@@ -66,7 +62,7 @@ class TtlClampProcessor(BaseProcessor):
 
             if clamped_ttl != original_ttl:
                 self.log.info(
-                    'Clamping TTL for %s (%s) %s -> %s',
+                    'process_source_zone: clamping TTL for %s (%s) %s -> %s',
                     record.fqdn,
                     record._type,
                     original_ttl,


ross (commented):
Looks reasonable. Cloudflare has a similar limit on ttl, https://github.com/octodns/octodns-cloudflare/blob/e7de32a75f716ca09e14be21b1a5ec4a9852a41e/octodns_cloudflare/__init__.py#L106C9-L106C16, though in its case there's also special meaning to the value of 1 which complicates things.

The main downside to implementing the clamping as a processor rather than in the provider itself is that it'll apply the changes to all targets and not just the one that needs it. That's probably fine, maybe even a good thing, but I thought I'd point it out. 

Will need tests before it can be 🚢ed

ross:
I think provider-specific processors should be a viable concept, but I wouldn't really know until I go and try and do it. The biggest issue may be that all providers would have to implement `*args, **kwargs` in their `__init__` methods to make use of them.

octodns/processor/clamp.py
```diff
@@ -0,0 +1,71 @@
...
+        super().__init__(id)
+        self.log = getLogger(
+            f'{self.__class__.__module__}.{self.__class__.__name__}'
+        )
+        if not min_ttl <= max_ttl:
+            raise TTLArgumentException(f'Min TTL {min_ttl} is not lower than max TTL {max_ttl}')
+        self.min_ttl = min_ttl
+        self.max_ttl = max_ttl
+        self.log.info(
+            f'TtlClampProcessor initialized: min={min_ttl}s, max={max_ttl}s'
```
ross: logging prefers placeholders rather than f-strings:

```suggestion
            f'__init__: min=%ss, max=%ss', min_ttl, max_ttx)
```

octodns/processor/clamp.py
```diff
@@ -0,0 +1,71 @@
...
+        Process records from source zone(s).
+
+        Args:
+            desired: Zone object containing the desired records
+            sources: List of source names
+
+        Returns:
+            The modified zone
+        """
+        self.log.debug(f'Processing source zone: {desired.name}')
```
ross:
```suggestion
        self.log.debug('process_source_zone: [rocessing source zone: %s', desired.name)
```

octodns/processor/clamp.py
```diff
@@ -0,0 +1,71 @@
...
+
+        for record in desired.records:
+            original_ttl = record.ttl
+            clamped_ttl = max(self.min_ttl, min(self.max_ttl, original_ttl))
+
+            if clamped_ttl != original_ttl:
+                self.log.info(
+                    f'Clamping TTL for {record.fqdn} ({record._type}): '
+                    f'{original_ttl}s -> {clamped_ttl}s'
+                )
```
ross:
```suggestion
                self.log.info(
                    'process_source_zone: clamping ttl for %s (%s): %ds -> %ds'
                    record.fqdn. 
                    record._type,
                    original_ttl,
                    clamped_ttlk
                )
```

ross (commented):
Pretty straightfoward and mostly makes sense. Comments/suggestions around around logging more like the rest of octoDNS. Suggestions may or may not be correctly formatted.

PR #1324, opened 2025-10-21
Fix venv activation not working when using a custom VENV_NAME

jleroy:
Setting `VENV_NAME` to a custom value broke some Bash scripts:

```sh
$ export VENV_NAME=/Users/jleroy/.pyvenv/octodns

$ ./script/format
./script/format: line 7: env/bin/activate: No such file or directory

$ ./.git_hooks_pre-commit
./.git_hooks_pre-commit: line 9: ./env/bin/activate: No such file or directory
```

This PR fixes this issue and move all duplicate code related to separate script named `common.sh`.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -15,10 +15,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           architecture: x64
-      - name: Install dependencies
+      - name: Setup Environment
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


jleroy:
The changelog check has failed because it doesn’t seems to be run inside a venv. Can I update it to make it so?

ross:
> The changelog check has failed because it doesn’t seems to be run inside a venv. Can I update it to make it so?

Not sure why it installs into the container and doesn't use a venv probably just what was implemented over in octodns/changelet. I guess it can change to work more like the others.

PR #1336, opened 2025-12-01
Fix slash_commands workflow, regex and target branch

ross:


ross:
/changelog create --type none Fix regex and target branch issues in slash_commands.yml

ross:
Still using old workflow. Will have to merge...

Issue #149, opened 2017-11-14
NS and SOA records


oryagel:
Hi,
General question, but especially relevant to OctoDNS users.

Should I match the NS and SOA records between the different providers we use?
Should I create the records in the YAML file and then sync them back to both providers or maybe these records can be ignored and we just need to add the DNS servers from both providers to the registrar?

Thanks

yzguy:
These are relevant issues:
https://github.com/github/octodns/issues/38
https://github.com/github/octodns/issues/39

I'd say having them in sync is a good idea, so you don't have differences between providers. However there are definitely differences between providers in their features, eg. Dyn you cannot remove their nameservers, but you can add some. Some providers do not allow you to change the SOA, eg DigitalOcean

The one SOA option I have seen that you cannot usually control is the Serial, so those would definitely be out of sync, but that's only really relevant to secondaries. If you're using OctoDNS you probably aren't using secondaries though.

If you're doing multi-provider, you would have your name servers from both providers in the registar, and configured in the providers, you would want both name servers, eg. Dyn in NS1, and NS1 in Dyn.

ross:
There are 2 WIP PRs (that have were created long ago) for SOA, https://github.com/github/octodns/pull/45,  and NS, https://github.com/github/octodns/pull/53, record management. The SOA one probably isn't going anywhere. I ran into too many problems with providers support of managing them and couldn't really come up with a cohesive plan for them. 

The NS record PR can probably ship with a bit of updating and care, though it's a VERY sensitive change and the behaviors of providers there are less than idea, some will let you add more NS records, but not remove their own etc. That usually isn't an issue, but could result in situations where the resulting values don't match what's configured.

oryagel:
Thanks!

I reviewed the issues you mentioned, but I still don't understand what the recommended configuration.
Currently we have a single provider. I was inspired by the post in the engineering blog and want to add a secondary provider and "DNS as code". We don't have secondaries, meaning that all the records for a specific domain are managed by the same provider. 
If it helps, I don't mind updating a few NS records manually. I'm not sure what you recommend regarding NS and SOA records and would appreciate clarification.

ross:
If you have multiple providers configured to serve your records it is best to have all of the name servers include in both the registrar and zone ROOT NS records.

github.com's registrar records:

```
otter:~ ross$ dig NS github.com. @ns1.p16.dynect.net.

[...]

PR #1350, opened 2026-02-01
Add SOA record support

JakobHaufe:
This more or less a resurrection of #45.

I have patches for octodns-bind and octodns-netbox-dns to add support there as well, but those are currently just in a "works on my machine" state. Nevertheless, I was already able to update a zone from netbox to both a zone file and a running bind.

I also intend to add support for pdns, if possible.

I have read #149 but I think it makes sense to add support. While some providers might not support management of SOA records, it's plain necessary for people self-hosting their infrastructure.

### DIFF
--- a/octodns/record/soa.py
+++ b/octodns/record/soa.py
@@ -1,0 +1,222 @@
+#
+#
+#
+
+
+from fqdn import FQDN
+
+from ..equality import EqualityTupleMixin
+from .base import Record, ValueMixin
+from .rr import RrParseError
+
+
+class SoaValue(EqualityTupleMixin, dict):
+    @classmethod
+    def parse_rdata_text(cls, value):
+        try:
+            mname, rname, serial, refresh, retry, expire, minimum = value.split(
+                ' '
+            )
+        except ValueError as ve:
+            raise RrParseError('Failed to split SOA rdata value: ' + str(ve))
+        if not FQDN(mname).is_valid:
+            raise RrParseError(f'Failed to parse mname {mname} as FQDN')
+        if not FQDN(rname).is_valid:
+            raise RrParseError(f'Failed to parse rname {rname} as FQDN')
+        try:
+            serial = int(serial)
+        except ValueError as ve:
+            raise RrParseError('Failed to parse serial: ' + str(ve))
+        try:
+            refresh = int(refresh)
+        except ValueError as ve:
+            raise RrParseError('Failed to parse refresh: ' + str(ve))
+        try:
+            retry = int(retry)
+        except ValueError as ve:
+            raise RrParseError('Failed to parse retry: ' + str(ve))
+        try:
+            expire = int(expire)
+        except ValueError as ve:
+            raise RrParseError('Failed to parse expire: ' + str(ve))
+        try:
+            minimum = int(minimum)
+        except ValueError as ve:
+            raise RrParseError('Failed to parse minimum: ' + str(ve))
+        return {
+            'mname': mname,
+            'rname': rname,
+            'serial': serial,
+            'refresh': refresh,
+            'retry': retry,
+            'expire': expire,
+            'minimum': minimum,
+        }
+
+    @classmethod
+    def validate(cls, data, _type):
+        reasons = []
+        try:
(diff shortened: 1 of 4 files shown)


octodns/record/soa.py
```diff
@@ -0,0 +1,215 @@
...
+        return new
+
+    def __hash__(self):
+        return hash(self.__repr__())
+
+    def _equality_tuple(self):
+        return (
+            self.mname,
+            self.rname,
+            self.serial,
```
ross: `serial` is going to be really complicated to handle cleanly for this type given that a lot of providers are going to do things "automatically" to it, e.g. auto-increment or date-of-last-change.  

Having it specified in the YAML would be enough to cause persistent changes for those providers that can never be applied if it's included in the computation of equality here.

For providers that will let you specify/manage a serial it needs to change any time a record in the zone is updated by octoDNS. That would likely need to be implemented with special handling in a processor.

Basically it should pretty much never be in the YAML and different providers will need different behaviors.

ross (commented):
I'm willing to consider this given that providers won't implement it by default and it'll mostly be ignored unless someone goes in and spends the time to figure out the details of what a provider needs to get it all right.

That said I think there's a couple shortcomings to the semantics/implementations. 

One mentioned inline is the handling of `serial` which is not straightforward and will pretty much never be able to match behavior when pushing to multiple providers.

The other is that it will likely need some special handling like ROOT NS so that if/when support is added to a provider it doesn't immediately start trying to delete the SOA records b/c people haven't previously included them in their YAML. I'd have to dig into the details of what all was implemented for root NS records, but the gist is that there's a warning and they're ignored if the target(s) support them, but there's not one in the sources (usually YAML, can't remember if source support comes into play). There's a warning if there's one in the sources and the target provider(s) don't support them. They're only managed if the provider supports them and there's one defined in the sources.

TL;DR unfortunately there's a lot of non-trivial details to implementing this beyond just defining the details of the type.

It also likely makes sense to create one or two provider PRs that use this before the PR shifts to make sure that the details are correct and that we don't have to come back with actually make things right PRs once it's merged and potentially seen/used by people.

ross:
> Thank you very much for the rapid review. I will adress the individual problems bit by bit. Should I set it to draft for the time being?

No problem. Draft is up to you. If you prefer that I not look at things again until they're "ready" you can move it to draft and I'll for the most part hold off until it's switched back. Otherwise doesn't make any difference to me.

octodns/record/soa.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @classmethod
+    def process(cls, value):
+        return cls(value)
+
+    def __init__(self, value):
+        super().__init__(
+            {
+                'mname': str(value['mname']),
+                'rname': str(value['rname']),
```
ross: Haven't looked into SOA records with utf-8 fqdns so I'm not sure, but if mname and/or rname need to support utf-8 values when configured in sources but translate them into the idna encoded version before pushing to providers these lines may need to use idna_encode, e.g.

https://github.com/octodns/octodns/blob/c4c90d3526ae28f47295a2a713892f13d5637aab/octodns/record/mx.py#L77

octodns/record/soa.py
```diff
@@ -0,0 +1,222 @@
...
+    def template(self, params):
+        if '{' not in self.mname + self.rname:
+            return self
+        new = self.__class__(self)
+        new.mname = new.mname.format(**params)
+        new.rname = new.rname.format(**params)
+        return new
+
+    def __hash__(self):
+        return hash(self.__repr__())
```
ross: In general other records use hash(tuple) for the fields that should be considered similar to _equality_tuple, e.g. 

https://github.com/octodns/octodns/blob/c4c90d3526ae28f47295a2a713892f13d5637aab/octodns/record/mx.py#L111-L112

ashleytayles:
Hey @JakobHaufe, this still something you're working on?

JakobHaufe:
@ashleytayles Not currently due to lack of time. I also got stuck on the octodns-bind side as I couldn't find out how to fix the tests after adding my changes. I'm still interested in having SOA support, though.

PR #1372, opened 2026-04-10
Validate SSHFP fingerprint length matches fingerprint_type

### LINKED ISSUE #1371: Add fingerprint length validation for SSHFP records
I recently ran into an issue where my DNS record returned `FORMERR` because of a simple copy-paste error in an `SSHFP` record.

I accidentally used a SHA-1 hash (40 characters) for a record defined as `fingerprint_type` 2 (SHA-256, which requires 64 characters).

Should octodns validate the length of the fingerprint based on the type?

ross:
## Summary
- Add length validation to `SshfpValue.validate`: SHA-1 (type 1) must be 40 hex chars, SHA-256 (type 2) must be 64, per RFC 4255/6594
- Check is gated on a known `fingerprint_type` and a present `fingerprint` so it doesn't stack redundant errors onto existing missing/invalid cases
- Updated a handful of test fixtures that were using arbitrary short fingerprints

Fixes #1371

## Test plan
- [x] `./script/test`
- [x] `./script/coverage` (100% branch coverage)
- [x] `./script/lint`
- [x] `./script/format --check`

### DIFF
--- a/.changelog/8fbdc64c7fb641c087dbcafe9bf2fcb4.md
+++ b/.changelog/8fbdc64c7fb641c087dbcafe9bf2fcb4.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Validate SSHFP fingerprint length matches fingerprint_type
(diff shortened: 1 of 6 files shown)


ross (commented):
/ci modules bump

ross:
Looks like there's a couple tests in providers that fail this validation. A quick spot check seems to show that they're just test data.

Issue #909, opened 2022-06-28
Publish a JSON Schema


sjparkinson:
I'm reviewing our VS Code extensions settings for an OctoDNS based repository.

It'd be amazing to get a JSON Schema published for the OctoDNS configuration.

The schema could be added to [SchemaStore/schemastore](https://github.com/SchemaStore/schemastore/tree/master/src/schemas/json) (published to https://www.schemastore.org/api/json/catalog.json), which is used by the [YAML extension](https://github.com/redhat-developer/vscode-yaml#extension-settings).

If there's support for this I'd be happy to look at raising a pull request.

sjparkinson:
Here's a first pass at a schema to add context:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://raw.githubusercontent.com/Financial-Times/dns/main/schema.json",
  "title": "OctoDNS zone file",
  "type": "object",
  "patternProperties": {
    ".*": {
      "properties": {
        "ttl": {
          "type": "integer",
          "minimum": 0
        },
        "type": {
          "type": "string",
          "enum": [
            "A",
            "AAAA",
            "ALIAS",
            "CAA",
            "CNAME",
            "DNAME",
            "LOC",
            "MX",
            "NAPTR",
            "NS",
            "PTR",
            "SPF",
            "SRV",
            "SSHFP",
            "TLSA",
            "TXT",
            "URLFWD"
          ]
        },
        "value": {
          "type": "string"
        },
        "values": {
          "type": "array",
          "items": {
            "type": "string"
          }
        }
      },
      "required": [
        "type"
      ],
      "anyOf": [
        {
          "required": [
            "value"
          ]
        },
        {
          "required": [
            "values"
          ]
        }
      ]
    }
  }
}
```

This doesn't deal with [multiple records](https://github.com/octodns/octodns/blob/master/docs/records.md#multiple-records).

ross:
Don't mind it conceptually, but the idea of schemas living divorced in a repo separate from the code that implements them seems like a bad. As soon as it's committed it'd start drifting out of sync.

sjparkinson:
> Don't mind it conceptually, but the idea of schemas living divorced in a repo separate from the code that implements them seems like a bad. As soon as it's committed it'd start drifting out of sync.

[...]

PR #1373, opened 2026-04-18
Generate JSON Schema for zone YAML files

ross:
## Summary

- Adds `build_zone_schema()` in a new `octodns.schema` module that walks `Record.registered_types()` and composes a JSON Schema Draft 2020-12 describing an octoDNS zone YAML file. Each value class exposes a `_schema()` classmethod so schema fragments live next to validation and new record types are forced to ship one.
- Adds an `octodns-schema` CLI (`--indent`, `--output`) that prints/writes the generated schema as JSON. Intended for IDEs (e.g. `redhat.vscode-yaml`), CI lint, and future SchemaStore submission.
- Adds `jsonschema` as a dev-only dependency and a test suite covering anti-drift (every registered type is present, valid JSON Schema), per-type valid/invalid cases for all 21 record types including dynamic/geo blocks, round-trip validation of zone fixtures under `tests/config/`, and CLI smoke tests.

octoDNS's imperative validation is unchanged — the schema is for external consumers, not internal error reporting (see history on #909 / #1272).

## Test plan

- [x] `./script/test tests/test_octodns_schema.py`
- [x] `./script/coverage` (100% branch coverage)
- [x] `./script/lint` and `./script/format --check`
- [x] `env/bin/octodns-schema | python -m json.tool | head` produces valid JSON
- [ ] Manual: point `yaml.schemas` at the generated file in VS Code and confirm autocomplete + inline errors on `tests/config/unit.tests.yaml`


ross (commented):
/ci modules bump

ross (commented):
This is all self contained so I'm good with shipping it as-is and iterating.

PR #1374, opened 2026-04-18
Register UriRecord and UriValue

ross:
Oversight back when the type was added. Including them in the octodns.record will ensure they're correctly loaded and registered.

### DIFF
--- a/.changelog/307ee403c7a54c5889257159c30c731a.md
+++ b/.changelog/307ee403c7a54c5889257159c30c731a.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Register UriRecord and UriValue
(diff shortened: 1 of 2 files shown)


jleroy:
@ross Is there any plan to publish a new release including this fix?

ross:
> @ross Is there any plan to publish a new release including this fix?

I have a series of PRs that I'm working through now that does a pretty major overhaul on the validators system, https://github.com/octodns/octodns/pull/1387, that will allow enabling/disabling record validations as well as adding custom validations. The end goal of that is to add support for full RFC based validations across all records as well as extend the details of the "best practice" style validation. Opt-in for now, default in 2.x. I've probably got another week or so on that after which I'll be looking to cut a release.

PR #1378, opened 2026-04-19
Refactor record and value validations into Validator classes

ross:
## Summary

Purely structural refactor of record and value validation — no behavior
changes. Existing `validate()` bodies on Record subclasses, mixins, and
value classes have been extracted into dedicated `RecordValidator` /
`ValueValidator` subclasses attached via a per-class `VALIDATORS` list.
Reason strings are preserved byte-for-byte; the full test suite and 100%
branch coverage pass unchanged.

This lays the groundwork for a follow-up that will make validations
extensible and configurable — users will be able to register custom
validators and enable/disable individual checks. That user-facing surface
is intentionally out of scope here.

## Design

Two small base classes in a new `octodns/record/validator.py`:

- `RecordValidator.validate(record_cls, name, fqdn, data)`
- `ValueValidator.validate(value_cls, data, _type)`

Each Record subclass / mixin / value class declares its own `VALIDATORS`
list. Composition still happens via the existing MRO `super().validate()`
chain — each layer iterates its own `VALIDATORS` (referenced by explicit
class name to avoid MRO shadowing) and chains to `super()`. Granularity is
one validator per value type, and record-level checks split along logical
concerns (name shape vs. TTL vs. healthcheck, etc.).

## Process

Work landed as three separately-reviewable commits on this branch:

1. **Extract Record-level validations into validator classes** — introduce
   `RecordValidator` / `ValueValidator` base classes and migrate
   `Record.validate()` (name / TTL / healthcheck) as the reference
   implementation.
2. **Migrate record-level validations to RecordValidator classes** —
   `_DynamicMixin`, `_GeoMixin`, and per-type overrides on SRV, CNAME,
   ALIAS, and URI.
3. **Migrate value-level validations to ValueValidator classes** — 15 value
   classes: CAA, chunked, DS, IP, LOC, MX, NAPTR, OPENPGPKEY, SRV, SSHFP,
   SVCB/HTTPS, target, TLSA, URI, URLFWD.

## Test plan

- [x] `./script/test` — 491 tests pass
- [x] `./script/coverage` — 100% branch coverage
- [x] `./script/lint` — clean
- [x] `./script/format --check` — clean
- [x] New direct unit tests for each validator class added alongside the
      migrations; existing tests untouched so equality of reason strings is
      enforced.

### DIFF
--- a/.changelog/c490fd3328724cd692afffa943e43fe5.md
+++ b/.changelog/c490fd3328724cd692afffa943e43fe5.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Introduce RecordValidator/ValueValidator base classes, laying groundwork for extensible record and value validations
--- a/.changelog/c490fd3328724cd692afffa943e43fe5.md
+++ b/.changelog/c490fd3328724cd692afffa943e43fe5.md
@@ -1,4 +1,4 @@
 ---
 type: minor
 ---
-Introduce RecordValidator/ValueValidator base classes, laying groundwork for extensible record and value validations
+Introduce RecordValidator/ValueValidator base classes, laying groundwork for extensible record and value validations. Overriding `validate` on a `Record` subclass or defining a `validate` classmethod on a value class is DEPRECATED; declare validators via the `VALIDATORS` class attribute instead. Will be removed in 2.0.
--- a/octodns/record/alias.py
+++ b/octodns/record/alias.py
@@ -12,6 +12,11 @@ class AliasValue(_TargetValue):
 
 
 class AliasRootValidator(RecordValidator):
+    '''
+    Restricts ALIAS records to the zone root — the non-standard ALIAS
+    type only has meaning at the apex.
+    '''
+
     @classmethod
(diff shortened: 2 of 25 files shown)


ross (commented):
I think my biggest feedback here is that I'm not a fan of all the subclasses having to implement `def validate` if they have any validators. I think I'd like to explore cleaning that up.

To do it cleanly in a backwards compatible way (that supports 3rd party records and their existing validations)  I suspect we'd need to add a new method, something like `Record._process_validators` that walks the class heirarchy running all the validators that's called after the existing `validate` and then remove all of the methods from the child classes. Part I'm not sure about is how that would fit in with ValueValidators. Perhaps those mixins would need to override `_process_validators` and call up to super like they do for `validate`

ross (commented):
I think I'm happy with this. 

For context I'm planning to make a pass through things to get validations much closer to RFC and best practices (over time.)  I want to do that in a way where it's opt-in for the 1.x series and default-on in 2.x with the abilty to customize what is/isn't run.

PR #1379, opened 2026-04-23
Add opt-in strict SRV validators per RFC 2782 and RFC 6335

ross:
## Summary

- Adds `SrvStrictNameValidator` enforcing RFC 2782 `_service._proto` name structure with RFC 6335 §5.1 service-name syntax for both labels.
- Adds `SrvStrictValueValidator` enforcing 0-65535 ranges, the "target `.` => priority/weight/port all 0" convention from RFC 2782, and rejection of port 0 when target is not `.`.
- Existing validators are unchanged. The new validators live alongside the existing ones and are **not** registered in `SrvRecord.VALIDATORS` / `SrvValue.VALIDATORS`; callers opt in by appending them.

## Test plan

- [x] `./script/coverage` — 499 passed, 100% branch coverage
- [x] `./script/lint`
- [x] `./script/format --check`
- [x] New tests directly exercise both strict validators and verify they're not in the default `VALIDATORS`.
- [x] End-to-end test wires the strict validators onto the classes and validates they surface through `Record.new`.

/cc https://github.com/octodns/octodns/pull/1387


ross (commented):
Looks good.

PR #1380, opened 2026-04-24
Make validators configurable and support third-party registration

ross:
## Summary

- Converts all built-in validators from class-static methods to instances with stable `id` attributes, enabling them to be referenced by name
- Introduces a unified validator registry process, `RecordType.VALIDATORS` and `ValueType.VALIDATORS` for automatic registration and `register_validator` / `unregister_validator` / `registered_validators` classmethods for programatic access.
- Wires a new top-level `validators:` config section and `manager.validators` / `manager.disable_validators` mappings into `Manager`, allowing validators to be added or disabled per record type (or globally) from the main config file
- Adds documentation in `docs/configuration.rst` covering built-in validator ids, config-driven add/disable, and programmatic registration

Builds on #1378.

## Test plan

- `./script/test` — 512 tests, all passing
- `./script/coverage` — 100% branch coverage
- `./script/lint` / `./script/format --check` — clean

### DIFF
--- a/.changelog/0e340f8428c9404d8371303a268453c7.md
+++ b/.changelog/0e340f8428c9404d8371303a268453c7.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Dynamic validator registration and configuration based management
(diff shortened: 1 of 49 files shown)


ross (commented):
/ci modules bump

ross (commented):
/ci modules bump

PR #1381, opened 2026-04-26
Add `ignore_subzone_adds` zone option

fmcglinn:
## Rationale

Some sources return records hierarchically — a parent zone's source emits everything under it, including records that belong to a configured child zone.
Today these overlapping records cause `Zone.add_record` to raise `SubzoneRecordException`, and the only workaround is `lenient: true` — but lenient is a rather large hammer.

`Zone.ignore_subzone_adds`
A per-zone YAML flag that silently drops sub-zone overlaps with no impact on existing leniency semantics.

## Usage

```yaml
zones:
  parent.example.:
    ignore_subzone_adds: true
    sources:
      - some-source
    targets:
      - some-target

  child.parent.example.:
    sources:
      - some-source
    targets:
      - some-target
```

When enabled on the parent zone, records that fall under a configured sub-zone are silently dropped at `Zone.add_record` (logged at `debug`). Without `--debug`, no warnings or exceptions are raised.

## Implementation

- `Zone.__init__` accepts `ignore_subzone_adds=False` and stores it.
- `Zone.add_record` gets an `elif self.ignore_subzone_adds` branch at each of the two existing sub-zone raise sites, between the lenient warn-path and the raise.
- `Manager.get_zone` reads the YAML key and passes it through, mirroring the existing `update_pcent_threshold` / `delete_pcent_threshold` plumbing.

Records explicitly marked `octodns: { lenient: true }` keep their existing warn-and-add behavior. NS and DS at the sub-zone boundary remain unaffected — record-type guard runs before either new branch is reached.

## Tests

- `tests/test_octodns_zone.py::test_ignore_subzone_adds` — both raise sites, NS/DS at boundary, record-level and call-level lenient precedence, records outside any sub-zone unaffected, default value.
- `tests/test_octodns_manager.py::test_zone_ignore_subzone_adds` — YAML key plumbing through `Manager.get_zone`, including default behavior when the key is omitted. Uses a new `tests/config/zone-ignore-subzone-adds.yaml` fixture.

100% statement coverage on `octodns/zone.py` and `octodns/manager.py` after the change.

## Initial Proposal
Initially proposed using a processor, however after feedback it was reworked.
- **(commits 88ef769, 76b3ef1) — `SubzoneOverlapFilter` processor.** A zero-config processor that mirrored `Zone.add_record`'s sub-zone rules and removed overlapping records after-the-fact. Reverted in commits d2dba62 and 85dd870.

## Related

- #1353 — added warnings for sub-zone records under lenient.
- #1362 / #1363 — established record-level lenient as the opt-in for legitimate sub-zone-overlap records (glue, etc.); this PR's precedence rules respect that opt-in.


ross:
Overall seems reasonable. The other approach I can think of would be to add support to `Zone` to `ignore_subzone` or something like that which would prevent them from being added in the first place, and I guess avoid the warnings, maybe downgrading or altogether.

Will need to sit down and dig into this a bit deeper. May be a couple days before I get there.

fmcglinn:
No worries - let me know if you have a preference, happy to rework the PR either way :)

ross:
Hi @fmcglinn, got a chance to sit down. I think I landed 51% towards the `Zone.ignore_subzone_adds`, basically b/c it'll avoid the need for `lenient` which will eat a lot of other stuff and still result in a lots of warnings being printed to the logs. 

My quick scan of what would be needed is to add it as a new `Zone.__init__` param and then add a couple of `elif self.ignore_subzone_adds`s to `Zone.add_record` that log.debug, `{msg}, ignore.` just before the two points that currently raise.

The only other bit would be plumbing the value through from the config data i.e.:

```yaml
zones:
  foo.com.:
    ignore_subzone_adds: true
    sources: ...
    targets: ...
```

with a bit of code in `Manager.get_zone` to look for the key and pass it down into the `Zone` object.

Lmk if there's any questions or if you have thoughts otherwise. 

Thanks for the PR and work thus far.

kabenin:
@ross, with the approach you lean towards to, will that still work if my run is limited to foo.com., and does not include it's subzones? It seems like this will be an issue with the filter suggested in this pr. 
I think this filter/zone parameter is very valuable, I have usecases for that as well. I think it would be great to make it stable regardless of limiting zones vs running with the full set of zones.

ross:
> @ross, with the approach you lean towards to, will that still work if my run is limited to foo.com., and does not include it's subzones? It seems like this will be an issue with the filter suggested in this pr.

Limiting the zones being sync'd should have an impact on the configuration of the Zone objects. So long as they're in the config file the Zone object will still get the same sub zone list during a limited sync that it gets in a full sync. The config doesn't change, just what zones are processed.

fmcglinn:
Thanks @ross for the feedback - i've updated the PR with the proposed changes.
This is far more elegant solution that what I'd proposed - no concerns or issues with the use-case I had :)

octodns/zone.py
```diff
@@ -174,6 +175,14 @@ def __init__(
         :param delete_pcent_threshold: Override for maximum delete percentage
                                        threshold. If None, uses provider default.
         :type delete_pcent_threshold: float or None
+        :param ignore_subzone_adds: If True, silently drop records that belong
+                                    under a configured sub-zone instead of
+                                    raising. Useful when a source returns
+                                    records that overlap a configured
+                                    sub-zone. Records explicitly marked
+                                    ``lenient=True`` keep their existing
+                                    warn-and-add behavior.
```
ross, line 184: Nice catch on the nuance of `lenient` still logging at warn.

ross (approved):

ross:
Thanks.

PR #1387, opened 2026-04-30
Add named validator sets and manager.enabled config key

ross:
## Summary

- Introduces named validator *sets* (`legacy`, `rfc`, `best-practice`, …) so operators can opt into a curated bundle with a single config key rather than naming every validator individually
- Adds `manager.enabled` config key (default `['legacy']`) that drives which sets are active; preserves existing behaviour when omitted
- Two-registry model: validators register into an *available* registry at import time; `Manager._configure_validators` copies matching validators into the *active* registry based on `manager.enabled`
- `sets=None` on a validator means always-active regardless of `manager.enabled` — right default for third-party validators
- All built-in validators explicitly tagged `sets={'legacy'}`; bridge validators (`_`-prefixed) naturally get `sets=None`
- `Record.enable_validators(sets)`, `Record.enable_validator(id, types)`, and `Record.disable_validator(id, types)` replace the old `unregister_validator` API
- Docs updated: configuration.rst validator section rewritten; `octodns.record.validator` added to API reference


ross (commented):
/ci modules bump

ross (commented):
/ci modules bump w/fallback install

ross (commented):
/ci modules bump

ross (commented):
I think we're ready on this. One final round of modules CI and :shipit:

PR #1389, opened 2026-05-01
Normalize hex data fields to lowercase for case-insensitive comparison

### LINKED ISSUE #1388: TLSA record "certificate association data" field must be treated as case-insensitive
I have the following TLSA records in a zone file:

```yaml
_443._tcp.www:
  - type: TLSA
    value:
      - certificate_association_data: 24049AA6E03051A0DF3AEFBBFAD4686E6207DD4E601858EE40C31C8B0BD6BC03
        certificate_usage: 3
        matching_type: 1
        selector: 1
      - certificate_association_data: 31EF2A4D6E285CC29A636C5171F7DA0AC69CC44CEBAF5CD039DA8CC81187482A
        certificate_usage: 3
        matching_type: 1
        selector: 1
```

When I apply changes, octoDNS keep asking me to update the records: 

```
* example.com.
* opusdns (OpusDNSProvider)
*   Update
*     <TlsaRecord TLSA 3600, _443._tcp.example.com., [''3 1 ''1 24049aa6e03051a0df3aefbbfad4686e6207dd4e601858ee40c31c8b0bd6bc03'', ''3 1 ''1 31ef2a4d6e285cc29a636c5171f7da0ac69cc44cebaf5cd039da8cc81187482a'']> ->
*     <TlsaRecord TLSA 3600, _443._tcp.www.example.com., [''3 1 ''1 24049AA6E03051A0DF3AEFBBFAD4686E6207DD4E601858EE40C31C8B0BD6BC03'', ''3 1 ''1 31EF2A4D6E285CC29A636C5171F7DA0AC69CC44CEBAF5CD039DA8CC81187482A'']> (config)
*   Summary: Creates=0, Updates=1, Deletes=0, Existing=10, Meta=False
```

My provider’s API returns the hash value of the "certificate association data" field in lowercase, which is causing the issue.
I see nothing in [RFC 6698](https://datatracker.ietf.org/doc/html/rfc6698#section-2.2) regarding case-sensitivity, so according to [RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.3) comparison must be case-insensitive?

Also there’s single quotes **and** double quotes in the string returned by `TlsaValue.__repr__()`:
https://github.com/octodns/octodns/blob/ea06f3a9ce2baf050283c22be035df800490487d/octodns/record/tlsa.py#L192-L196

ross:
## Summary

- TLSA `certificate_association_data`, SSHFP `fingerprint`, and DS `digest` are hex data fields that are case-insensitive by nature; normalizing to lowercase on store prevents spurious changes when a provider returns the same value in a different case
- Also fixes a `TlsaValue.__repr__` bug that produced two adjacent single-quoted strings (`'1 0 ''0 abcdef...'`) instead of a single quoted string

Follows the same pattern as #1360 (NAPTR flags normalization).

Fixes #1388
Related #1360

### DIFF
--- a/.changelog/23088495a46f46458e4cf5ed967f7ead.md
+++ b/.changelog/23088495a46f46458e4cf5ed967f7ead.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Normalize hex data fields (TLSA certificate_association_data, SSHFP fingerprint, DS digest) to lowercase for case-insensitive comparison
--- a/octodns/record/ds.py
+++ b/octodns/record/ds.py
@@ -146,14 +146,14 @@ class DsValue(EqualityTupleMixin, dict):
                 'key_tag': int(value['flags']),
                 'algorithm': int(value['protocol']),
                 'digest_type': int(value['algorithm']),
-                'digest': value['public_key'],
+                'digest': str(value['public_key']).lower(),
             }
         else:
             init = {
                 'key_tag': int(value['key_tag']),
                 'algorithm': int(value['algorithm']),
                 'digest_type': int(value['digest_type']),
-                'digest': value['digest'],
+                'digest': str(value['digest']).lower(),
             }
         super().__init__(init)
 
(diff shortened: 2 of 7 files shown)


ross (commented):
/ci modules bump

PR #1390, opened 2026-05-03
Add new strict validators set and include existing rfc-compliant validators in it 

ross:
This adds/extends two new sets, `strict` and `best-practice`. The first include RFC compliance checks and the latter includes general best practice validation, e.g. trailing `.`. 


ross (commented):
This should be a complete no-op for existing setups and only impact things that opt into the new behavior.

PR #1391, opened 2026-05-04
Add best-practice validators for SSHFP, DS, TLSA, and CAA

ross:
## Summary

- `sshfp-value-best-practice` — warns when `fingerprint_type 1` (SHA-1) is used; SHA-256 (type 2) is the modern standard
- `ds-value-best-practice` — warns on deprecated signing algorithms (1/RSA-MD5, 3/DSA, 5/RSA-SHA1, 6/DSA-NSEC3, 7/RSASHA1-NSEC3 per RFC 8624 §3.1) and `digest_type 1` (SHA-1, NOT RECOMMENDED per RFC 8624 §3.3)
- `tlsa-value-best-practice` — warns when `matching_type 0` (full DER-encoded data) is used; RFC 7671 §4.1 advises against it because cert renewals require a DNS update before the new cert can be used
- `caa-value-best-practice` — warns when an `issue` tag is present without a corresponding `issuewild`; per RFC 8659 §4.2, omitting `issuewild` leaves wildcard certificate issuance policy implicit rather than explicit

All four are opt-in via the `best-practice` validator set (or individually via `Record.enable_validator`) and are not active by default. Each ships with 100% branch coverage.

## Test plan

- [x] `./script/coverage` passes at 100%
- [x] Direct validator class tests covering valid, invalid, and edge-case inputs
- [x] Opt-in integration tests via `Record.enable_validator` / `Record.disable_validator`
- [x] `test_best_practice_not_in_defaults` asserting each validator is absent from the default active set

### DIFF
--- a/.changelog/b75fb51a1bf948ba8d56c2a2b922ba35.md
+++ b/.changelog/b75fb51a1bf948ba8d56c2a2b922ba35.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add best-practice validators for SSHFP, DS, TLSA, and CAA record types
--- a/octodns/record/caa.py
+++ b/octodns/record/caa.py
@@ -73,12 +73,42 @@ class CaaValueRfcValidator(ValueValidator):
         return reasons
 
 
+class CaaValueBestPracticeValidator(ValueValidator):
+    '''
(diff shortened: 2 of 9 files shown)


ross (commented):
/ci modules bump

PR #1396, opened 2026-05-07
Add ZoneValidator framework for cross-record zone-level validation

ross:
## Summary

- Adds a `ZoneValidator` framework that mirrors `RecordValidator`/`ValueValidator` but operates on a fully assembled `Zone`, enabling cross-record DNS best-practice checks that per-record validators cannot perform
- Ships two built-in validators in a new `best-practice` set (off by default): `multi-value-mx` (every MX has ≥2 values) and `apex-spf-presence` (apex has an SPF TXT)
- All validator configuration is now nested under `manager.validators` (`.enabled`, `.record.*`, `.zone.*`); the old flat keys are removed and a deprecated fallback (`manager.validators.validators` / `.disable_validators`) is provided for the record-validator rename
- `Zone.validate(lenient=False)` is called in `_populate_and_plan`, `dump`, and `validate_configs`; `octodns-validate --all` picks up zone warnings via the `Zone` logger
- `octodns/zone/` is restructured as a package: `zone/__init__.py` (Zone + zone exceptions), `zone/exception.py` (ZoneException, ValidationError), `zone/validator.py` (registry, base class, built-ins)

## Test plan

- [x] `./script/coverage` — 710 tests, 100% branch coverage
- [x] `./script/lint` and `./script/format --check` pass
- [x] Opt in to built-ins via `manager.validators.enabled: [legacy, best-practice]` and confirm `multi-value-mx` / `apex-spf-presence` fire on a zone missing those records
- [x] Confirm `octodns-validate --all` exits non-zero when zone validators warn in lenient mode

/cc #1380


ross (commented):
/ci modules bump

PR #1398, opened 2026-05-09
Migrate CNAME coexistence check to CnameCoexistenceValidator and sub-zone check to SubzoneRecordValidator

ross:
This PR standardizes zone-level validation by migrating inline checks from `Zone.add_record` to dedicated `ZoneValidator`s. Validation is now deferred to the `zone.validate()` phase.

Key changes:
- **CNAME Coexistence**:
  - Migrated CNAME coexistence check from `Zone.add_record` to `CnameCoexistenceValidator`.
  - Removed the now unused `InvalidNodeException`.
- **Sub-zone Overlaps**:
  - Migrated `SubzoneRecordException` checks to `SubzoneRecordValidator`.
  - Removed `SubzoneRecordException` class.
  - Refactored `Zone.add_record` to retain `ignore_subzone_adds` logic while deferring validation to the validator.
- **General**:
  - Registered new validators in `legacy` and `strict` sets.
  - Updated all affected tests to reflect the new validation model.
  - Ensured 100% code coverage and lint compliance.

### DIFF
--- a/tests/test_octodns_zone.py
+++ b/tests/test_octodns_zone.py
@@ -18,13 +18,8 @@ from octodns.record import (
     Record,
     Update,
 )
-from octodns.zone import (
-    DuplicateRecordException,
-    InvalidNameError,
-    InvalidNodeException,
-    SubzoneRecordException,
-    Zone,
-)
+from octodns.zone import DuplicateRecordException, InvalidNameError, Zone
+from octodns.zone.exception import ValidationError
 
 
 class TestZone(TestCase):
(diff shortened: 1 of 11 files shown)


tests/test_octodns_zone.py
```diff
@@ -612,25 +612,25 @@ def test_cname_coexisting(self):
...
-        with self.assertRaises(InvalidNodeException) as ctx:
-            zone.add_record(cname)
-        self.assertIn(', has some context', str(ctx.exception))
-        self.assertEqual(set([a]), zone.records)
-        zone.add_record(cname, lenient=True)
+        zone.add_record(cname)
         self.assertEqual(set([a, cname]), zone.records)
+        # but fails validation
+        with self.assertRaises(ValidationError) as ctx:
+            zone.validate()
```
ross, line 628: This is a known/accepted change. Some of what used to happen in add_record will now happen here. It is expected that zone's will always have their validation called (by core octoDNS code) after add_record'ing is complete. That should hold true unless people are poking at the guts of octoDNS in unknown/unexpected ways. If they are the fix is simple enough...

ross (commented):
Verified the behavior locally. Both zone-level and record level `lenience` will convert the errors to warnings as expected.

ross (commented):
/ci modules bump

PR #1399, opened 2026-05-10
Implement NS-related zone validators

ross:
This PR adds GlueForInZoneNsZoneValidator (strict) and MultiValueApexNsZoneValidator (best-practice) to enforce circular dependency and redundancy checks for NS records.

/cc https://github.com/octodns/octodns/pull/1396


ross (commented):
/ci modules bump

PR #1397, opened 2026-05-08
Implement comprehensive suite of ZoneValidators (Strict & Best Practice)

ross:
This PR implements a collection of new `ZoneValidator` classes to improve DNS zone integrity and adherence to best practices.

### RFC Compliance (`strict` set)
- ~~**`no-cname-loop`**: Detects circular redirection chains (RFC 1034).~~
- ~~**`glue-for-in-zone-ns`**: Ensures address records exist for in-zone nameservers (RFC 1033).~~
- ~~**`single-spf`**: Validates that at most one SPF policy is published (RFC 7208).~~
- **`no-self-referencing-target`**: Detects records pointing to their own FQDN.
- ~~**`ns-target-not-cname`**: Prohibits NS targets from being CNAMEs (RFC 2181).~~
- ~~**`mx-target-not-cname`**: Prohibits MX exchanges from being CNAMEs (RFC 5321).~~
- ~~**`srv-target-not-cname`**: Prohibits SRV targets from being CNAMEs (RFC 2782).~~
- **`apex-ns-presence`**: Ensures root NS records exist (RFC 1034).

### Best Practices (`best-practice` set)
- ~~**`apex-dmarc-presence`**: Encourages DMARC publishing (RFC 7489).~~
- ~~**`apex-spf-presence`**: Encourages SPF publishing (RFC 7208).~~
- ~~**`consistent-ttl-at-name`**: Prevents cache skew by ensuring uniform TTLs at a hostname.~~
- ~~**`cname-target-resolvable-in-zone`**: Catches dangling internal CNAME/ALIAS references.~~
- ~~**`multi-value-apex-ns`**: Encourages root NS redundancy.~~
- ~~**`overlapping-subzone`**: Detects shadowed records existing below a delegation boundary.~~
- ~~**`apex-caa-presence`**: Encourages use of CAA records (RFC 8659).~~
- ~~**`mx-target-resolvable-in-zone`**: Verifies address records for in-zone MX targets.~~
- ~~**`srv-target-resolvable-in-zone`**: Verifies address records for in-zone SRV targets.~~

### Verification
- 100% statement and branch coverage in `octodns/zone/validator.py`.
- All tests passed.
- Formatting and linting checks are clean.

PR #1401, opened 2026-05-11
Add NoCnameLoopZoneValidator

ross:
## Summary

- Ports `NoCnameLoopZoneValidator` from the experimental PR #1397 into the established zone-validator structure
- Detects circular CNAME/ALIAS chains within a zone (RFC 1034 §3.6.2); e.g. `a → b → c → a`
- Lives in its own module `octodns/zone/cname_loops.py`, auto-registered at import time via `Zone.register_zone_validator`
- Returns `ValidationReason` objects carrying the in-zone records on the cycle so `lenient` and zone `context` handling work correctly
- Registered into the `strict` set only (`sets={'strict'}`)

## Test plan

- [x] 5 unit tests covering: no-loop pass, direct self-loop, 3-node indirect cycle, ALIAS-in-loop, and the merging-chains early-break path
- [x] `./script/coverage` — 694 tests, 100% branch coverage
- [x] `./script/lint` and `./script/format --check` pass

/cc #1396 — this PR targets the `zone-validators` branch from that PR; both need to land together
/cc #1397 — source of the original `NoCnameLoopZoneValidator` implementation this ports

### DIFF
--- a/.changelog/9c95d5342c5f45259755b9c3007ee654.md
+++ b/.changelog/9c95d5342c5f45259755b9c3007ee654.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add NoCnameLoopZoneValidator to detect circular CNAME/ALIAS chains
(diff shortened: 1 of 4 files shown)


ross (commented):
/ci modules bump

PR #1402, opened 2026-05-12
Handle null values in TXT/SPF records gracefully

ross:
Fixes crashes when YAML zone files contain TXT or SPF records with `null` values (`value:` with no content, or `null` entries in a `values:` list). These inputs previously raised `TypeError` or `AttributeError` deep in the loader/validator instead of producing a clean `ValidationError`.

## Changes

- **`octodns/record/chunked.py`** — `ChunkedValueValidator.validate` now detects `None` entries in the values list and appends `'missing value(s)'` rather than crashing on the regex search.
- **`octodns/record/base.py`** — `ValuesTypeValidator.validate` maps a scalar `None` value to `[]` so `value: null` is treated identically to a missing value.
- **`octodns/provider/yaml.py`** — `_populate_from_file` guards the `escaped_semicolons=False` escape pass so it skips `.replace()` on `None` values.
- **`octodns/record/validator.py`** — `process_values` was missing the same auto-enable-legacy-set guard that `process_record` has, causing direct calls to `_process_value_validators` to silently return `[]` before any `Record.new()` had triggered initialization. Fixed as a pre-existing bug.

## Tests

- New fixture directory `tests/config-txt-none/` with three zone YAML files covering scalar null, mixed list, and all-null list cases.
- `TestYamlProvider.test_txt_none_values` — loads all three fixtures with both `escaped_semicolons=True` and `=False`, asserting clean `ValidationError` in each case.
- Extended `TestChunkedValue.test_validate` with unit-level assertions for `[None]`, `[None, None]`, and `[None, 'foo', None]`.

### DIFF
--- a/.changelog/8e4da8e4367346fb8bdac252af524659.md
+++ b/.changelog/8e4da8e4367346fb8bdac252af524659.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Handle null values in TXT/SPF records gracefully, raising ValidationError instead of TypeError/AttributeError
--- a/octodns/provider/yaml.py
+++ b/octodns/provider/yaml.py
@@ -351,11 +351,16 @@ class YamlProvider(BaseProvider):
                             'SPF',
                             'TXT',
                         ):
-                            if 'value' in d:
+                            if 'value' in d and d['value'] is not None:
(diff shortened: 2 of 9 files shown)


ross (commented):
/ci modules bump

PR #1406, opened 2026-05-15
Add support for using alias zones with wildcards

ross:
This PR adds support for the `alias` directive in dynamic zone configurations (e.g., `*`).

Currently, dynamic zones require `sources` to discover which zones to configure via `list_zones()`. When using an `alias`, `sources` is typically omitted. This change allows dynamic zones to fall back to the `sources` of the aliased zone for discovery.

Additionally, it fixes a potential crash when `eligible_sources` filters out all sources for a dynamic zone.

Includes tests and ensures 100% coverage.

### DIFF
--- a/.changelog/dc53f775a1d4447f82d73bf747eefdc2.md
+++ b/.changelog/dc53f775a1d4447f82d73bf747eefdc2.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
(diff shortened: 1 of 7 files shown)


ross:
This doesn't actually make sense. You'd have to do back-flips to get a list of zones defined if they're fully aliased.

PR #1408, opened 2026-05-15
Add consistent-ttl-at-name zone validator

ross:
Ports the consistent-ttl-at-name validator from PR #1397 into the current ZoneValidator system. When multiple records exist at the same name (node) they must share the same TTL. The validator is registered with the 'strict' set.

- octodns/zone/ttl.py: new ConsistentTtlAtNameZoneValidator
- octodns/zone/__init__.py: export the new validator
- tests/test_octodns_zone_validators_ttl.py: dedicated test coverage

/cc https://github.com/octodns/octodns/pull/1396

### DIFF
--- a/.changelog/5ab7a2a80439432ea6228c760e3e7990.md
+++ b/.changelog/5ab7a2a80439432ea6228c760e3e7990.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add consistent-ttl-at-name zone validator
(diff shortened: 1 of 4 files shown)


ross:
Closing this as I think Claude or Gemini got things a bit confused when it originally came up with the recommended validation. Within a RRSets a single TTL value has to be used, but octoDNS has a different schema (to make things less opaque) and there's a single TTL at the Record level and not one per value so that isn't an issue.

Issue #3, opened 2017-03-16
Enhanced record validation 


ross:
Refactor validation logic to be a method on each record type and to accept a flags parameter to indicate how thoroughly to validate things beyond what's currently there.

Things like:
* ~~Ensure `values` are unique~~
* [ ] cname targets, mx, srv, etc. hostnames are resolvable
* [ ] maybe ip addresses are routable, pingable, http/https able... though not clear how it'd know what to try
* [x] `CNAME` can't coexist

ross:
* `CNAME` aren't allowed to co-exist alongside other records

ross:
> Ensure values are unique

That doesn't make sense. There are valid cases where you want/need to do so.

> MX & CNAME can't coexist

Done in https://github.com/github/octodns/pull/74. Also reject `CNAME`s on the root node.

ross:
> cname targets, mx, srv, etc. hostnames are resolvable
>  maybe ip addresses are routable, pingable, http/https able... though not clear how it'd know what to try

I've decided these 2 don't make sense as "record" validations, but I would like to build tooling, e.g. `octodns-validate`, that would walk a zone and look for these types of problems. I think that's a much cleaner way to go about this. The work involved/details of doing this stuff would be too heavyweight during a normal octodns-sync/dump/etc.

brianeclow:
I just ran into this:
 - A `cnames` cannot point to itself.

A DNS provider happily allowed us to create this condition via the web ux, however, route53 and dyn rejected it upon attempting to create the entries via octodns with the `--doit` flag.

o-be-one:
Please also add more logs/verbosity for errors. I had to find my answers by checking other issues.
Like for "out of key" ... Or "Invalid record domain.tld., missing type" ... It's not evident at first from where the issue is. Probably you can give more informations regarding the error or give a link to GitHub issue, something like that eheh :). It will help people less friendly with Python to understand issues.

yzguy:
https://github.com/github/octodns/issues/63 I think is related to making log messages more clear. 

There could definitely be a bit more descriptive errors, in-depth knowledge of DNS is needed to understand sometimes what went wrong, whether it be on the OctoDNS side or something that's happening because of behavior in a provider.

Thanks for the feedback!

evaryont:
[...]

PR #1411, opened 2026-05-18
Add type hints to core modules, record system, zone system, and provider implementations

ross:
This PR adds comprehensive type hints across the octodns codebase.

### DIFF
--- a/octodns/processor/base.py
+++ b/octodns/processor/base.py
@@ -1,6 +1,10 @@
-#
-#
-#
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from ..provider.base import BaseProvider
+    from ..zone import Zone
 
 
 class ProcessorException(Exception):
@@ -55,7 +59,7 @@ class BaseProcessor(object):
         - :class:`octodns.processor.acme.AcmeMangingProcessor`
     '''
 
-    def __init__(self, name, lenient=False):
+    def __init__(self, name: str, lenient: bool = False) -> None:
         '''
         Initialize the processor.
 
@@ -75,7 +79,9 @@ class BaseProcessor(object):
         self.id = self.name = name
         self.lenient = lenient
 
-    def process_source_zone(self, desired, sources, lenient=False):
+    def process_source_zone(
+        self, desired: Zone, sources: list[BaseProvider], lenient: bool = False
+    ) -> Zone:
         '''
         Process the desired zone after all sources have populated.
 
@@ -113,7 +119,9 @@ class BaseProcessor(object):
         '''
         return desired
 
-    def process_target_zone(self, existing, target, lenient=False):
+    def process_target_zone(
+        self, existing: Zone, target: BaseProvider, lenient: bool = False
+    ) -> Zone:
         '''
         Process the existing zone after the target has populated.
 
@@ -139,7 +147,7 @@ class BaseProcessor(object):
            - Must return ``existing`` which will normally be the ``existing`` param.
            - Must not modify records directly; ``record.copy`` should be called,
              the results of which can be modified, and then ``Zone.add_record``
-             may be used with ``replace=True``.
+             may be used with ``replace=True`` and ``lenient=lenient``.
            - May call ``Zone.remove_record`` to remove records from ``existing``.
            - Implementations should combine ``self.lenient or lenient`` and pass
              the result to any record and zone calls that accept ``lenient`` as
@@ -148,8 +156,12 @@ class BaseProcessor(object):
         return existing
 
     def process_source_and_target_zones(
-        self, desired, existing, target, lenient=False
-    ):
+        self,
+        desired: Zone,
+        existing: Zone,
+        target: BaseProvider,
+        lenient: bool = False,
+    ) -> tuple[Zone, Zone]:
         '''
         Process both desired and existing zones before computing changes.
 
@@ -197,7 +209,13 @@ class BaseProcessor(object):
         '''
         return desired, existing
 
-    def process_plan(self, plan, sources, target, lenient=False):
+    def process_plan(
+        self,
+        plan: Any | None,
+        sources: list[BaseProvider],
+        target: BaseProvider,
+        lenient: bool = False,
+    ) -> Any | None:
         '''
         Process the plan after it has been computed.
 
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -3,6 +3,7 @@
 #
 
 from ..deprecation import deprecated
+from ..processor.base import BaseProcessor
 from ..source.base import BaseSource
 from ..zone import Zone
 from . import SupportsException
@@ -399,7 +400,12 @@ class BaseProvider(BaseSource):
             raise SupportsException(f'{self.id}: {msg}')
         self.log.warning('%s; %s', msg, fallback)
 
-    def plan(self, desired, processors=[], lenient=False):
+    def plan(
+        self,
+        desired: Zone,
+        processors: list[BaseProcessor] | None = None,
+        lenient: bool = False,
+    ) -> Plan | None:
         '''
         Compute a plan of changes needed to sync the desired state to this provider.
 
@@ -458,35 +464,43 @@ class BaseProvider(BaseSource):
             existing, desired, lenient=lenient
         )
 
-        for processor in processors:
-            try:
-                existing = processor.process_target_zone(
-                    existing, target=self, lenient=lenient
-                )
-            except TypeError as e:
-                if "unexpected keyword argument 'lenient'" not in str(e):
-                    raise
-                deprecated(
-                    f'`process_target_zone` method does not support the `lenient` param, fallback is DEPRECATED. Will be removed in 2.0. Class {processor.__class__.__name__}',
-                    stacklevel=99,
-                )
-                existing = processor.process_target_zone(existing, target=self)
+        if processors:
+            for processor in processors:
+                try:
+                    existing = processor.process_target_zone(
+                        existing, target=self, lenient=lenient
+                    )
+                except TypeError as e:
+                    if "unexpected keyword argument 'lenient'" not in str(e):
+                        raise
+                    deprecated(
+                        f'`process_target_zone` method does not support the `lenient` param, fallback is DEPRECATED. Will be removed in 2.0. Class {processor.__class__.__name__}',
+                        stacklevel=99,
+                    )
+                    existing = processor.process_target_zone(
+                        existing, target=self
+                    )
 
-        for processor in processors:
-            try:
-                desired, existing = processor.process_source_and_target_zones(
-                    desired, existing, self, lenient=lenient
-                )
-            except TypeError as e:
-                if "unexpected keyword argument 'lenient'" not in str(e):
-                    raise
-                deprecated(
-                    f'`process_source_and_target_zones` method does not support the `lenient` param, fallback is DEPRECATED. Will be removed in 2.0. Class {processor.__class__.__name__}',
-                    stacklevel=99,
-                )
-                desired, existing = processor.process_source_and_target_zones(
-                    desired, existing, self
-                )
+        if processors:
+            for processor in processors:
+                try:
+                    desired, existing = (
+                        processor.process_source_and_target_zones(
+                            desired, existing, self, lenient=lenient
+                        )
+                    )
+                except TypeError as e:
+                    if "unexpected keyword argument 'lenient'" not in str(e):
+                        raise
+                    deprecated(
+                        f'`process_source_and_target_zones` method does not support the `lenient` param, fallback is DEPRECATED. Will be removed in 2.0. Class {processor.__class__.__name__}',
+                        stacklevel=99,
+                    )
+                    desired, existing = (
+                        processor.process_source_and_target_zones(
+                            desired, existing, self
+                        )
+                    )
 
         # compute the changes at the zone/record level
         changes = existing.changes(desired, self)
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -481,7 +481,6 @@ class BaseProvider(BaseSource):
                         existing, target=self
                     )
 
-        if processors:
             for processor in processors:
                 try:
                     desired, existing = (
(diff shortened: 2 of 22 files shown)


octodns/provider/base.py
```diff
@@ -458,35 +464,43 @@ def plan(self, desired, processors=[], lenient=False):
...
+                        raise
+                    deprecated(
+                        f'`process_target_zone` method does not support the `lenient` param, fallback is DEPRECATED. Will be removed in 2.0. Class {processor.__class__.__name__}',
+                        stacklevel=99,
+                    )
+                    existing = processor.process_target_zone(
+                        existing, target=self
+                    )
+
+        if processors:
```
ross: No need to repeat the `if` here, we can just have both for loops in a single block

```suggestion
```

octodns/processor/base.py
```diff
@@ -139,7 +147,7 @@ def process_target_zone(self, existing, target, lenient=False):
            - Must return ``existing`` which will normally be the ``existing`` param.
            - Must not modify records directly; ``record.copy`` should be called,
              the results of which can be modified, and then ``Zone.add_record``
-             may be used with ``replace=True``.
+             may be used with ``replace=True`` and ``lenient=lenient``.
```
ross, line 150: Interesting catch given that it wasn't even related to the work/plan.

octodns/provider/plan.py
```diff
@@ -335,10 +344,9 @@ def run(self, plans, fh=stdout, *args, **kwargs):
                     fh.write('<h2>')
                     fh.write(current_zone)
                     fh.write('</h2>\n')
-
-                fh.write('<h3>')
-                fh.write(target.id)
-                fh.write('''</h3>
+                    fh.write('<h3>')
+                    fh.write(target.id)
+                    fh.write('''</h3>
```
ross, line 349: Not clear why this was indented into the if block

ross:
claude opus feedback, some of which is inline as well:

# PR #1412 Review — Add type hints across the octodns codebase

## Context

PR #1411 (branch `typing`, 6 commits) introduces type hints across 22 files in the
core octoDNS library, adds a `.mypy.ini` config, adds `mypy` to test/typing
extras, and ships a `minor` changelog entry. Most of the diff is annotation
work, but it also includes a handful of behavior, API, and formatting changes
that warrant attention before merging.

Overall the typing direction is good and the structural choices
(`from __future__ import annotations`, `TYPE_CHECKING`-gated imports, `.mypy.ini`
in strict-ish mode with relaxations for `cmds`, `processor`, and `tests`) are
reasonable. But several non-typing changes look like accidental refactors that
slipped in alongside the annotations, and a few typing decisions are
incorrect or overly narrow.

---

## Critical issues (likely bugs)

### 1. `PlanLogger.run` — trailing horizontal rule moved into `else` branch
`octodns/provider/plan.py` lines 227-231

Before, the closing `buf.write(hr)` and `buf.write('\n')` ran for both the
"have plans" and "no plans" paths. After this PR they're indented inside the
`else`, so when there ARE plans the output no longer ends with the closing
horizontal rule + newline. This is a user-visible regression in the logger
output. There is nothing typing-related driving this change.

```python
        else:
            buf.write(hr)
            buf.write('No changes were planned\n')
            buf.write(hr)        # <-- previously at the else's outdent
            buf.write('\n')      # <-- previously at the else's outdent
```

The existing `TestPlanLogger.test_create` only asserts a substring is present,
so it does not catch this.

### 2. `PlanHtml.run` — `<h3>` + `<table>` now only emitted on zone boundary
`octodns/provider/plan.py` lines 347-359

Original:
```python
if plan.desired.decoded_name != current_zone:
    current_zone = plan.desired.decoded_name
    fh.write('<h2>')
    fh.write(current_zone)
    fh.write('</h2>\n')

fh.write('<h3>')            # per-target heading
fh.write(target.id)
fh.write('''</h3>
<table>
  ...
```

New (indentation pulled inside the `if` block):
```python
if plan.desired.decoded_name != current_zone:
    current_zone = plan.desired.decoded_name
    fh.write('<h2>')
    fh.write(current_zone)
    fh.write('</h2>\n')
    fh.write('<h3>')        # now nested
    fh.write(target.id)
    fh.write('''</h3>
<table>
  ...
```

Two consequences: (a) when multiple targets share a zone, only the FIRST target
gets `<h3>` + a `<table>` header; subsequent targets dump rows into nowhere /
prior table; (b) the existing `<table>` is never closed because the closing
`</table>` writes only happen at the end of the function. `PlanMarkdown.run`
correctly emits `### target.id` per iteration (lines 281-283), confirming this
is an accidental change in `PlanHtml`, not an intended uniformity fix.

`TestPlanHtml.test_simple` uses a single zone+target and asserts a substring,
so it does not catch this either.

### 3. `Change` lost most comparison operators
`octodns/record/change.py`

`Change` no longer extends `EqualityTupleMixin`. It defines `__eq__`, `__lt__`,
and `__hash__` but drops `__ne__`, `__le__`, `__gt__`, `__ge__`. With `__eq__`
defined, `__ne__` is auto-derived correctly, but `__le__`, `__gt__`, `__ge__`
are NOT auto-derived from `__lt__` unless you use `functools.total_ordering`.
Callers that did `change_a >= change_b` (or used comparisons beyond `__lt__`)
now hit the default `object` comparison and either TypeError or compare-by-id.

Additionally, the new `Change.__eq__` returns plain `False` for non-`Change`
others while the rest of the codebase was just updated (in `equality.py`) to
return `NotImplemented` in the same situation. This is inconsistent and breaks
symmetric equality (`other == change`).

Recommendation: keep `Change(EqualityTupleMixin)`, drop the custom dunders, and
let the mixin do its job. The new `_HasRecordAttributes` Protocol can stay if
needed.

### 4. `Record.__repr__` typed as `-> None`
`octodns/record/base.py` line ~497

```python
def __repr__(self) -> None:
    raise NotImplementedError('Abstract base class, __repr__ required')
```

The correct annotation here is `-> str` (the abstract contract) or
`typing.NoReturn` (since the body unconditionally raises). `-> None` is wrong:
subclasses overriding this return `str`, and a Liskov-aware type checker will
complain.

### 5. `Record.disable_validator` annotated `-> str` but returns `int`
`octodns/record/base.py` (Phase 3 commit)

```python
def disable_validator(cls, validator_id, types=None) -> str:
    return cls.validators.disable(validator_id, types=types)
```

`ValidatorRegistry.disable` returns `removed` (an `int` count of removals) —
see `octodns/record/validator.py:73-92`. Should be `-> int`. mypy didn't catch
this because `validators.disable` itself is still untyped (excluded from this
PR's scope).

### 6. `IdnaDict.__init__` accepts the wrong type
`octodns/idna.py`

```python
def __init__(self, data: MutableMapping[str, Any] | None = None) -> None:
```

`self.update(data)` (inherited from `MutableMapping`) accepts any
`Mapping`/iterable of pairs, and existing callers may well pass plain dicts or
arbitrary mappings. `MutableMapping` is unnecessarily restrictive — should be
`Mapping[str, Any] | None`.

Similarly, `ContextDict.__init__` types `*args` as `Mapping[_K, _V] | None`,
which means `*args` is treated as a tuple of those — but `dict.__init__` also
accepts `Iterable[tuple[K, V]]`. The annotation is overly narrow.

### 7. `ZoneValidatorRegistry.enable_sets` annotated `set[str]` but accepts iterable
`octodns/zone/validator.py`

```python
def enable_sets(self, sets: set[str]) -> None:
    ...
    sets_set = set(sets)    # <-- the conversion proves sets isn't required to be a set
```

The original code accepted any iterable (it called `set(sets)`). The new
annotation `set[str]` falsely narrows the contract. Should be `Iterable[str]`.
The variable rename `sets -> sets_set` is also unnecessary; the local
rebinding is allowed.

---

## Non-typing changes piggy-backing on this PR

These are not strictly bugs, but they are out-of-scope for a typing PR and
either belong in their own commits or should be reverted:

- **`equality.py`** — Added `_HasEqualityTuple` Protocol and `isinstance`
  checks returning `NotImplemented`. This is a real behavior change:
  previously comparing to a non-mixin raised `AttributeError`; now returns
  `NotImplemented` (the idiomatic Python). It's an improvement, but it's not
  typing, it's a semantic change, and it warrants its own changelog entry /
  commit. The associated `test_mismatched` test is good. Note inconsistency
  with `Change.__eq__` returning `False` (see issue #3 above).

- **`provider/base.py`** — `plan(processors=[], ...)` (mutable default!)
  changed to `processors: list[BaseProcessor] | None = None` and the body
  wrapped in `if processors:`. Removing the mutable-default footgun is a good
  cleanup but it's a public-API contract change and unrelated to typing. The
  re-formatted nested try/except blocks add noise to the diff. A simpler
  pattern: `processors = processors or []` and leave the loops untouched.

- **`record/change.py`** — Full refactor (drop `EqualityTupleMixin`, add
  Protocol, new ValueError in `__init__`, asserts everywhere). Issues #3 and
  #6 above. The `__init__` `ValueError` for `(None, None)` is a new invariant
  and warrants discussion: is it actually impossible today, or do callers
  somewhere construct empty `Change` objects? At minimum it deserves its own
  commit and changelog note.

- **`provider/plan.py`** — Issues #1 and #2 above. The indentation changes
  were definitely not typing.

- **`idna.py`** — A spelling fix (`deocde` → `decode`) is bundled in. The
  comment in `idna_decode` was moved out of the `try` block (harmless). The
  `str(_encode(s).decode('utf-8'))` and `str(_decode(s))` wrappers are
  presumably to satisfy mypy under `ignore_missing_imports = true` since the
  `idna` library has no stubs, but `bytes.decode('utf-8')` already returns
  `str`, so the outer `str(...)` is dead code. A `cast(str, ...)` would
  document intent better than `str(...)`.

- **`source/base.py`** — Stripped a blank line after `__init__`'s docstring.
  Cosmetic, but pure formatting noise in a typing PR.

- **`processor/base.py`** — Tweaked a docstring (`replace=True` → `replace=True
  and lenient=lenient`). Probably correct but again, not typing.

- **`schema/config.py`** — Removed the top `#\n#\n#` banner. Other files in
  this PR added `from __future__ import annotations` above the banner; this
  one dropped the banner. Pick one convention.

---

## Smaller observations

- `Record._CLASSES: dict[str, type[Record]] = {}` — good. The
  `from __future__ import annotations` at the top makes the self-reference
  fine. Worth a smoke-test against an actual mypy run.

- `Plan` change-list typed as `list[Change]` — `Change` here comes from
  `..record.change`. `Plan` stores `changes` after sorting, so it relies on
  the `__lt__` discussed in issue #3. If you fix issue #3 by going back to
  the mixin you keep the sort behavior.

- The new `Change._HasRecordAttributes` Protocol declares `name`, `_type`,
  `source`, `data` — but the actual `Record.source` is set by callers (and
  may be `None`); `Record.data` is a `@property`. Protocols match on
  attribute name only, so this works, but `Any` for `source` weakens type
  safety where you could use `BaseSource | None`. If you keep the Protocol,
  tighten it.

- `processor/base.py` `process_plan(plan: Any | None, ...) -> Any | None`.
  The `plan` argument should be `Plan | None` (importable under
  `TYPE_CHECKING` from `..provider.plan`).

- `.mypy.ini` excludes `tests/` AND has a `[mypy-tests.*]` section relaxing
  rules — the section is dead since `tests/` is excluded entirely. Pick one
  approach. If you intend to type-check tests eventually, drop the
  `exclude = ^tests/` line.

- `.mypy.ini` sets `python_version = 3.10`. octoDNS generally supports a
  wider range; recommend setting mypy to the minimum supported version so
  the annotations actually hold for end users.

- `setup.py` adds `mypy>=1.0.0` to `tests_require` AND to
  `extras_require['typing']`. Duplicate; if `typing` is opt-in, drop it from
  `tests_require`. (Note: the existing `script/test` should not need mypy
  unless you want CI type-checking, which is fine — but then add a
  `script/type-check` or similar.)

- No `script/type-check` or CI hook was added. Without one, the annotations
  will rot. Consider adding a script and wiring it into CI.

- Changelog entry says "Implement typing thoughout the codebase" — typo
  ("thoughout" → "throughout"), and "throughout the codebase" oversells
  scope (large modules like `record/<types>.py`, `processor/*.py`,
  `cmds/*.py`, `manager.py` are not touched; `provider/yaml.py` only got a
  `from __future__` import).

---

## Suggested path forward

Priority order if narrowing scope:

1. **Fix the regressions in `provider/plan.py`** (issues #1, #2) — these
   produce broken output.
2. **Fix `Change` comparison/equality semantics** (issue #3) — likely breaks
   downstream consumers that use the comparison operators or compare changes
   against non-Change objects.
3. **Correct the wrong annotations** (issues #4, #5, #6, #7) — these defeat
   the purpose of adding types in the first place.
4. **Split out the non-typing changes** into their own commits/PR
   (equality protocol/NotImplemented behavior, mutable default fix in
   `BaseProvider.plan`, `Change.__init__` validation, comment/docstring
   tweaks).
5. **Tighten `.mypy.ini`** (drop dead `[mypy-tests.*]` section or stop
   excluding tests; align `python_version` with supported range).
6. **Add `script/type-check`** and a CI invocation so the types are enforced
   going forward.

## Verification

When the user implements fixes, validate end-to-end with:

- `./script/test tests/test_octodns_plan.py` — run plan-output tests
  (consider strengthening assertions to catch the indentation regressions;
  e.g., count occurrences of `<h3>` and ensure `<table>...</table>` is
  balanced).
- `./script/test tests/test_octodns_record_change.py` — exercise `Change`
  comparisons including `>=`, `<=`, `>`, and equality with non-Change
  values.
- `./script/test tests/test_octodns_equality.py` — confirm
  `NotImplemented` returns hold.
- `./script/coverage` — ensure 100% branch coverage still holds (the new
  `if processors:` branches need both arms covered).
- `python -m mypy octodns` — run mypy from the project root using the new
  `.mypy.ini`; should report zero errors on the typed surface.

## Critical files to revisit

- `octodns/provider/plan.py` (issues #1, #2)
- `octodns/record/change.py` (issues #3, #6)
- `octodns/record/base.py` (issues #4, #5)
- `octodns/idna.py` (issue #6)
- `octodns/zone/validator.py` (issue #7)
- `octodns/provider/base.py` (non-typing API change)
- `octodns/equality.py` (non-typing semantic change)
- `.mypy.ini` (config cleanup)
- `setup.py` (duplicate mypy declaration)
- `.changelog/e01f583298224173950e306ca521bb19.md` (typo)

PR #1413, opened 2026-05-23
Implement additional DNS validators for octoDNS

ross:
This PR introduces several new validators to enforce DNS best practices across octoDNS:

- **SpfRecordTypeValidator**: Raises validation error when SPF record type is used to enforce transition to TXT.
- **RootCnameZoneValidator**: Rejects CNAME records at the zone apex.
- **CnameTargetNotCnameZoneValidator**: Discourages internal CNAME chaining.
- **ALIAS Coexistence**: Ensures ALIAS records do not coexist with A or AAAA records.
- **TargetNotIpValidator**: Enforces that target fields for CNAME, MX, NS, SRV, ALIAS, PTR, and DNAME records are hostnames rather than direct IP addresses.
- **Tests**: Comprehensive tests added for all new validator logic.

/cc #1396

### DIFF
--- a/.changelog/24133387c13041949fa4cf79276736f3.md
+++ b/.changelog/24133387c13041949fa4cf79276736f3.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Implement additional DNS validators (SPF, CNAME, ALIAS, IPs)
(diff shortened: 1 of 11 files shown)


ross (commented):
/ci modules build

PR #1416, opened 2026-05-24
Add script/coverage-report helper to parse coverage.json

ross:
This PR adds a helper script to list files with less than 100% coverage, including missing line numbers and branch coverage information.


ross:
This should be a huge help for AI agents that often seem to have issues with digesting and taking action on the LARGE code coverage reports, regardless of format.

Issue #1218, opened 2024-10-09
V2: explore options for removing the requirement to escape ;


ross:
This is my biggest regret w/octoDNS. Everything else just works as people expect, but for some reason I didn't internally handle the escaping and expected it to be pre-escaped 😦 

I'm not sure what the options for revising this are and whatever route is taken it needs to be reasonable pain-free to configure things to keep the old behavior.

ross:
Addressed in #1419

PR #1419, opened 2026-05-25
Deprecate YamlProvider escaped_semicolons=True default

ross:
Deprecates the default value of `escaped_semicolons` in `YamlProvider` to prepare for changing the default to `False` in 2.0. 

/cc #1218 
/cc #1253


ross (commented):
/cc modules bump

Issue #1382, opened 2026-04-26
!include should do tilde expansion


rikfaith:
It is often convenient to keep, for example, tokens, outside the octodns config directory so they are not committed to git.  To make the configuration user-agnostic, it would be helpful if **!include ~/foo** expanded to include /home/username/foo.

Even better, would be to allow **!include $(bash-commands)** which would run the bash-commands and use the output as the parameter for !include.  This way, tokens could be easily accessed from encrypted key stores.

Both of the above solutions avoid storing keys in environment variables.

rikfaith:
octoDNS 1.16.0

ross:
> It is often convenient to keep, for example, tokens, outside the octodns config directory so they are not committed to git. To make the configuration user-agnostic, it would be helpful if **!include ~/foo** expanded to include /home/username/foo.

I've always done this with .gitignore, usually in a .env file.

> it would be helpful if **!include ~/foo** expanded to include /home/username/foo.

I'd be open to that. We probably need to use a built-in python library for resolving file paths to avoid reinventing the wheel and potential security issues.

> Even better, would be to allow **!include $(bash-commands)** which would run the bash-commands and use the output as the parameter for !include. This way, tokens could be easily accessed from encrypted key stores.

No real interest in this. It seems like a 🦶 🔫.

rikfaith:
os.path.expanduser and Path.expanduser are in the standard Python library.

ross:
Implemented in #1420

PR #1420, opened 2026-05-25
Support tilde (~) expansion in !include directives

ross:
Supports expanding the tilde (`~`) character to the user's home directory in `!include` directives within YAML configurations.

/cc #1382 Support tilde (~) expansion in !include directives


ross (commented):
/ci modules bump

ross:
Fixes #1382

PR #1430, opened 2026-06-08
MailZoneValidator: min_mx + single_mx_regexes exemptions for single-MX providers

### LINKED ISSUE #1423: Requesting a non-lenient way to allow single MX
Related to #1422, some MTA providers only give a single MX record, specifically SendGrid. Can an easy(tm) way to created an exemption on a single MX record entry (via patterns would be cool) to not throw this exception error? Sprinkling `octodns: lenient: true` into those entries adds operational overhead.

ross:
## Summary

Closes #1423.

- Adds `min_mx=2` parameter to `MailZoneValidator` — the minimum number of MX values required for the redundancy check. Setting `min_mx=1` effectively disables it.
- Adds `DEFAULT_SINGLE_MX_REGEXES` class constant: a tuple of regexes (matched via `re.search` against the exchange, including its trailing dot) for providers known to legitimately operate with a single MX record. Seeded with **SendGrid** (`mx.sendgrid.net.`), **Amazon SES** (`inbound-smtp.<region>.amazonaws.com.`), and **Postmark** (`inbound.postmarkapp.com.`). Carries a PR-invite comment for community additions.
- Adds `single_mx_regexes=None` parameter to append user-supplied regexes to the built-in list without replacing it.
- The exemption check runs only when `len(record.values) == 1` to avoid regex overhead for records that already have enough values.
- Invalid regex patterns in `single_mx_regexes` raise `ValueError` at construction time with the offending pattern in the message.

## Configuration example

```yaml
validators:
  mail:
    class: octodns.zone.mail.MailZoneValidator
    # lower the threshold (or set to 1 to disable the redundancy check)
    min_mx: 1
    # add your own single-MX provider exemptions
    single_mx_regexes:
      - '^mx\.myprovider\.com\.$'
```

## Test plan

- [x] `test_mail_validator_init` — verifies defaults and new param storage
- [x] `test_single_mx_provider_sendgrid_exempt` — SendGrid is exempt
- [x] `test_single_mx_provider_aws_ses_exempt` — SES regional hosts exempt; non-SES amazonaws.com still flagged
- [x] `test_single_mx_provider_postmark_exempt` — Postmark is exempt
- [x] `test_min_mx_one_disables_redundancy_check` — `min_mx=1` passes a single non-exempt MX
- [x] `test_min_mx_three_requires_three_values` — `min_mx=3` flags a 2-value MX with correct message
- [x] `test_single_mx_regexes_custom_extends_defaults` — custom regex exempts new provider; built-in defaults still apply
- [x] `test_single_mx_regexes_invalid_pattern_raises` — bad regex raises `ValueError` with pattern in message
- [x] 778/778 tests pass, 100% branch coverage

/cc Fixes #1423

### DIFF
--- a/octodns/zone/mail.py
+++ b/octodns/zone/mail.py
@@ -3,6 +3,8 @@
 #
 
 from logging import getLogger
+from re import compile as re_compile
+from re import error as re_error
 
 from .base import Zone
 from .validator import ValidationReason, ZoneValidator
@@ -30,7 +32,11 @@ class MailZoneValidator(ZoneValidator):
 
     'mail' mode enforces:
 
-    - Multiple MX records for redundancy (at apex and throughout the zone).
+    - At least ``min_mx`` MX records for redundancy (default 2) at the apex
+      and throughout the zone. MX records with a single value whose exchange
+      matches a known single-MX provider (see ``DEFAULT_SINGLE_MX_REGEXES``)
+      or a user-supplied ``single_mx_regexes`` pattern are exempt from this
+      check.
     - Presence of an SPF record at the apex.
     - SPF record terminates with ~all or -all.
     - Presence of a DMARC record at _dmarc.
@@ -48,12 +54,41 @@ class MailZoneValidator(ZoneValidator):
     parent zone per RFC 7489 §6.6.3.
     '''
 
-    def __init__(self, id, mode='auto', sets=None):
+    # MX exchanges of providers known to (correctly) operate with a single MX
+    # record. A record with exactly one value whose exchange matches one of
+    # these patterns is exempt from the ``min_mx`` redundancy check. Patterns
+    # are matched via re.search against the exchange including its trailing dot.
+    #
+    # PRs welcome: if you run across another reputable provider that only hands
+    # out a single MX, please add it here.
+    DEFAULT_SINGLE_MX_REGEXES = (
+        r'^mx\.sendgrid\.net\.$',  # SendGrid
+        r'^inbound-smtp\..+\.amazonaws\.com\.$',  # Amazon SES (region-specific)
+        r'^inbound\.postmarkapp\.com\.$',  # Postmark
+    )
+
+    def __init__(
+        self, id, mode='auto', min_mx=2, single_mx_regexes=None, sets=None
+    ):
         super().__init__(id, sets=sets)
         self.log = getLogger(f'MailZoneValidator[{id}]')
         if mode not in ('auto', 'mail', 'no-mail'):
             raise ValueError(f'Unknown mode "{mode}"')
         self.mode = mode
+        self.min_mx = min_mx
+        self._single_mx_res = []
+        for r in (*self.DEFAULT_SINGLE_MX_REGEXES, *(single_mx_regexes or [])):
+            try:
+                self._single_mx_res.append(re_compile(r))
+            except re_error as e:
+                raise ValueError(
+                    f'Invalid single_mx_regexes pattern "{r}": {e}'
+                )
+
+    def _is_single_mx_provider(self, mx_record):
+        # called only when mx_record has exactly one value
+        exchange = str(mx_record.values[0].exchange)
+        return any(r.search(exchange) for r in self._single_mx_res)
 
     def _is_null_mx(self, mx_record):
         return (
@@ -308,10 +343,16 @@ class MailZoneValidator(ZoneValidator):
             if self._is_null_mx(record):
                 continue
 
-            if len(record.values) < 2:
+            if len(record.values) < self.min_mx:
+                # only a lone MX can belong to an exempt single-MX provider;
+                # skip the regex matching when it couldn't help
+                if len(record.values) == 1 and self._is_single_mx_provider(
+                    record
+                ):
+                    continue
                 reasons.append(
                     ValidationReason(
-                        f'MX record "{record.fqdn}" should have at least 2 values for redundancy, found {len(record.values)}',
+                        f'MX record "{record.fqdn}" should have at least {self.min_mx} values for redundancy, found {len(record.values)}',
                         [record],
                     )
                 )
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci modules bump

brianeclow:
I am having difficulty adding user-supplied regexes. Gleaning from the primary validator docs and the config snippet, I added:

```
manager:
  validators:
    enabled:
    - strict
    - best-practice
    mail:
      class: octodns.zone.mail.MailZoneValidator
      single_mx_regexes:
      - ^smtp\.google\.com\.$
      - ^feedback-smtp\.us-west-2\.amazonses\.com\.$
      - ^mx\.hubapi\.com\.$
```

Which does not appear to affect any change.
- Added `min_mx`, no change
- Removed `single_mx_regexes` key, no change.
- Only `min_mx`, no change.
- Moved `validators: mail` to a sibling key of `manager`, octoDNS zoneValidator duplicate name exception.
    I expected this one based on the custom validators.

Did I miss something when adding it to the configuration?

**EDIT:**

I read the manager code, and it looks like the configuration should be:

```
manager:
  validators:
    enabled:
    - strict
    - best-practice
    zone:
      validators:
        mail:
          class: octodns.zone.mail.MailZoneValidator
          single_mx_regexes:
          - ^smtp\.google\.com\.$
          - ^feedback-smtp\.us-west-2\.amazonses\.com\.$
          - ^mx\.hubapi\.com\.$
```

However, I am unable to get this to work either.

ross:
I believe the issue is that custom `validators` are defined in a top level key like `processors`. 

Given the name `mail` it should hopefully overwrite the default one and be enabled, least that's the intention.

ross:
Looks like things landed on not allowing built-ins to be replaced. The following config should work:

```yaml
# yaml-language-server: $schema=https://octodns.readthedocs.io/en/stable/_static/octodns-config.schema.json
---
manager:
  validators:
    enabled:
      - strict
      - best-practice
    zone:
      disable_validators:
        - mail
      validators:
        - mail-extra

providers:
  config:
    directory: ./config/1430/
  out:
    directory: /tmp/octodns

validators:
  mail-extra:
    class: octodns.zone.mail.MailZoneValidator
    single_mx_regexes:
      - ^smtp\.google\.com\.$
      - ^feedback-smtp\.us-west-2\.amazonses\.com\.$
      - ^mx\.hubapi\.com\.$

zones:
  '*':
    sources:
      - config
    targets:
      - out
```

```yaml
# yaml-language-server: $schema=https://octodns.readthedocs.io/en/stable/_static/octodns-zone.schema.json
---
'':
  - type: MX
    value:
      exchange: smtp.google.com.
      preference: 0
  - type: TXT
    value: v=spf1 include:_spf.google.com -all
_dmarc:
  type: TXT
  value: v=DMARC1\; p=reject\; pct=100\; fo=1\; rua=mailto:user@example.com\;
    ruf=mailto:user@example.com\; ri=3600
```

Note that also includes the new JSON schema which is a nice to have if your editor utilizes it https://octodns.readthedocs.io/en/latest/configuration.html#json-schema-for-zone-yaml-files.

I'll go ahead and add those servers to the default list. Intention is for anything common to be in there.

brianeclow:
> I'll go ahead and add those servers to the default list. Intention is for anything common to be in there.

The `feedback-smtp` AWS one is rather universal, but there are a few more I have which are from bespoke vendors. Which is why I was trying to use the user definted.

brianeclow:
After updating to the new version of this branch, octoDNS exempts all but the AWS matches and the bespoke vendors.

When adding the `validators: mail-extras` block with a non-default, user-defined regex, it appears to append all non-listed exemptions from its checks. If I include all default and user-defined to the `single_mx_regexes` array, then I am left with the original set of AWS and bespoke vendors, as found by the first run of the validator.

octodns/zone/mail.py
```diff
@@ -48,12 +54,43 @@ class MailZoneValidator(ZoneValidator):
...
 
-    def __init__(self, id, mode='auto', sets=None):
+    # MX exchanges of providers known to (correctly) operate with a single MX
+    # record. A record with exactly one value whose exchange matches one of
+    # these patterns is exempt from the ``min_mx`` redundancy check. Patterns
+    # are matched via re.search against the exchange including its trailing dot.
+    #
+    # PRs welcome: if you run across another reputable provider that only hands
+    # out a single MX, please add it here.
+    DEFAULT_SINGLE_MX_REGEXES = (
```
brianeclow, line 64: Can we add to the defaults `smtp.siftrock.com.`, which is Marketo.

octodns/zone/mail.py
```diff
@@ -48,12 +54,43 @@ class MailZoneValidator(ZoneValidator):
...
+        r'^inbound\.postmarkapp\.com\.$',  # Postmark
+        r'^(feedback|inbound)-smtp\..+\.amazonaws\.com\.$',  # Amazon AWS (region-specific)
```
brianeclow: My `feedback-smtp` is on a different SLD `amazonses.com.` I am going to guess there might be `inbound-smpt` on that SLD as well for people because naming is hard.

ross: Added both to the regex.

brianeclow: Thank you!

brianeclow:
Thank you for the additional "known" remailers.

Also... Eureka! Iterating through variations to see if I can insert into the user-defined regexes, I found this the magic incantation:

```
---
manager:
  validators:
    enabled:
    - strict
    - best-practice
    zone:
      disable_validators:
      - mail
validators:
  mail-extra:
    class: octodns.zone.mail.MailZoneValidator
    single_mx_regexes:
    - ^smtp\.example\.com\.$
```

To access the var, one needs to disable the mail validator, then add it as a custom validator under a unique name, eg`validators: mail-extra`. This provide the surface to insert user-defined regexes into the exemption var. The mail validator runs as expected.

Based on the above conversation, I don't believe this was the intended method to access the user-defined regexes, however it does work. The duplicated exceptions when adding user-defined regexes was the clue. The manager would run the default mail validator, then run the `mail-extra` validator. Disabling the default mail validation run in favor of the mail-extra run, resolves the duplication.

ross:
> Based on the above conversation, I don't believe this was the intended method to access the user-defined regexes, however it does work

If I understand correctly that is the intended method. 

I'm thinking about allowing overwriting the built-in ids to simplify things though.

brianeclow:
Works for me.

brianeclow (approved):

PR #1432, opened 2026-06-20
OwnershipProcessor: refuse to take over foreign-owned records by default

### LINKED ISSUE #1368: Ownership processor takes over foreign ownership records
Currently when an ownership record already exists for a record octoDNS manages, the ownership record will be overwritten regardless of the previous value.

I find the current behavior to be at least unintuitive. I understood the txt_value of an ownership record as the "owner" of the corresponding record. If octoDNS is not the owner of a record it wants to modify it should throw. If the current behavior is prefered, I think an "allow_takeover" parameter could make sense, which toggles the described behavior.

I'm aware that my use case of using multiple octoDNS instances is not supported but I can imagine this to be a problem in other scenarios as well (e.g. an external process setting ownership records).

If you consider this a bug, I'd be happy to contribute a fix. Just let me know.

**Expected behavior**

octoDNS will throw an error if it encounters an ownership records that does not belong to it.

**Actual behavior**

octoDNS will override the tag with its own ownership txt_value.

ross:
## Summary
* `OwnershipProcessor` previously silently overwrote an existing `_owner` TXT marker even when its value belonged to someone/something else, effectively taking over a record octoDNS didn't actually own.
* Adds `allow_takeover` (default `false`). When disallowed, `process_plan` now raises `OwnershipException` if it finds a foreign-valued ownership marker on a record octoDNS is about to claim, naming the record and the foreign value. Setting `allow_takeover: true` restores the old silent-takeover behavior.
* Unmanaged records with foreign ownership markers are left untouched either way, matching current behavior.

Closes #1368

## Test plan
- [x] `./script/test tests/test_octodns_processor_ownership.py` — new `test_allow_takeover` covers default-raises, `allow_takeover=True` opt-out, and unmanaged-record sanity case
- [x] `./script/test` — full suite (779 tests) passes
- [x] `./script/lint` — clean
- [x] `./script/coverage` — 100% maintained

### DIFF
--- a/.changelog/1b6ad68b3a95427fa75cea2c124e62fb.md
+++ b/.changelog/1b6ad68b3a95427fa75cea2c124e62fb.md
@@ -1,4 +1,4 @@
 ---
 type: minor
 ---
-OwnershipProcessor: refuse to take over records whose existing ownership marker has a foreign value; add allow_takeover (default false) to restore prior behavior. Closes #1368
+OwnershipProcessor: refuse to take over records whose existing ownership marker has a foreign value; add allow_takeover (default false) to restore prior behavior.
--- a/.changelog/1b6ad68b3a95427fa75cea2c124e62fb.md
+++ b/.changelog/1b6ad68b3a95427fa75cea2c124e62fb.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci modules bump

PR #1434, opened 2026-06-21
Allow config validators to override built-in validators by id

ross:
## Summary
- A config-declared validator (in the top-level `validators:` section) whose key matches a built-in's id now *replaces* that built-in instead of raising an "already registered" error.
- Makes it easy to tweak a built-in's parameters without disabling it and inventing a new id — e.g. forcing the `mail` zone validator into `no-mail` mode:
  ```yaml
  validators:
    mail:
      class: octodns.zone.mail.MailZoneValidator
      mode: no-mail
  ```
- Genuine built-in-vs-built-in id collisions (an import-time programming bug) still raise, since built-ins register without `replace=True`.

## Test plan
- [x] `./script/test` — 781 passed
- [x] `./script/coverage` — 100%
- [x] `./script/lint` — clean
- [x] Repurposed `validators-id-collision.yaml` / `zone-validators-id-collision.yaml` fixtures into override-success tests; added registry-level unit tests for `replace=True`/`False` for both record and zone validator registries

/cc https://github.com/octodns/octodns/pull/1430
/cc https://github.com/octodns/octodns/issues/1423
/cc @brianeclow fyi


ross (commented):
/ci modules build

brianeclow:
Updated my configurations, tested, works as expected. Thank you.

PR #1436, opened 2026-06-24
Add --honor-lenient flag to octodns-validate

### LINKED ISSUE #1431: Validate accept `lenient`
I would love to see the `validate` method when called from a custom script (or the CLI) accept the in-line entry declaration of `octodns: lenient: true` or the zone level declaration of `lenient: true` with the same grace as the sync method. Additional flagging/options?

The specific case I am addressing is Route53 is a be loose on requiring trailing dots (https://github.com/octodns/octodns-route53/issues/96). Some inherited zones, which I cannot simply delete the offending entries and recreate via an octoDNS sync, I have enabled lenient to avoid the noise created by this, um, Route53 feature. However, our tooling leverages the validate method, and it throws warnings. As we are all moving toward more agent based adventures, redirecting the agent every time has become cumbersome.

Route53 is the only provider with this feature, and disabling the trailing dot validation on all our Route53 zones seems like the wrong direction.

**EDIT**: I am attempting to create a solution on my fork.

brianeclow:
Closes #1431

## Summary

- Add `--honor-lenient` flag to `octodns-validate` that suppresses validation warnings for zones and records with `lenient: true` configured, exiting 0 when no non-lenient issues remain
- Make `--all` and `--honor-lenient` composable: `--all` keeps going on errors; `--honor-lenient` suppresses lenient-configured warnings; both together show all non-lenient issues and suppress lenient ones
- Thread `zone_config_lenient` separately from the `--all` lenient flag through `Zone.validate()` so `suppress_lenient_warnings` only silences reasons that are lenient by config, not by `--all`
- Introduce `_suppress_lenient_record_warnings` / `_zone_config_lenient` ContextVars in `record/base.py` so that record-level best-practice validator warnings emitted during `populate()` are also suppressed for lenient-configured records/zones — previously these bypassed `Zone.validate()` entirely and always triggered exit 1
- Fix pre-existing bug in `validate_configs()` where the `lenient` loop variable was overwritten per-zone, causing lenient to bleed across zones

## Behavior

| Flags | Lenient-configured issues | Non-lenient issues | Exit |
|---|---|---|---|
| _(none)_ | warns, exits 1 | raises exception | 1 |
| `--all` | warns, exits 1 | warns, exits 1 | 1 |
| `--honor-lenient` | silent | raises exception | 0 or 1 |
| `--all --honor-lenient` | silent | warns, exits 1 | 0 or 1 |

## Test plan

- [x] `test_validate_suppress_lenient_warnings` — `Zone.validate()` emits no warning when `suppress_lenient_warnings=True`; non-lenient issues still raise
- [x] `test_validate_suppress_emits_debug` — `Zone.validate()` emits DEBUG log when suppressing (audit trail)
- [x] `test_validate_zone_config_lenient_suppresses_empty_record_reasons` — `zone_config_lenient=True` suppresses zone-wide reasons that have no records (which `reason.lenient` can't handle); `lenient=True` (--all) alone without `zone_config_lenient` does NOT suppress
- [x] `test_validate_configs_suppress_lenient_warnings` — `validate_configs()` suppresses zone-level warnings end-to-end; confirms `--all` + `--honor-lenient` on non-lenient zone still warns
- [x] `test_validate_configs_suppress_lenient_record_warnings` — verifies `validate_configs()` activates `suppress_lenient_record_warnings` with the correct `zone_config_lenient` argument for lenient zones, and does not activate it when `suppress_lenient_warnings=False`
- [x] `test_validate_configs_lenient_not_inherited_across_zones` — regression: lenient zone warns, strict zone raises; loop variable bug fixed
- [x] `test_suppress_lenient_record_warnings_context_manager` — unit test for the ContextVar mechanism with real `Record.new()` calls: lenient-by-record-config suppressed, lenient-by-zone-config suppressed, lenient-by-`--all`-only NOT suppressed
- [x] CLI tests: clean exit, `--honor-lenient` exits 0 on lenient-only issues, `--honor-lenient` raises on non-lenient issues, `--all` exits 1 on lenient issues, `--all` + `--honor-lenient` exits 0 when only lenient issues, `--all` + `--honor-lenient` exits 1 when non-lenient issues present
- [x] 795 tests pass, 100% coverage maintained on all measured files

## Code review follow-ups applied

- Renamed `zone_lenient` → `effective_lenient` in `validate_configs()` to distinguish the resolved `--all`-or-config combination from `zone_config_lenient` (config-only) at a glance
- Replaced `test_validate_configs_suppress_lenient_record_warnings` stub implementation: removed private ContextVar imports and `instrumented_new` that re-implemented suppression logic; new test patches `suppress_lenient_record_warnings` in `octodns.manager` and asserts correct call arguments, testing the integration boundary rather than internal mechanics

### DIFF
--- a/.changelog/ea0ae7c844ee4b5eab36caaa26a2ea2a.md
+++ b/.changelog/ea0ae7c844ee4b5eab36caaa26a2ea2a.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add suppress_lenient_warnings param to Zone.validate() and --honor-lenient flag to octodns-validate CLI
--- a/.changelog/ea0ae7c844ee4b5eab36caaa26a2ea2a.md
+++ b/.changelog/ea0ae7c844ee4b5eab36caaa26a2ea2a.md
@@ -1,4 +1,4 @@
 ---
 type: minor
 ---
-Add suppress_lenient_warnings param to Zone.validate() and --honor-lenient flag to octodns-validate CLI
+Add `--honor-lenient` flag to `octodns-validate`; suppress validation warnings for zones and records configured with `lenient: true`, exiting 0 when no non-lenient issues remain. Also fixes a bug where a zone with `lenient: true` in config could cause subsequent zones in the same run to incorrectly inherit lenient mode.
--- a/octodns/cmds/validate.py
+++ b/octodns/cmds/validate.py
@@ -31,13 +31,11 @@ def main():
     parser.add_argument(
         '--all',
         action='store_true',
-        default=False,
         help='Validate records in lenient mode, printing warnings so that all validation issues are shown',
     )
     parser.add_argument(
         '--honor-lenient',
         action='store_true',
-        default=False,
         help='Suppress warnings for zones and records configured with lenient: true, exiting 0 if no non-lenient issues remain',
     )
 
--- a/octodns/cmds/validate.py
+++ b/octodns/cmds/validate.py
@@ -43,6 +43,9 @@ def main():
 
     args = parser.parse_args(WARNING)
 
+    if args.all and args.honor_lenient:
+        parser.error('--all and --honor-lenient are mutually exclusive')
+
     flagging = FlaggingHandler()
     getLogger('Record').addHandler(flagging)
     getLogger('Zone').addHandler(flagging)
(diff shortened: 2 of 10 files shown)


brianeclow:
Ross and I have mostly solved the core item I was looking for without leveraging this PR. I was also able to adjust my validation tooling to address the verbosity and exit codes for "acceptable" validations. Looking at the tool holistically, leaving the validation as it is, makes more sense.
Closing.

PR #1438, opened 2026-06-25
Collapse remaining multi-line string concatenations

ross:
## Summary
- #1437 fixed two real bugs where a multi-line implicit string concatenation lost its `f` prefix on continuation lines, so a `{placeholder}` was emitted literally instead of interpolated.
- Audited the rest of `octodns/` for the same pattern: 57 other multi-line concatenations, 31 of them f-strings. None were actually broken (every piece was correctly prefixed), but collapsing them removes the risk and matches the style of the fix.
- Collapsed 55 of the 57 sites to single lines. Left two multi-line where collapsing would have pushed the line past ~175 chars: the `manager.py` `disable_validators` deprecation message, and `LOC.__repr__`'s 12-field string.

/cc #1436 found the original problem
/cc #1437 fixed the two real bugs; this follows up with the rest of the audit

## Test plan
- [x] `./script/test` passes (781 passed)
- [x] `./script/coverage` at 100%
- [x] `./script/lint` and `./script/format` clean

### DIFF
--- a/.changelog/b01976da4f474745b5c71935640d719a.md
+++ b/.changelog/b01976da4f474745b5c71935640d719a.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Collapse remaining multi-line string concatenations to avoid f-string interpolation bugs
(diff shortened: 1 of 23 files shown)


ross:
There's been several rounds of small bugs;/issues with the old multi-line string concatentations. Now that black doesn't break them up/consider them too long I'd prefer to re-join them.

PR #1439, opened 2026-06-29
Add required validator_id parameter to ValidatorReason

ross:
First step of a few improving Validators. This adds a required `validator_id` param to the `ValidationReason` object. This should allow the warnings/errors to include the id of the validator that originated the message. Which should make it easier to know what to disable if users are so-inclined or otherwise allow them to dig deeper into what's going in if need be.

/cc https://github.com/octodns/octodns/issues/1431#issuecomment-4837536493 which lead to these changes

### DIFF
--- a/.changelog/03bb923602674dd795bc41e348ecc850.md
+++ b/.changelog/03bb923602674dd795bc41e348ecc850.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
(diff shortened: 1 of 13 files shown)


ross (commented):
/ci modules bump

ross (commented):
/ci modules build

PR #1440, opened 2026-07-01
Fix lenient parameter leak across zones in validate_configs

brianeclow:
Reassigning the `lenient` function parameter inside the zone loop caused it to stay True for all subsequent zones once any zone had `lenient: true` in its config. Use a loop-local `zone_lenient` variable instead so each zone is evaluated independently.


ross (approved):
Thanks for extracting this. I remember seeing it in your PR and making a mental note to pull it out, but it didn't stick.

brianeclow:
Glad I could help. 😀

PR #1441, opened 2026-07-01
Add per-zone validator disabling

ross:
## Summary

- Adds a per-zone `validators` config key (mirroring the global `manager.validators` shape) that lets a zone additively disable specific built-in record, value, or zone validators without affecting any other zone.
- `Zone.__init__` now takes a `validators` dict, parses it itself via `_parse_validators_config`, and raises `ZoneException` for invalid entries (e.g. attempting to disable a framework bridge validator) — `Manager.get_zone` just passes the zone's config through.
- The per-zone disable set is threaded explicitly through `Record.new` → `RecordValidator.validate(..., disabled=None)` → the registry's `process_record`/`process_values`, with a TypeError-based deprecation fallback (matching the existing `lenient`-param pattern in `manager.py`/`provider/base.py`) so third-party validators on the old 4-arg signature keep working.
- Fixes a related bug found during review: `BaseProvider.plan()` built the "existing" state zone without forwarding the desired zone's per-zone validators config, so a validator disabled for a zone would still run (and warn) while populating current state.

```yaml
zones:
  legacy.example.com.:
    sources: [config]
    targets: [route53]
    validators:
      record:
        disable_validators:
          '*': [healthcheck]
          MX: [mx-value]
      zone:
        disable_validators:
          - dname-coexistence
```

Builds on the validator infrastructure from #1378, #1380, #1387, #1396, and follows the same disable-by-id shape as #1434.

## Test plan

- [x] `./script/test` — 801 tests passing
- [x] `./script/coverage` — 100% coverage maintained
- [x] `./script/lint` / `./script/format --check` — clean
- [x] New unit tests for skip logic, the TypeError deprecation fallback, and bridge-id rejection in `record/validator.py`, `zone/base.py`, `zone/validator.py`
- [x] New manager-level integration test (`test_zone_disable_validators`) driving a real `sync()` through the thread-pool executor across two zones — one with a per-zone disable, one without — confirming no cross-zone leakage
- [x] New regression test (`test_plan_existing_zone_inherits_validators_config`) covering the `BaseProvider.plan()` fix; confirmed it fails without the fix
- [x] Docs updated in `docs/configuration.rst`

### DIFF
--- a/octodns/provider/base.py
+++ b/octodns/provider/base.py
@@ -429,7 +429,11 @@ class BaseProvider(BaseSource):
         '''
         self.log.info('plan: desired=%s', desired.decoded_name)
 
-        existing = Zone(desired.name, desired.sub_zones)
+        existing = Zone(
+            desired.name,
+            desired.sub_zones,
+            validators=desired.validators_config,
+        )
         exists = self.populate(existing, target=True, lenient=True)
         if exists is None:
             # If your code gets this warning see Source.populate for more
--- a/octodns/record/alias.py
+++ b/octodns/record/alias.py
@@ -17,7 +17,7 @@ class AliasRootValidator(RecordValidator):
     type only has meaning at the apex.
     '''
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         if name != '':
             return ['non-root ALIAS not allowed']
         return []
--- a/octodns/record/base.py
+++ b/octodns/record/base.py
@@ -28,7 +28,7 @@ class NameValidator(RecordValidator):
     limits from RFC 1035, and flags empty/double-dot labels.
     '''
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         reasons = []
         if name == '@':
             reasons.append('invalid name "@", use "" instead')
@@ -56,7 +56,7 @@ class TtlValidator(RecordValidator):
     integer.
     '''
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         reasons = []
         try:
             ttl = int(data['ttl'])
@@ -73,7 +73,7 @@ class HealthcheckValidator(RecordValidator):
     present, is one of the supported protocols.
     '''
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         reasons = []
         try:
             if data['octodns']['healthcheck']['protocol'] not in (
@@ -99,9 +99,12 @@ class ValueTypeValidator(RecordValidator):
     def __init__(self):
         super().__init__(id='_value-type')
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         return _process_value_validators(
-            record_cls._value_type, data.get('value', None), record_cls._type
+            record_cls._value_type,
+            data.get('value', None),
+            record_cls._type,
+            disabled=disabled,
         )
 
 
@@ -117,7 +120,7 @@ class ValuesTypeValidator(RecordValidator):
     def __init__(self):
         super().__init__(id='_values-type')
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
         values = data.get('values', data.get('value', []))
         values = (
             values
@@ -125,7 +128,7 @@ class ValuesTypeValidator(RecordValidator):
             else ([] if values is None else [values])
         )
         return _process_value_validators(
-            record_cls._value_type, values, record_cls._type
+            record_cls._value_type, values, record_cls._type, disabled=disabled
         )
 
 
@@ -230,7 +233,8 @@ class Record(EqualityTupleMixin):
             if context:
                 msg += f', {context}'
             raise Exception(msg)
-        reasons.extend(_class.validate(name, fqdn, data))
+        disabled = zone.disabled_record_validators
+        reasons.extend(_class.validate(name, fqdn, data, disabled=disabled))
         try:
             lenient |= data['octodns']['lenient']
         except KeyError:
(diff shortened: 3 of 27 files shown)


ross (commented):
/cc @brianeclow this is WIP/first pass, but it's roughly what I'd see zone-level disabling of validators looking like. Funny thing is Claude wanted to do a ContextVar var solution rather than just pass a param with a no-op default down through things similar to (i assume its solution) for the suppression stuff.

octodns/provider/base.py
```diff
@@ -429,7 +429,11 @@ def plan(self, desired, processors=[], lenient=False):
         '''
         self.log.info('plan: desired=%s', desired.decoded_name)
 
-        existing = Zone(desired.name, desired.sub_zones)
+        existing = Zone(
+            desired.name,
+            desired.sub_zones,
+            validators=desired.validators_config,
```
ross, line 435: Took me a minute of thinking about this one to decide that it's OK. There'd be no other way to configure them for the existing zone (the one the provider fills out with the current state) since desired is the one created from based on the octoDNS config YAML when things are kicked off.

octodns/record/alias.py
```diff
@@ -17,7 +17,7 @@ class AliasRootValidator(RecordValidator):
     type only has meaning at the apex.
     '''
 
-    def validate(self, record_cls, name, fqdn, data):
+    def validate(self, record_cls, name, fqdn, data, disabled=None):
```
ross, line 20: Majority of the changes are just piping this value through.

octodns/record/base.py
```diff
@@ -435,8 +441,10 @@ def __repr__(self):
         raise NotImplementedError('Abstract base class, __repr__ required')
 
 
-def _process_value_validators(value_type, values, _type):
-    return Record.validators.process_values(value_type, values, _type)
+def _process_value_validators(value_type, values, _type, disabled=None):
+    return Record.validators.process_values(
+        value_type, values, _type, disabled=disabled
+    )
```
ross, line 479: This is where the disabled config is passed from Record level to value validators.

octodns/record/validator.py
```diff
@@ -121,19 +121,39 @@ def available(self):
...
             )
             self.enable_sets({'legacy'})
+        disabled = disabled or {}
         reasons = []
         for key in ('*', record_cls._type):
+            skip = disabled.get(key, ())
             for validator in self.active_record.get(key, {}).values():
-                reasons.extend(validator.validate(record_cls, name, fqdn, data))
+                if validator.id in skip and not validator.id.startswith('_'):
+                    continue
```
ross, line 136: This is the it that actually skips Record validators.

octodns/record/validator.py
```diff
@@ -121,19 +121,39 @@ def available(self):
...
+                try:
+                    reasons.extend(
+                        validator.validate(
+                            record_cls, name, fqdn, data, disabled=disabled
+                        )
+                    )
+                except TypeError as e:
```
ross, line 144: This handles any 3rd party-validators that happened to be created prior to this change. It's a really small window and I don't imagine there are any, but if so this will deprecate the behavior and give them until 2.x to add it.

octodns/record/validator.py
```diff
@@ -143,7 +163,10 @@ def process_values(self, value_type, values, _type):
             )
             reasons.extend(legacy(values, _type))
         for key in ('*', _type):
+            skip = disabled.get(key, ())
             for validator in self.active_value.get(key, {}).values():
+                if validator.id in skip and not validator.id.startswith('_'):
+                    continue
```
ross, line 177: This skip for Value validators

octodns/zone/validator.py
```diff
@@ -67,8 +67,11 @@ def process_zone(self, zone):
             )
             self.enable_sets({'legacy'})
 
+        disabled = zone.disabled_zone_validators
         reasons = []
         for validator in self.active.values():
+            if validator.id in disabled and not validator.id.startswith('_'):
+                continue
```
ross, line 77: Zone skip

brianeclow:
> Claude wanted to do a ContextVar var solution

Yup. I hit that a few times. You can usually deflect it into a different path by telling it to not assume and to ask clarifying questions.

brianeclow:
Did some testing against this branch:
- Disabling a zone validator on a specific zone works as expected with no detected leaking of disablement to other zones
- Disabling a record validator on a specific zone works as expected with no detected leaking of the disablement to other zones.
- Interesting find: When disabling the trailing dot validator for MX and having `lenient: false`, the singleton MX validator triggers because a known singleton MX without a trailing dot does not match. I accept adding to the `single_mx_regexes` var is a global configuration change because it makes the most sense based on previous conversations.

ross:
> * Interesting find: When disabling the trailing dot validator for MX and having `lenient: false`, the singleton MX validator triggers because a known singleton MX without a trailing dot does not match. I accept adding to the `single_mx_regexes` var is a global configuration change because it makes the most sense based on previous conversations.

Interesting nuance. I'll look at a PR that makes the trailing dot on those optional since it shouldn't be covering that given there's a dedicated check.

ross:
> > * Interesting find: When disabling the trailing dot validator for MX and having `lenient: false`, the singleton MX validator triggers because a known singleton MX without a trailing dot does not match. I accept adding to the `single_mx_regexes` var is a global configuration change because it makes the most sense based on previous conversations.
> 
> Interesting nuance. I'll look at a PR that makes the trailing dot on those optional since it shouldn't be covering that given there's a dedicated check.

/cc https://github.com/octodns/octodns/pull/1442/changes

brianeclow:
> https://github.com/octodns/octodns/pull/1442/changes

I am still getting the validation exception for a MX singleton.

The configuration snippet:

```
zones:
  example.com.:
    validators:
      record:
        disable_validators:
          CNAME:
          - target-value-best-practice
          MX:
          - mx-value-best-practice
```

Assuming, I am doing something wrong.

ross:
> I am still getting the validation exception for a MX singleton.

Not able to reproduce on main:

```yaml
manager:
  validators:
    enabled:
      - strict
      - best-practice
    record:
      disable_validators:
        MX:
          - mx-value-best-practice
...
```

```yaml
---
"":
  - type: MX
    value:
      exchange: smtp.google.com
      priority: 0
  - type: TXT
    value: v=spf1 -all
_dmarc:
  type: TXT
  value: v=DMARC1; p=reject;
```

```console
...
2026-07-03T14:15:29  [139914477046720] INFO    YamlProvider[out] plan:   Creates=3, Updates=0, Deletes=0, Existing=0, Meta=False
2026-07-03T14:15:29  [139914477046720] INFO    Plan 
* exxampled.com.
* out (YamlProvider)
*   Create Zone<exxampled.com.>
*   Create <MxRecord MX 3600, exxampled.com., [''0 smtp.google.com'']> (config)
*   Create <TxtRecord TXT 3600, exxampled.com., ['v=spf1 -all']> (config)
*   Create <TxtRecord TXT 3600, _dmarc.exxampled.com., ['v=DMARC1\; p=reject\;']> (config)
*   Summary: Creates=3, Updates=0, Deletes=0, Existing=0, Meta=False
```

If I comment out the bit that appended the `.` from the PR I get:

```console
octodns.zone.exception.ValidationError: Invalid zone "exxampled.com.", config/dev.yaml, line 30, column 5
  - MX record "exxampled.com." should have at least 2 values for redundancy, found 1 (config/1431/exxampled.com.yaml, line 3, column 5)
```

ross:
> [ @ross ](/ross)
> ` [Merge branch 'main' into per-zone-disable-validators](/octodns/octodns/pull/1441/commits/e49d05299ed6e0d087fe93ff148beee1835595a0)`

Ah if you were running in this branch main hadn't been merged into it yet.

brianeclow:
🤦 I thought the default was already merged.

Confirmed, works as expected. Thank you.

PR #1442, opened 2026-07-03
Fix single-MX provider exemption when exchange lacks trailing dot

ross:
## Summary

- `MailZoneValidator._is_single_mx_provider` matches an MX record's exchange against `DEFAULT_SINGLE_MX_REGEXES`/`single_mx_regexes` to exempt known single-MX providers (SendGrid, HubSpot, etc.) from the `min_mx` redundancy check. Those patterns are anchored on a trailing dot (e.g. `r'^mx\.sendgrid\.net\.$'`).
- If the exchange value doesn't have a trailing dot — e.g. the `mx-value-best-practice` trailing-dot validator has been disabled for that zone (see #1441), or the data simply omits it — the pattern no longer matches, and a legitimately single-MX zone starts failing the redundancy check with `should have at least 2 values for redundancy, found 1`.
- Fixes this by normalizing the exchange to always include a trailing dot before matching, rather than making each regex's trailing dot optional. This covers both the built-in patterns and any user-supplied `single_mx_regexes` uniformly, since it's the input being matched that's normalized rather than every pattern needing its own optional-dot handling.

Found via review feedback on #1441 (per-zone validator disabling) — disabling the trailing-dot check for one zone exposed this pre-existing coupling in `MailZoneValidator`, unrelated to that PR's actual changes.

## Test plan

- [x] `./script/test` — 782 tests passing
- [x] `./script/coverage` — 100% coverage maintained
- [x] `./script/lint` / `./script/format --check` — clean
- [x] New regression test `test_single_mx_provider_matches_without_trailing_dot`; confirmed it fails without the fix (`should have at least 2 values for redundancy, found 1`) and passes with it

### DIFF
--- a/.changelog/a4c2beb49bad46cfa2d6a1e06d11f3a6.md
+++ b/.changelog/a4c2beb49bad46cfa2d6a1e06d11f3a6.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Fix MailZoneValidator's single-MX provider exemption to match known providers even when the MX exchange is missing its trailing dot (e.g. when the mx-value-best-practice validator is disabled for a zone)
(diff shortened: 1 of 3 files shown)


ross (commented):
/ci modules build

brianeclow (commented):
Normalization for the win. 🎉

PR #1443, opened 2026-07-03
Convert record/value validator reasons to ValidationReason

ross:
## Summary
- Unify record/value validator output with the existing zone-level `ValidationReason` mechanism (previously record/value validators returned bare `list[str]`, while zone validators already returned `list[ValidationReason]`).
- `ValidationReason` moves to `octodns/record/validator.py` as a `str` subclass (bare reason text as content, rich `via: <id>` rendering via `__str__`), and stays importable from `octodns.zone.validator` (re-exported via `__all__`) for 3rd-party back-compat.
- Every first-party record/value validator now constructs `ValidationReason(text, validator_id=self.id)` directly; the registry (`ValidatorRegistry.process_record`/`process_values`) wraps any bare `str` still returned by 3rd-party validators (emitting a deprecation warning) and stamps `validator_id` onto reasons that don't already have one — conditional stamping means the `_values-type`/`_value-type` bridge validators never clobber the real inner value-validator's id.
- Record/value `ValidationError` messages now render with `via: <id>` attribution, matching zone error messages.

Related prior work: #1439 (zone validator ids), b6c6851b (moved `via:` into `ValidationReason.__str__`).

## Test plan
- [x] `./script/test` — 785 passed, 100% coverage
- [x] `./script/lint` clean
- [x] Manually verified `via: ttl-rfc` / `via: mx-value` etc. appear in record `ValidationError` messages
- [x] Confirmed bare-string list equality (`e.reasons == ['missing ttl']`) still holds due to the `str` subclass design — zero pre-existing test churn
- [x] Confirmed 3rd-party validators returning bare strings get wrapped + stamped with a deprecation warning, and the legacy `value_type.validate` classmethod path doesn't double-warn
- [x] Added `test_process_record_stamps_missing_validator_id` covering the "already a `ValidationReason` but missing `validator_id`" stamping branch

### DIFF
--- a/octodns/record/validator.py
+++ b/octodns/record/validator.py
@@ -130,7 +130,10 @@ class ValidatorRegistry:
         reasons = []
         for key in ('*', record_cls._type):
             for validator in self.active_record.get(key, {}).values():
-                reasons.extend(validator.validate(record_cls, name, fqdn, data))
+                reasons.extend(
+                    _as_reason(r, validator.id)
+                    for r in validator.validate(record_cls, name, fqdn, data)
+                )
         return reasons
 
     def process_values(self, value_type, values, _type):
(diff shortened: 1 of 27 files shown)


ross (commented):
/ci modules bump

ross:
bind's test is looking for an exact string match on reason and getting tripped up by the "via".

powerdns is hitting the deprecation (as intended) on it's custom validator.

octodns/record/validator.py
```diff
@@ -130,7 +130,10 @@ def process_record(self, record_cls, name, fqdn, data):
         reasons = []
         for key in ('*', record_cls._type):
             for validator in self.active_record.get(key, {}).values():
-                reasons.extend(validator.validate(record_cls, name, fqdn, data))
+                reasons.extend(
+                    _as_reason(r, validator.id)
```
ross, line 134: As reason auto-converts string only reasons. Either from older record validation methods that haven't migrated to validators or from 3rd parties that managed to add Validators in the short period between their first release and now. Both have been deprecated.

octodns/record/validator.py
```diff
@@ -141,13 +144,84 @@ def process_values(self, value_type, values, _type):
                 f'`{value_type.__name__}.validate` classmethod is DEPRECATED. Add a ValueValidator to `VALIDATORS` instead. Will be removed in 2.0',
                 stacklevel=3,
             )
-            reasons.extend(legacy(values, _type))
+            legacy_id = f'{value_type.__name__}.validate'
```
ross, line 147: This is the case that handles `[ValueType].validate` methods that haven't migrated.

octodns/record/validator.py
```diff
@@ -141,13 +144,84 @@ def process_values(self, value_type, values, _type):
...
+                reasons.extend(
+                    _as_reason(r, validator.id)
+                    for r in validator.validate(value_type, values, _type)
+                )
         return reasons
 
 
+class ValidationReason(str):
+    '''
+    A single validation failure reason.
```
ross, line 163: Moved from `octodns.zone.validator`, updated to inherit from `str`

ross (commented):
This should get all record, value, and zone validation reasons attributed with a "via: ID` so that it's always clear which validator is causing it.

PR #1444, opened 2026-07-08
Modernize packaging: move setup.py into pyproject.toml

ross:
## Summary

- Adds `[build-system]` and a full PEP 621 `[project]` table to `pyproject.toml` (dependencies, `dev`/`docs`/`test` extras, the 7 `octodns-*` console scripts, dynamic version via `attr = "octodns.__version__"`), plus classifiers/keywords/URLs. Removes `setup.py` entirely.
- Uses the raw `README.md` as the PyPI long description (setup.py previously stripped yaml fences and the "Supported providers" section before publishing — that processing is dropped).
- Renames `script/cibuild-setup-py` → `script/cibuild-package` and the corresponding `setup-py` CI job → `package` to match.
- Pins `[tool.black] target-version` to the actively CI-tested Python range (3.10–3.14). This is required: Black now infers its target version from `requires-python` per PEP 621, and an unbounded `>=3.9` caused it to consider unreleased future interpreter versions and crash its own AST safety check. Pinning this also legitimately unlocked a newer, already-supported syntax style Black was previously too conservative to use, hence the one incidental reformat in `tests/test_octodns_config_schema.py`.
- Fixes `script/format`/`script/lint`: both built their file list via `find *.py octodns tests docs -name '*.py'`, relying on `setup.py` being the one root-level `.py` file for that glob to expand; removed the now-dead `*.py` glob.

## Test plan

- [x] `python -m build --sdist --wheel` succeeds, no new warnings
- [x] `twine check` passes on sdist + wheel
- [x] Wheel `METADATA` has correct license expression, deps, extras, URLs, classifiers
- [x] `./script/cibuild-package` full run (clean venv, install wheel + `[test]` extra, pytest) — 784 passed
- [x] All 7 `octodns-*` console scripts resolve and run `--help` from a fresh venv install
- [x] `./script/update-requirements` (proviso) regenerates correctly from the new `[project]` metadata
- [x] `./script/generate-docs` unaffected (imports `__version__` directly)
- [x] Full pre-commit hook (changelog check, lint, format, coverage) passes; 100% coverage maintained


ross:
This is something I'm thinking about exploring. It seems a bit cleaner than the old school setup.py path. Need to verify that packaging and doc/string bits all work as desired before moving forward.

ross (commented):
/ci modules build

PR #1451, opened 2026-07-29
Fix script/ tooling to work correctly in git worktrees

ross:
## Summary
- `.git_hooks_pre-commit` resolved the hook's `$0` symlink back to the primary checkout via `readlink -f`, so committing from a linked `git worktree` ran lint/format/coverage against the wrong tree (the primary's, not the worktree being committed). Now uses `git rev-parse --show-toplevel` to find the correct tree.
- `script/bootstrap`'s git-setup block guarded on `[ -d ".git" ]`, which is false in a worktree (`.git` is a file there), silently skipping blame config and hook install. Now guards with `git rev-parse --git-dir`, installs into the *common* hooks dir via `git rev-parse --git-path hooks` (shared across all worktrees), and anchors the hook symlink to the primary checkout rather than `$ROOT` so it can't dangle once a worktree is removed.
- `script/cibuild` had no `set -e` of its own, so a failing `script/bootstrap` at the top wasn't fatal. Added `set -e` and switched to absolute script invocations.
- `CONTRIBUTING.md` documents that each worktree needs its own bootstrap, while the hook install is shared.

## Test plan
- [x] `./script/bootstrap`, `./script/lint`, `./script/format --check`, `./script/coverage` all still pass unchanged in the primary checkout (807 tests, 100% coverage)
- [x] `./script/bootstrap` completes in a fresh `git worktree` (previously silently skipped git setup)
- [x] A formatting violation introduced only in a worktree correctly fails its commit (regression test for the actual bug — before the fix this silently passed by checking the clean primary checkout instead)
- [x] Converse: a violation left in the primary checkout does not block a clean commit from a worktree
- [x] After removing a test worktree, the shared hook symlink still resolves (no dangling-symlink regression)

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -4,7 +4,7 @@
 
 set -e
 
-cd "$(dirname "$0")"/..
+cd "$(dirname -- "$(readlink -f -- "$0")")"/..
 ROOT=$(pwd)
 
 if [ -z "$VENV_NAME" ]; then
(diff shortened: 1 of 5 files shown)


script/bootstrap
```diff
@@ -4,7 +4,7 @@
 
 set -e
 
-cd "$(dirname "$0")"/..
+cd "$(dirname -- "$(readlink -f -- "$0")")"/..
```
ross, line 7: This isn't directly related to the worktree fix, it's just a belt-and-suspenders check to avoid issues if bootstrap is ever invoked via a symlink or something like that.

PR #1452, opened 2026-07-29
Add uniform value-level RDATA conversion APIs

ross:
Introduce value-level `to_rrs()` and `from_rrs()` APIs with 1.x deprecated compatibility methods.

Adds strict TXT/SPF presentation conversion, third-party legacy fallbacks, and record/filter migration.

Tests: `./script/test`, `./script/coverage`, `./script/lint`

/cc #1447 Implement the planned RDATA conversion API rework

### DIFF
--- a/octodns/record/base.py
+++ b/octodns/record/base.py
@@ -21,28 +21,25 @@ def unquote(s):
     return s
 
 
-class RdataValueMixin(object):
-    @classmethod
-    def parse_rdata_text(cls, rdata):
-        deprecated(
-            f'`{cls.__name__}.parse_rdata_text` is DEPRECATED. Use '
-            f'`{cls.__name__}.from_rrs()` instead. Will be removed in 2.0.',
-            stacklevel=2,
-        )
-        return cls.from_rrs(rdata)
+def _deprecated_parse_rdata_text(value_type):
+    deprecated(
+        f'`{value_type.__name__}.parse_rdata_text` is DEPRECATED. Use '
+        f'`{value_type.__name__}.from_rrs()` instead. Will be removed in 2.0.',
+        stacklevel=3,
+    )
 
-    @property
-    def rdata_text(self):
-        cls = self.__class__
-        deprecated(
-            f'`{cls.__name__}.rdata_text` is DEPRECATED. Use '
-            f'`{cls.__name__}.to_rrs()` instead. Will be removed in 2.0.',
-            stacklevel=2,
-        )
-        return self.to_rrs()
+
+def _deprecated_rdata_text(value):
+    value_type = value.__class__
+    deprecated(
+        f'`{value_type.__name__}.rdata_text` is DEPRECATED. Use '
+        f'`{value_type.__name__}.to_rrs()` instead. Will be removed in 2.0.',
+        stacklevel=3,
+    )
 
 
 def value_from_rrs(value_type, rdata):
+    '''Convert one presentation-format RDATA string to raw value data.'''
     method = getattr(value_type, 'from_rrs', None)
     if method:
         return method(rdata)
@@ -55,6 +52,7 @@ def value_from_rrs(value_type, rdata):
 
 
 def value_to_rrs(value):
+    '''Render one logical value as one presentation-format RDATA string.'''
     method = getattr(value, 'to_rrs', None)
     if method:
         return method()
--- a/octodns/record/base.py
+++ b/octodns/record/base.py
@@ -371,11 +371,11 @@ class Record(EqualityTupleMixin):
     def from_rrs(cls, zone, rrs, lenient=False, source=None):
         '''Create records from deprecated, individual :class:`~octodns.record.rr.Rr` objects.
 
-        The flat input is grouped by owner name and type and converted through
-        :meth:`from_rrsets`. Input order is retained within each group and
-        output records are ordered deterministically by owner name and type.
-        ``lenient`` and ``source`` are passed unchanged to
-        :meth:`Record.new`.
+        The flat input is grouped by owner name and type and converted with
+        each record class's legacy ``data_from_rrs()`` implementation. Input
+        order is retained within each group and output records are ordered
+        deterministically by owner name and type. ``lenient`` and ``source``
+        are passed unchanged to :meth:`Record.new`.
 
         :param octodns.zone.Zone zone: zone containing the records
         :param collections.abc.Iterable rrs: individual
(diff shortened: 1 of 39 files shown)


ross:
/cc @jleroy after a couple false starts at trying to fix TXT's handling without breaking existing providers I (with codex/sol's help) decided this was the best path. I think it results in a cleaner and more consistent API across the board with a minimal amount of churn. All the actual details of dealing with the TXT rdata processing are off-loaded to dnspython which is vastly preferable to trying to re-implement a complete version of it in octoDNS. If this moves forward the existing stuff would go away as part of 2.x

ross:
All the above said now that I actually started looking over and reviewing the code it's complete and absolute :hankey:. The plan `sol` came up after a couple of rounds of small suggestions/tweaks was solid, but this is an abysmal implementation of that plan. I dropped it back to the default model for the implementation. I guess that was a mistake. Another round incoming where I find an option that isn't completely useless.

ross:
The plan:

# Introduce Uniform Value-Level RR Conversion APIs

## Summary

Add `to_rrs()` and `from_rrs()` to every core value type currently exposing
`rdata_text` and `parse_rdata_text`. Internal RR conversion will use the new
APIs immediately, while the old APIs remain behaviorally compatible and
deprecated throughout octoDNS 1.x.

## API and Implementation Changes

- For already-compliant value types:
  - Move `rdata_text` logic into `to_rrs()`.
  - Move `parse_rdata_text` logic into the classmethod `from_rrs()`.
  - Keep `rdata_text` as a deprecated property delegating to `to_rrs()`.
  - Keep `parse_rdata_text()` as a deprecated classmethod delegating to
    `from_rrs()`.
  - Use notices such as: `` `CaaValue.rdata_text` is DEPRECATED. Use
    `CaaValue.to_rrs()` instead. Will be removed in 2.0. ``

- Define the value-level contracts explicitly:
  - `value.to_rrs()` returns one valid presentation-format RDATA string for
    that logical value.
  - `ValueType.from_rrs(rdata)` accepts one presentation-format RDATA string
    and returns the raw data used to construct that value type.
  - TXT character-string chunking remains contained within one returned RDATA
    string.

- For TXT/SPF values:
  - `to_rrs()` performs strict RFC presentation rendering through dnspython,
    including quoting, escaping, and byte-based 255-octet chunking.
  - `from_rrs()` strictly parses presentation text, concatenates chunks,
    decodes escapes, and restores octoDNS's internal semicolon representation.
  - Deprecated `rdata_text` retains its raw-value behavior throughout 1.x.
  - Deprecated `parse_rdata_text()` retains its legacy raw-input behavior
    throughout 1.x.

- Update record-level conversion:
  - Existing `record.rrs` uses each value's `to_rrs()`.
  - Existing `Record.from_rrs()` uses the relevant value type's `from_rrs()`.
  - Preserve the current `record.rrs` tuple shape and the existing `rr_values`,
    `chunked_value()`, and `chunked_values` interfaces.
  - TXT serialization renders directly from raw values so presentation text is
    never wrapped as raw `TxtValue` and rendered twice.

- Migrate internal callers:
  - Use `to_rrs()`/`from_rrs()` for serialization, `Record.from_rrs`, filters,
    and other presentation-format paths.
  - Keep raw-format sources such as TinyDNS on an explicit raw TXT path.
  - Ensure normal octoDNS operations do not invoke deprecated methods.

- Preserve third-party value types during 1.x:
  - Record conversion first looks for `to_rrs()`/`from_rrs()`.
  - Fall back to `rdata_text`/`parse_rdata_text()` for custom value types
    implementing only the old API, with a deprecation warning naming the type.
  - Remove the compatibility fallback in 2.0.

## Test Plan

- Verify every core RDATA value type exposes `to_rrs()` and `from_rrs()`,
  including inherited record types.
- Verify deprecated methods retain their exact 1.x output and emit
  `DeprecationWarning`.
- Test TXT/SPF spaces, quotes, backslashes, semicolons, control bytes, empty
  values, multiple character-string chunks, and 255-octet boundaries.
- Verify `record.rrs -> Record.from_rrs()` round trips every record type through
  the new value APIs.
- Add a mock third-party value type implementing only the legacy API and verify
  fallback behavior.
- Run core test, coverage, lint, and formatting checks plus representative
  provider suites.

## Assumptions

- The plural `rrs` name follows the existing record API, while each value-level
  call converts exactly one logical value and one RDATA string.
- Non-RDATA helper values such as Geo metadata are excluded.
- Deprecated APIs are removed in octoDNS 2.0.

octodns/record/base.py
```diff
@@ -21,6 +22,71 @@ def unquote(s):
...
+    for index, owner in enumerate(value_type.__mro__):
+        if name in owner.__dict__:
+            return index
+    return None
+
+
+def value_from_rdata_text(value_type, rdata):
+    '''Convert one presentation-format RDATA string to internal value data.'''
+    new_owner = _mro_owner(value_type, 'from_rdata_text')
+    legacy_owner = _mro_owner(value_type, 'parse_rdata_text')
```
ross, line 55: This would likely result in a lot of repeated walking of MROs. It wouldn't be too bad in reality since most of the time the first thing we come to is going to implement both. If there are clean and simple options to avoid it it might be worth exploring.

octodns/record/base.py
```diff
@@ -21,6 +22,71 @@ def unquote(s):
...
+        f'`{value_type.__name__}.parse_rdata_text` is DEPRECATED. '
+        'Implement `from_rdata_text()` instead. Will be removed in 2.0.',
+        stacklevel=3,
+    )
+    return value_type.parse_rdata_text(rdata)
+
+
+def _value_to_rdata_text_uses_legacy(value_type):
+    new_owner = _mro_owner(value_type, 'to_rdata_text')
+    legacy_owner = _mro_owner(value_type, 'rdata_text')
```
ross, line 86: Same MRO walking here, for different methods.

jleroy:
Updated my provider to use the new API and everything seems to be working well. See: https://github.com/relkian/octodns-opusdns/pull/1.

ross:
> Updated my provider to use the new API and everything seems to be working well. See: [relkian/octodns-opusdns#1](https://github.com/relkian/octodns-opusdns/pull/1).

Cool. I'm planning to do PRs to (at least some of) the providers that utilize the rdata functionality to make sure the new APIs are a good fit for their needs. Once I've gotten a chance to sit down and do that this will be ready to move forward.

ross:
My user.name and user.email got corrupted at some point during this work, though things were still signed w/my key, so I did a --reset-author and force pushed :-(

ross (commented):
/ci modules build

ross:
OK, 3 linked PRs that do updates to 3 provider repos making use of these changes. A lot less code/details in two of the three as a result and the 3rd is complicated b/c of inconsistencies in the underlying services quoting of TXT values so I think this is a win all around. 

Going to let it sit for a few days and come back and take a close :eyes: at all the changes here and in those PRs before moving forward.

PR #1454, opened 2026-07-30
Add uniform value-level RR conversion APIs

### LINKED ISSUE #1447: Best practices regarding string values quotation

ross:
## Summary

- add to_rrs/from_rrs across all core RDATA value types while preserving deprecated 1.x APIs
- use strict dnspython presentation conversion for TXT/SPF and retain explicit raw TinyDNS handling
- preserve third-party legacy fallbacks and migrate internal RR conversion/filter callers
- add all-record round-trip, deprecation, compatibility, and TXT edge-case coverage

## Validation

- ./script/test (815 passed)
- ./script/coverage (100% statements and branches)
- ./script/lint
- representative Route53, Bind, and Cloudflare functional suites

/cc Fixes #1447

### DIFF
--- a/.changelog/a0e3957ab7d6443a9005fabf5ce91b29.md
+++ b/.changelog/a0e3957ab7d6443a9005fabf5ce91b29.md
@@ -1,0 +1,4 @@
+---
+type: minor
+---
+Add uniform value-level RR conversion APIs
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -337,9 +337,9 @@ Processors
    * - `NameRejectlistFilter`_
      - Filter that IGNORES records that match specified naming patterns, all others will be managed
    * - `ValueAllowlistFilter`_
-     - Filter that ONLY manages records that match specified value patterns based on `rdata_text`, all others will be ignored
+     - Filter that ONLY manages records that match specified value patterns based on presentation-format RDATA from `to_rrs()`, all others will be ignored
    * - `ValueRejectlistFilter`_
-     - Filter that IGNORES records that match specified value patterns based on `rdata_text`, all others will be managed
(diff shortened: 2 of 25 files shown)


ross:
This is being abandoned in favor of https://github.com/octodns/octodns/pull/1452 which is a more complete and robust solution.

PR #1456, opened 2026-07-30
Fix OwnershipProcessor dropping changes to record types with lower-case letters

### LINKED ISSUE #1455: OwnershipProcessor silently drops changes to provider-specific record types with lower-case letters
## Summary

`OwnershipProcessor.process_plan` builds its ownership index with `_type.upper()` but filters the plan using the record's real `_type`. Core types are already upper case, so the mismatch is invisible for them. A provider-specific type containing lower-case letters never matches, and changes to those records are dropped from the plan with no warning, even when the ownership TXT exists.

`Route53Provider/ALIAS` from `octodns-route53` is one such type, so an apex alias record managed with `OwnershipProcessor` enabled can never be created, updated or deleted.

## Where

`octodns/processor/ownership.py` on `main`:

- line 84 — `_decode_ownership_name` returns `name, _type.upper()`, so the index is keyed `'ROUTE53PROVIDER/ALIAS'`
- line 136 — the filter tests `record._type not in owned[record.name]`, where `record._type` is `'Route53Provider/ALIAS'`

The two never compare equal, so the change is skipped by the `continue` below.

## Reproduction

No credentials or network access required.

```python
import octodns_route53  # noqa: F401  registers the Route53Provider/ALIAS type
from octodns.processor.ownership import OwnershipProcessor
from octodns.provider.plan import Plan
from octodns.record import Record, Update
from octodns.zone import Zone

ZONE = 'example.com.'


def alias(zone, target):
    return Record.new(zone, '', {
        'type': 'Route53Provider/ALIAS', 'ttl': 300,
        'value': {'evaluate-target-health': False,
                  'hosted-zone-id': 'Z2FDTNDATAQYW2',
                  'name': target, 'type': 'A'}})


def cname(zone, name, target):
    return Record.new(zone, name, {'type': 'CNAME', 'ttl': 300, 'value': target})


def ownership_txt(zone, name):
    return Record.new(zone, name,
                      {'type': 'TXT', 'ttl': 60, 'value': 'octodns'}, lenient=True)


def run(label, old, new, marker_name):
    existing, desired = Zone(ZONE, []), Zone(ZONE, [])
    marker = ownership_txt(existing, marker_name)
    for r in (old, marker):
        existing.add_record(r, lenient=True)
    for r in (new, marker):
        desired.add_record(r, lenient=True)

    plan = Plan(existing, desired, [Update(old, new)], True)
    proc = OwnershipProcessor('ownership', txt_name='_octodns', txt_value='octodns')
    result = proc.process_plan(plan, sources=[], target=None)
    kept = [c for c in (result.changes if result else [])
            if c.record._type == new._type]
    print(f'  {label:22} type={new._type:24} -> {"KEPT" if kept else "DROPPED"}')


z = Zone(ZONE, [])
run('core type', cname(z, 'www', 'old.example.net.'),
    cname(z, 'www', 'new.example.net.'), '_octodns.CNAME.www')
run('provider type', alias(z, 'old.cloudfront.net.'),
    alias(z, 'new.cloudfront.net.'), '_octodns.route53provider/alias')
```

Output:

```
  core type              type=CNAME                    -> KEPT
  provider type          type=Route53Provider/ALIAS    -> DROPPED
```

Both records have their ownership marker present in the existing and desired zones. Only the mixed-case one is dropped.

## Impact

Silent, and the printed plan contradicts the provider's own log line. Against a real zone with one stale apex alias and no other pending change:

```
INFO  Route53Provider[route53] plan:   Creates=0, Updates=1, Deletes=0, Existing=1
...
*   Summary: Creates=0, Updates=0, Deletes=0, Existing=1
```

The provider detects the diff, the processor removes it, and `octodns-sync` reports `No changes were planned`. Editing the record has no effect and there is no indication why.

## Possible fix

Compare case-insensitively at line 136, or stop upper-casing at line 84 and normalise both sides consistently. Happy to open a PR with tests if you have a preference between the two.

## Versions

- octodns 1.21.0, also reproduced on 1.15.0
- octodns-route53 1.3.0, also reproduced on 1.0.1
- Python 3.14.6

## Note on `octodns.excluded`

A per-record processor exclusion is not a workaround. `record.excluded` is a list of provider ids checked in `Zone.changes` (see #1036), so `octodns: {processors: {excluded: [ownership]}}` is silently ignored and `process_plan` still filters the change. The workaround that does work is a second config for the affected record with `OwnershipProcessor` omitted, scoped down with `NameAllowlistFilter` and `TypeAllowlistFilter` so nothing else is in range.

albority:
Fixes #1455.

## The bug

`_decode_ownership_name` upper cases the type it reads out of the ownership TXT name. It has to: record names are lower cased on the way in, so the marker for an `A` record arrives as `_owner.a.the-a`, as the existing `test_process_source_zone` assertions show.

The plan filter then compared the record's real `_type` against that upper cased index:

```python
and record._type not in owned[record.name]
```

Core types are already upper case, so the mismatch is invisible for them. A provider-specific type containing lower-case letters never matches, and its changes are dropped from the plan with no warning even when we own the record. `octodns-route53`'s `Route53Provider/ALIAS` is one such type, which makes an apex alias unmanageable whenever `OwnershipProcessor` is enabled. The symptom is quiet: the provider logs `Updates=1` and the printed plan `Summary` says `Updates=0`.

## The fix

Normalise the comparison side too, so both sides are consistent.

I considered removing the upper casing at `_decode_ownership_name` instead and comparing raw. That does not work: the type is recovered from a lower cased record name, so raw comparison fails for every mixed-case type in the other direction. Both sides have to be normalised, not neither.

## Tests

`test_process_plan_mixed_case_type` registers a `Provider/MiXeD` record type, following the pattern in `tests/test_octodns_processor_templating.py`, and covers both directions:

- an owned record of that type keeps its change, which fails on `main` with `AssertionError: None is not true`
- an unowned record of that type is still left alone, so the fix does not widen what the processor takes over

No dependency on `octodns-route53` — the lower casing that causes this is octodns's own, so a locally registered type reproduces it.

`MixedCaseValue` also implements `rdata_text`, because `test_records_have_rdata_methods` audits every registered type and a fixture that skips it fails that test.

## Checks

- `./script/test` — 808 passed
- `./script/coverage` — 100.00%, requirement met
- `./script/lint` — clean
- `./script/format` — applied
- changelog fragment added via `./script/changelog create -t patch`

### DIFF
--- a/.changelog/4894bc8785d846b19681fd9557eb96f3.md
+++ b/.changelog/4894bc8785d846b19681fd9557eb96f3.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Fix OwnershipProcessor dropping changes to records whose type contains lower-case letters, e.g. octodns-route53's Route53Provider/ALIAS - #1455
--- a/octodns/processor/ownership.py
+++ b/octodns/processor/ownership.py
@@ -133,7 +133,7 @@ class OwnershipProcessor(BaseProcessor):
 
             if (
                 not self._is_ownership(record)
-                and record._type not in owned[record.name]
+                and record._type.upper() not in owned[record.name]
                 and record.name != 'octodns-meta'
             ):
                 # It's not an ownership TXT, it's not owned, and it's not
--- a/tests/test_octodns_processor_ownership.py
+++ b/tests/test_octodns_processor_ownership.py
@@ -8,7 +8,7 @@ from helpers import PlannableProvider
 
 from octodns.processor.ownership import OwnershipException, OwnershipProcessor
 from octodns.provider.plan import Plan
-from octodns.record import Delete, Record
+from octodns.record import Delete, Record, Update, ValueMixin
 from octodns.zone import DuplicateRecordException, Zone
 
 zone = Zone('unit.tests.', [])
@@ -26,6 +26,34 @@ for record in [
 ]:
     records[record.name] = record
     zone.add_record(record)
+
+
+class MixedCaseValue(str):
+    @classmethod
+    def parse_rdata_text(cls, value):
+        return value
+
+    @classmethod
+    def validate(cls, data, _type):
+        return []
+
+    @classmethod
+    def process(cls, value):
+        return MixedCaseValue(value)
+
+    @property
+    def rdata_text(self):
+        return self
+
+
+class MixedCase(ValueMixin, Record):
+    # Provider-specific types are not required to be upper case, e.g.
+    # octodns-route53's Route53Provider/ALIAS
+    _type = 'Provider/MiXeD'
+    _value_type = MixedCaseValue
+
+
+Record.register_type(MixedCase, 'Provider/MiXeD')
 
 
 class TestOwnershipProcessor(TestCase):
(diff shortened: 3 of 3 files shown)


ross (approved):
Good catch/call. Definitely an oversight. Custom record types weren't being considered. Thanks.

ross:
selectel failure is unrelated.

Issue #1427, opened 2026-06-07
Record is under a managed subzone


muhlba91:
Hi,

I tried to upgrade octodns to 1.18.0 and now receive a ValidationError for my NS records' ownership record:

The config can be found here: https://github.com/muhlba91/muehlbachler-global-dns/blob/main/config.yaml.
I have also tried the following `zone` block but it didn't change anything:
```
...
zones:
  muehlbachler.io.: &dns_muehlbachler_dns
    ignore_subzone_adds: true
    sources:
      - yaml-config
    processors:
      - ownership
      - trailing-dots
      - exclude-root-ns-changes
      # - no-root-ns
    targets:
      - googlecloud-muehlbachler-dns
  email.muehlbachler.io.: *dns_muehlbachler_dns
  muehlbachler.xyz.: *dns_muehlbachler_dns
  niftyside.io.: *dns_muehlbachler_dns
  muehlbachler.org.: *dns_muehlbachler_dns
  niftyside.com.: *dns_muehlbachler_dns
```

The full debug log is:
```
❯ octodns-sync --config-file ./config.yaml --debug
2026-06-07T13:04:35  [8412700352] INFO    Manager __init__: config_file=./config.yaml, (octoDNS 1.18.0)
2026-06-07T13:04:35  [8412700352] INFO    Manager _config_executor: max_workers=2
2026-06-07T13:04:35  [8412700352] INFO    Manager _config_include_meta: include_meta=False
2026-06-07T13:04:35  [8412700352] INFO    Manager _config_enable_checksum: enable_checksum=False
2026-06-07T13:04:35  [8412700352] DEBUG   Manager _config_secret_handlers: configuring secret_handlers
2026-06-07T13:04:35  [8412700352] INFO    Manager _config_auto_arpa: auto_arpa=False
2026-06-07T13:04:35  [8412700352] INFO    Manager __init__: global_processors=[]
2026-06-07T13:04:35  [8412700352] INFO    Manager __init__: global_post_processors=[]
2026-06-07T13:04:35  [8412700352] DEBUG   Manager _config_providers: configuring providers
2026-06-07T13:04:35  [8412700352] DEBUG   Manager _build_kwargs: failed to find handler for key "directory"
[...]

PR #1457, opened 2026-07-30
Fix Sphinx docs build warnings

ross:
## Summary

`./script/generate-docs` (which passes `--fail-on-warning`) currently fails with 55
warnings against the pinned Sphinx 9.1.0 / docutils 0.22.4 toolchain. Nothing in CI
actually runs the docs build, so this regression went unnoticed on `main`.

- **51x** `more than one target found for cross-reference 'ValidationReason'` — the
  back-compat re-export in `octodns/zone/validator.py` was listed in `__all__`, which made
  autodoc document `ValidationReason` a second time on that module's page (its `__module__`
  skip-check only applies when a module has no `__all__`). Dropped `__all__` and switched
  to a bare-name re-export (same idiom already used in `octodns/zone/__init__.py`) so the
  class is documented once, at its real home (`octodns.record.validator`).
  `from octodns.zone.validator import ValidationReason` and `import *` both keep working.
- **2x** `[autosummary] failed to import octodns.zone.subzone` — that module was deleted in
  #1427 (`e464aac2`) but the docs listing in `docs/api/validators.rst` was never updated.
  Removed the stale entry.
- **1x** `Inline literal start-string without end-string` — a table row in
  `docs/configuration.rst` (`openpgpkey-value-rfc`) overflowed its column, clipping the
  closing double-backtick. Widened the column.
- **1x** `Blank line required after table` — missing blank line after a table in the same
  file. Added it.
- Extended `script/generate-docs`'s cleanup to remove all 8 gitignored generated API stub
  dirs (it previously only cleaned 2 of them), so a deleted module can't leave a stale
  generated `.rst` behind and mask exactly the kind of drift in the `subzone` case above.
- Added a `docs` job to `.github/workflows/main.yml` that bootstraps and runs
  `./script/generate-docs` on PRs, so a future docs regression fails the build instead of
  going unnoticed.

/cc #1427 removed `octodns/zone/subzone.py`, leaving the stale docs entry this PR cleans up

## Test plan

- [x] `./script/generate-docs` → `build succeeded.` with zero warnings
- [x] Confirmed `ValidationReason` back-compat import and `import *` still work
- [x] Confirmed `ValidationReason` renders on the `octodns.record.validator` docs page and
      no longer has a duplicate entry on `octodns.zone.validator`'s page
- [x] Confirmed the fixed table renders correctly in `configuration.html`
- [x] `./script/lint`, `./script/format --check`, `./script/test`, `./script/coverage` all
      pass (100% coverage maintained)

### DIFF
--- a/.changelog/8609e000ae4949e898462db3d76112b0.md
+++ b/.changelog/8609e000ae4949e898462db3d76112b0.md
@@ -1,4 +1,4 @@
 ---
-type: patch
+type: none
 ---
 Fix Sphinx docs build warnings and add a CI job to catch future regressions
--- a/.changelog/8609e000ae4949e898462db3d76112b0.md
+++ b/.changelog/8609e000ae4949e898462db3d76112b0.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Fix Sphinx docs build warnings and add a CI job to catch future regressions
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,3 +66,20 @@ jobs:
       - name: CI package build
         run: |
           ./script/cibuild-package
+  docs:
+    needs: config
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          # Most recent release from https://devguide.python.org/versions/#versions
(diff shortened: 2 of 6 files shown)


.changelog/8609e000ae4949e898462db3d76112b0.md
```diff
@@ -0,0 +1,4 @@
+---
+type: patch
```
ross:
```suggestion
type: none
```

